PR #1, opened 2020-07-14
Initial Commit for QIP

sarsid:
This commit is meant to get the ball-rolling for QIP. The basic code is added. The namespace issue is not yet addressed. To try it out, I would recommend installing it from `setup.py` locally and having the latest QuTiP dev version (or atleast the version containing measurement). My only gripe currently (admittedly with myself) is the main code folder is called `qip` which means that if someone wants to use it without the `qutip.qip` namespace trick, they have to import qip which sounds rather pompous (and might lead to more than a few conflicts?). 

The next step is to get namespaces correct and tests up and running. 

PS. One can import from either `qutip.qip` and `qip` (provided you have qutip_qip) installed and it seems to work although I haven't tested anything exhaustively.


BoxiLi:
I remember that it is recommended that the code folder is called something like `src` instead of the namespace name, for which qutip is not a good example. Should we change this in `qutip-qip`? @ajgpitch

BoxiLi:
Hi @sarsid, is this PR still up-to-date? It would be great if you could make a PR with all the stuff that you have at the moment. It doesn't have to be package-ready. Just including the effort you've made so far like using it as a plug-in package for `qutip` and keep the name of some other contributors. No worry if it is not up-to-date to the current `qutip.qip` module.

sarsid:
Hello @BoxiLi, thanks for checking in, I need to tie up a few things, I have been a bit busy but I'll get back to it as soon as possible so that work can proceed.

PR #3, opened 2021-02-09
QIP subpackage: Commit Histories

sarsid:
This is supposed to be the basic QIP subpackage commit with the commit histories intact from the QuTiP package. We can modify the setup structure as we go but for now, it can be installed as: 

`pip install -e .`

and imports can be done as: 

`from qutip_qip.x import y`

as opposed to 

`from qutip.qip.x import y`

before. 

Things to check/decide: 

1. I have kept the test files from the  `qutip/tests` in the `tests` folder that contained atleast one `qutip.qip` import. It would be nice to make sure that these are the exact test files we want as it is a bit cumbersome to import files from the QuTiP package while maintaining the commit history. I can redo it now if we are missing something but something we should avoid in going forward. 

2. It would be good to fix things like the package name and basic structure under the `src/qip` folder. 

The next step would be to get the tests up and running to make sure that the imports in each file under `src/qip` are working fine.


BoxiLi (commented):
@sarsid Excellent! Many Thanks!

I think we can remove `qip` from he path and keep it like `src/pulse.py` instead of `src/qip/pulse.py`. `qip` seems a bit unnecessary here because it is the only folder. Is this feasible or very hard (like destroying the contribution history)? @sarsid 

The following files in the tests can be removed:
- `Hadamard_params.ini` (I don't know what this is for, probably optimal control? @ajgpitch) 
- `test_control_pulseoptim.py`
- `test_metric.py` 
- `test_entroy.py`
- `test_qpt.py`
- `test_superop_reps.py`

They all use quantum gates defined in qutip, but not test qip. As a todo list for later, I propose to keep those simple gates in the main qutip repo and import them into qip for consistency.

Everything else looks pretty nice to me. I didn't see any missing files from `qutip.qip`. 

@ajgpitch @nathanshammah @hodgestar Would you also like to have a look for this initial PR?

sarsid:
Thanks Boxi, I will remove the unnecessary test files. 

I think @goerz will be able to best tell us whether putting the files in `src` is a good idea. Most implementations I have seen of this style seem to go with `src/package-name/`.

hodgestar:
The name of the package folder should match the name we want to use to import the package, so likely `src/qutip_qip` rather than `src/qip`, otherwise odd things will happen when one tries to import the package directly from the source files.

I've never found the `src/` folder to be that useful, but it's not an uncommon thing to see.

setup.py
```diff
@@ -0,0 +1,190 @@
...
+# all information about QuTiP goes here
+MAJOR = 0
+MINOR = 0
+MICRO = 1
+ISRELEASED = False
+VERSION = '%d.%d.%d' % (MAJOR, MINOR, MICRO)
+REQUIRES = ['numpy (>=1.12)', 'scipy (>=1.0)', 'qutip (>=4.5)']
+EXTRAS_REQUIRE = {'graphics': ['matplotlib(>=1.2.1)']}
+INSTALL_REQUIRES = ['numpy>=1.12', 'scipy>=1.0', 'qutip>=4.5']
+PACKAGES = ['src/qip', 'src/qip/device', 'src/qip/operations', 'src/qip/compiler',
```
hodgestar, line 64: You can probably get away with specifying only `src/qutip_qip` here and having a good MANIFEST.in file.

setup.py
```diff
@@ -0,0 +1,190 @@
...
+    finally:
+        a.close()
+
+local_path = os.path.dirname(os.path.abspath(sys.argv[0]))
+print(local_path)
+os.chdir(local_path)
+sys.path.insert(0, local_path)
+print(local_path)
+sys.path.insert(0, os.path.join(local_path, 'src/qip'))  # to retrive _version
+print(local_path)
```
hodgestar, line 131: This whole paragraph looks like debugging code that can be removed? The `prints` should definitely no be here and I don't see `version` is only written here, not imported, so probably it doesn't need to be on `sys.path` anywhere?

hodgestar:
I obviously didn't read all of the qip code, but this looks like a great start and thank you for bringing across the history from the original repo. I left a couple of comments, but probably only the one about the debugging code in `setup.py` needs to be resolved before this PR is merged.

Possible next steps I'd recommend for upcoming PRs:

* Get tests running locally.
* Set up running of tests in CI.

Since we're likely to make multiple `qutip_XXX` projects, it might be an idea to use something like cookiecutter to make a template for creating repositories with a consistent layout, rather than doing it manually each time.

setup.py
goerz: `find_packages(where="src")` should do it, where `find_packages` is imported from `setuptools`

goerz:
The case for `./src` is being made in https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure, and I've only had good experiences adopting it.

setup.py
sarsid: Done, thanks for the comment! I think I'd left it in there when I was working on the same PR before (now closed).

setup.py
sarsid: I have made these changes.

sarsid:
All the changes requested have been made. I also made a minor import change since imports of the type 

`import qutip_qip` 

were not working. The PR should be good to go modulo any new comments.

setup.py
goerz: See e.g. https://github.com/QAlgebra/qalgebra/blob/master/setup.py (and the corresponding [`MANIFEST.in`](https://github.com/QAlgebra/qalgebra/blob/master/MANIFEST.in) for a project where I use the `src` layout

sarsid: Does it look okay now?

BoxiLi (approved):

hodgestar:
> The case for `./src` is being made in https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure, and I've only had good experiences adopting it.

I don't think any of the points in the article are hard to have without `/src` but if people like it, let's go with it. :)

Issue #2, opened 2020-07-14
Steps to Maintaining Commit History


sarsid:
It seems like maintaining the commit history is slightly more tricky than anticipated. I am trying to see if there is a clean way to do it but in the meantime it would be nice to decide on the name for the folder where all our source files should sit in. I think @BoxiLi proposed `src`. We could also go with something like qutip_qip if we want to stick to that convention of naming. 

There is some guidance if we go down the `src` path: 

https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure

but I am not sure how well this will all gel with core qutip if we do entrypoints.

BoxiLi:
I found it, @goerz once argued in the google dev group that `src` is a better choice with this link https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure.

goerz:
I'm definitely a proponent of `src`, but it shouldn't have any impact on how `qutip-qip` interacts with `qutip-core` or any other package: The `src` subdirectory only simplifies packaging and testing a little bit, in that there's no valid python package in the project root that might be accidentally be imported over the version of the package that's installed in the testing-virtual-environment.

In any case, I have quite a bit of experience with package organization/testing/Sphinx (much of it condensed into my personal [cookiecutter template](https://github.com/goerz/cookiecutter-pypackage)), so just let me know if there's anything I can help with.

sarsid:
@BoxiLi @ajgpitch I find the argument convincing enough to go with it. While I don't think I'm going to end up using @goerz's package, it is immensely helpful to get an idea of all the things we need to set up. I figured out a way using the[ git-filter-repo ](https://github.com/newren/git-filter-repo) package which is immensely useful. I am still figuring out a way to do it for the docs and the tests!

PR #5, opened 2021-02-14
Move pytest.ini to root

BoxiLi:
I propose to move `pytest.ini` to the root folder.

Currently `pytest.ini` is in the `tests` folder and running `pytest` in the root directory will not collect all the customized `pytest` marks. One needs to change the test path to `qutip_qip/tests`. In other packages, it seems common to put it in the root folder.

@jakelishman Is there any particular reason for putting `pytest.ini` under the `tests` folder? It seems to me that in `qutip.testing.py`, the test path is also not correctly specified, in contrast to `travis.yml`. https://github.com/qutip/qutip/blob/9b3671ed0e9c6342743299c93a0b9092789af499/qutip/testing.py#L57-L61

This should also be merged after #2


jakelishman:
There was no particular reason I put it there - it was right when I was first learning pytest, and the docs say you can put it either in the root or in the tests directory.  Nowadays you can just [put all the pytest stuff in `pyproject.toml`](https://docs.pytest.org/en/stable/customize.html#pyproject-toml), so you don't even need a separate file.

PR #9, opened 2021-03-14
Fix circuit latex plot

BoxiLi:
Same as https://github.com/qutip/qutip/pull/1447


nathanshammah:
Is this PR ready for review?

BoxiLi:
Yes, I used git patches to move it from commits on main qutip to qutip-qip, instead of copying by hand. So if there are no strange behaviours in the changed files, it should be the same.

BoxiLi:
BTW, @nathanshammah if you have time planned for QuTiP at the moment, could you also check the qip PRs on the main QuTiP repo? They are a bit big, but Jake already had the first round of review. It would be good to have them merge before GSoC.

PR #19, opened 2021-05-01
Create the superconducting qubits model

BoxiLi:


nathanshammah:
@BoxiLi maybe calling it "chain" can be a bit confusing with respect to the spin chain? what about 1d_circuitqed? or circuitqed? technically I understand the parallelism with the spin chain, it's just I've never heard "transmon chain" as a term.

BoxiLi:
I see, well `1d_circuitqed` won't work because names can never start from a number. What about just `Transmon`? `Circuitqed` is a bit too general as there are other types of qubits.

PR #20, opened 2021-05-03
Create a clean setup.cfg and remove setup.py

BoxiLi:
I'm preparing a release for qutip_qip. This PR cleans up the configuration files.

Since qutip_qip requires no cython compilation, the configuration in `setup.cfg` is sufficient for the build. So I removed setup.py and copied+edited @jakelishman's `setup.cfg` for qutip.

`graft src` is removed from MANIFEST because it includes all the pycaches. Without it, `setuptools` automatically includes all src py files and removes all the pyc.

@nathanshammah @jakelishman @hodgestar @ajgpitch You are more experienced in packaging. Does these configurations look reasonable to you? Any suggestions?


pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[build-system]
+requires = [
+    "setuptools>=42",
+    "wheel",
+    "numpy>=1.16.6",
+    "scipy>=1.0",
+    "qutip>=4.6",
```
jakelishman, line 7: If you have no code that actually executes or runs in a `setup.py`, do you need `numpy`, `scipy` or `qutip` at build time?

setup.cfg
```diff
@@ -0,0 +1,46 @@
...
+author = Alexander Pitchford, Paul D. Nation, Robert J. Johansson, Chris Granade, Arne Grimsmo, Nathan Shammah, Shahnawaz Ahmed, Neill Lambert, Eric Giguere, Boxi Li, Jake Lishman
+author_email = qutip-admin@googlegroups.com
+platforms = Linux, Mac OSX, Unix, Windows
+version = 0.1.0
+
+[options]
+package_dir=
+    =src
+packages = find:
+zip_safe = False
```
jakelishman, line 30: Since your tests (which have data files) aren't accessible from an `import qutip_qip` command, you probably _are_ zip-safe.

setup.cfg
```diff
@@ -0,0 +1,46 @@
...
+include_package_data = True
+install_requires =
+    numpy>=1.16.6
+    scipy>=1.0
```
jakelishman, line 34: You're probably don't actually have any version constraints - QuTiP currently requires `numpy>=1.16.6` because that had some changes to the C-API.  That said, v1.16 of Numpy and v1.0 of Scipy are both quite old now, so this isn't much more than no constraint at all.

setup.cfg
```diff
@@ -0,0 +1,46 @@
...
+    qutip>=4.6
```
jakelishman, line 35: Since you use `QobjEvo`'s `tlist`, you're probably not compatible with QuTiP v5 right now - you can be `qutip>=4.6,<5`.

jakelishman (commented):
They all look reasonable to me - maybe a little overconstrained in some places (comments), but that shouldn't actually cause any problems.

When you do the release, remember to change the first line of your PyPI classifiers.  I always forget that they exist.

pyproject.toml
BoxiLi: I'm always a bit confused by what should be in `requirements.txt `and `pyporject.toml`.

So `pyproject.toml` is used by `pip` to build the project when installing. Without `setup.py`, none of the python script in `qutip_qip` is executed so that I don't have any dependencies for the build except for the build tools.

And `requirements.txt` includes things needed for running the code.

That roughly correct?

setup.cfg
BoxiLi: Ok, I'll add that. Although there is no qutip v5 at the moment :)

pyproject.toml
jakelishman: `requirements.txt` is basically nothing,  It has no formal part of the build/installation process at all - I think maybe it was used more in the days before the dependency manager was reliable?  I use `requirements.txt` in the documentation build to exactly specify a complete environment to do the builds in; it inclues exact pinned versions of every package I had installed when I ran a successful documentation build, similar to a `package-lock.json` file in the npm world.  You don't even need to have one if it's not appropriate.

The `build-system.requires` key in `pyproject.toml` has everything that needs to be installed to create the sdists and wheels - if you don't actually execute any Python code/do any compilation, you're probably fine with `setuptools` and `wheel` in there only.  This key deprecates the `setup_requires` key in `setup.cfg` or `setup.py` if you're using `pip install -e .` to get a development version; pip will read from `pyproject.toml` _before_ setuptools would have a chance to see what was going on.  If you're still using `setup.py develop` to get an editable build (useful in QuTiP because skipping pip's venvs makes Cython development a little more convenient), then setuptools won't see `pyproject.toml` (because you ran `setup.py` directly), and having the version requirements also available there _can_ be a useful check.  For you, you just need `pyproject.toml`.

The `install_requires` key in `setup.cfg` is the "classic" dependencies.  This is what a user actually needs to have installed for your package to work at run-time.

setup.cfg
jakelishman: Yeah, though this version of your package will still be available on `pip` after it is.  We should probably pin `QuTiP` to the current major releases of Numpy and SciPy as well.

pyproject.toml
hodgestar: Historical note: `requirements.txt` was in vogue for awhile some years ago because it provided a means to provide dependencies in a way that gave access to all of pip's capabilities and didn't require executing potentially complex Python code (e.g. `setup.py`). It's now been superseded by `pyproject.toml` although it's still useful when one wants to provide a list of dependencies without any other functionality for some reason (e.g. a `requirements-doc.txt` might specify the packages needed to build documentation, but not be part of any bigger packaging process).

hodgestar (approved):
Looks great. Don't have any comments beyond Jake's.

PR #24, opened 2021-05-06
Make test_pulse consistent with pytest

purva-thakre:
The tests in `test_pulse` were not run when tested with `pytest test_pulse.py`. 

The new changes make sure pytest runs the tests instead of skipping them.


BoxiLi:
Thanks for spotting this!

PR #28, opened 2021-05-09
Unify the unit for different processors

BoxiLi:
Unify the unit of parameters. All the parameters should be given in the unit of frequency (GHz, MHz etc), and will be saved in the same unit (instead of been multiplied by 2pi as before). The operators are multiplied by 2 pi to produce the correct result.


src/qutip_qip/device/spinchain.py
```diff
@@ -52,44 +52,53 @@ class SpinChain(ModelProcessor):
...
-        each qubit. A list of size `num_qubits` or a float for all qubits.
+        each qubit. A list of size ``num_qubits`` or a float for all qubits.
+
+    **params:
+        Keyword argument for hardware parameters, in the unit of frequency
+        (MHz, GHz etc, the unit of time list needs to be adjusted accordingly).
+        Qubit parameters can either be a float or an array of the length
+        ``num_qubits``.
+        ``sxsy``, should be either a float or an array of the length
+        ``num_qubits-1`` (for :class:`.LinearSpinChain`) or ``num_qubits``
```
nathanshammah, line 82:
```suggestion
        ``num_qubits-1`` (for :class:`.LinearSpinChain`) or ``num_qubits``
```
```suggestion
        ``num_qubits=1`` (for :class:`.LinearSpinChain`) or ``num_qubits``
```

src/qutip_qip/device/spinchain.py
```diff
@@ -283,62 +264,33 @@ def topology_map(self, qc):
...
-        The interaction strength for each of the qubit pair in the system.
-
-    t1: list or float, optional
-        Characterize the decoherence of amplitude damping for
-        each qubit.
-
-    t2: list of float, optional
-        Characterize the decoherence of dephasing for
-        each qubit.
+    Spin chain model with circular topology. See :class:`.SpinChain`
```
nathanshammah, line 267: Include the `Parameters` blurb though?

src/qutip_qip/device/cavityqed.py
```diff
@@ -75,32 +75,31 @@ class DispersiveCavityQED(ModelProcessor):
...
 
+    **params:
+        Keyword argument for hardware parameters, in the unit of GHz.
+        Qubit parameters can either be a float or a list of the length
+        ``num_qubits``.
+
+        - ``deltamax``: the pulse strength of sigma-x control, default ``1.0``
+        - ``epsmax``: the pulse strength of sigma-z control, default ``9.5``
+        - ``eps``: the epsilon for each of qubit, default ``10``
+        - ``delta``: the delta for each of qubit, default ``0.0``
```
nathanshammah, line 94:
```suggestion
        - ``eps``: the epsilon for each of qubit, default ``10``
        - ``delta``: the delta for each of qubit, default ``0.0``
```
Can you add something more descriptive here for `eps` and `delta`?

nathanshammah (changes requested):
Added a couple of comments. Can you add some information about the `\pi` convention somewhere?

src/qutip_qip/device/spinchain.py
BoxiLi: No, it is `num_qubits-1`, for LinearSpinChain the list of coupling strength is one shorter than the number of qubits

PR #29, opened 2021-05-10
Fix bugs in circuitqed

BoxiLi:
ZZ cross talk was wrongly commented.
The initial value of anharmonicity was set wrong.

### DIFF
--- a/src/qutip_qip/device/circuitqed.py
+++ b/src/qutip_qip/device/circuitqed.py
@@ -69,7 +69,7 @@ class SCQubits(ModelProcessor):
                 )[: self.num_qubits]
             ),
             "wr": self.to_array(5.96, num_qubits - 1),
-            "alpha": self.to_array(-1, num_qubits),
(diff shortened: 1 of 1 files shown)


src/qutip_qip/device/circuitqed.py
```diff
@@ -69,7 +69,7 @@ def __init__(
                 )[: self.num_qubits]
             ),
             "wr": self.to_array(5.96, num_qubits - 1),
-            "alpha": self.to_array(-1, num_qubits),
+            "alpha": self.to_array(-0.3, num_qubits),
```
nathanshammah, line 72: Why -0.3? Is this documented?

BoxiLi: It is a commonly used parameter in the experiment. The anharmonicity is between 0.25~0.35GHz, e.g. in the two papers that the dosc refers to. I changed it to `-1` for debugging but forget to change it back.

Issue #32, opened 2021-05-13
Warning on gates when running basic code block from documentation


nathanshammah:
I created a fresh conda environment with python==3.8 and run
pip install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.python.org/simple/ qutip-qip
It installed qutip-qip very fast. Do we have some sort of qutip.about() for qutip.qip? Or some basic test to test the installation?

I ran the code blocks at 
and all worked fine. I got the following Warning when plotting the circuit: https://qutip-qip.readthedocs.io/en/latest/qip-basics.html

```
/opt/miniconda3/envs/qutip-qip/lib/python3.8/site-packages/qutip_qip/circuit.py:274: UserWarning: Unknown gate H
  warnings.warn("Unknown gate %s" % name)
```

BoxiLi:
No, we don't yet have something like `qutip.about()`. Maybe I can import that qutip module and use it.

In qutip-qip the `tests` folder is not packed in the installed version, because it is not in `src`. So I don't think there is a way to do something like `qutip.testings.run()`. The only way to test the code is to download the tests code and run pytest in that folder I think. Maybe @hodgestar knows something about this?

BoxiLi:
About `qutip.about()` I think the easiest way is to modify the `qutip.about()` function, try to find if `qutip-qip` is installed and  display it if it is installed.

hodgestar:
> In qutip-qip the `tests` folder is not packed in the installed version, because it is not in `src`. So I don't think there is a way to do something like `qutip.testings.run()`. The only way to test the code is to download the tests code and run pytest in that folder I think. Maybe @hodgestar knows something about this?

Generally I think keeping the test files outside the package is a good idea (it make the separation between the two very clear & keeps the size of the installed package down). If they're separate then there isn't a good way to run the tests without the source package. I think that's fine -- the time to run the tests is when the wheels are built, I think (i.e. in our GitHub deploy action).

BoxiLi:
This should have already been fixed long ago.

PR #34, opened 2021-05-13
Fix the wrong gate name in the doc

BoxiLi:
The H gate is named SNOT in the current version. Correct the docs. #32 
We cannot really test circuit plotting because it requires latex and lots of others software to work.

At some point, we should support both names. This will be much easier after the refactor of the gate representation. @purva-thakre I think it won't be much work once the gate classes are built, right? We can have `H` as a subclass of `SNOT` or vice versa.


purva-thakre:
@BoxiLi Shouldn't be an issue as long as this is made clear in docs as well. I think `SNOT` should be a sub-class of `H` because `H` gate is popular.

PR #42, opened 2021-05-14
Prepare release v0.1.0 step 2

BoxiLi:
I realized that since this is the first release, I also need to update the README with the installation guide for `pip` install. This README will be shown on the PyPI website after the release just like this one with TestPyPI https://test.pypi.org/project/qutip-qip/.

The doc is not packed with the package so it doesn't really matter, but sooner or later I will need to do this anyway.

Please confirm if this is correct. @nathanshammah @quantshah @jakelishman @hodgestar

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,11 +18,10 @@ Installation
 ------------
 To install the package, download to source code and run
 ```
-pip install .
+pip install qutip_qip
 ```
-under the directory containing the ``setup.py`` file.
 
-If you want to edit the source code, use instead
+If you want to edit the source code, please download the source code and run the following command under the folder with `setup.cfg`
 ```
 pip install -e .
 ```
@@ -33,7 +32,7 @@ To build and test the documentation, additional packages need to be installed:
 pip install matplotlib sphinx numpydoc sphinx_rtd_theme
 ```
 
-Under the docs directory, use
+Under the `docs` directory, use
 ```
 make html
 ```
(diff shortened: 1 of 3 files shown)


jakelishman:
Looks about right to me.  Two things I thought of to check:

- In your `MANIFEST.in`, you don't list `setup.cfg`.  I _think_ it gets included in the sdist automatically by `setuptools`, but it's just worth checking.
- Do you want to specify a `__version__` attribute for your package?  I don't think one is being created, and having it means downstream packages/users can test against the installed version from within Python.

README.md
```diff
@@ -18,11 +18,10 @@ Installation
 ------------
 To install the package, download to source code and run
 ```
-pip install .
+pip install qutip_qip
 ```
-under the directory containing the ``setup.py`` file.
 
-If you want to edit the source code, use instead
+If you want to edit the source code, please download to source code and run the following command under the folder with `setup.cfg`
```
jakelishman, line 24: "download to" -> "download the"

docs/source/installation.rst
```diff
@@ -2,6 +2,14 @@
 Installation
 ************
 
+Quick start
+===========
+To install the package qutip-qip from PyPI, use
```
jakelishman, line 7: maybe ` ``qutip-qip`` `

BoxiLi:
Thanks @jakelishman ! About `setup.cfg`, yes it is included. And a `__version__` is added to `__init__.py`.

jakelishman (approved):

hodgestar (commented):
:+1:

PR #45, opened 2021-05-18
Upload code used in the paper

BoxiLi:


nathanshammah (commented):
Thanks @BoxiLi. Can we add a CI about it? Like a CI in the top `.github` folder that will run this code snippets upon installation with pip. I propose that we create two folders of the files, one `arxiv_1`, the other one that instead is up to date with the current pip version of qutip-qip. In CI, we can set up one or two jobs, i.e. pinning the first one to `qutip_qip` stable and the second one working with the `==0.1.0` version.

nathanshammah:
Note that python 3.6 failed.

nathanshammah:
It's strange it does not show where it fails, on py36 with windows.

BoxiLi:
I think that was GitHub having some bug... The number of tests was also wrong. There should be only 7 tests.

nathanshammah:
Oh, I see. I suggest using `pulse-paper` instead of `paper` for the directory name, it's more precise.

.github/workflows/test.yml
```diff
@@ -49,3 +49,20 @@ jobs:
...
+      with:
+        python-version: 3.9
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install .[full]
+    - name: Test with pytest
+      run: |
+        cd docs
+        make doctest
```
nathanshammah: Why don't we just check that the paper code snippets run?

BoxiLi: That is not the up-to-date version. The following is the new version
```
  test-paper-code:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
    - name: Set up Python
      uses: actions/setup-python@v2
      with:
        python-version: 3.9
    - name: Install dependencies
      run: |
        python -m pip install --upgrade pip
        pip install .[full]
    - name: Test with pytest
      run: |
        cd docs/paper
        mkdir figures
        pytest *.py
```
I need to make a figures directory so that `figsave("figures/...")` works. But maybe `mkdir` is not allowed in GitHub action. Maybe I'll just remove the subfolder `figures` and just use the current folder.

BoxiLi:
GitHub Action is having a problem https://www.githubstatus.com/. We can only wait.

BoxiLi:
Finally, all tests pass. It is best to squash merge this.

.github/workflows/test.yml
```diff
@@ -43,9 +43,15 @@ jobs:
...
+    - name: Test code snippets in the documentation
       run: |
         cd docs
         make doctest
+    - name: Test code examples for the pulse paper
+      run: |
+        python -m pip install joblib pytest pytest-custom_exit_code
+        cd docs/pulse-paper
+        mkdir figures
+        pytest *.py --suppress-no-test-exit-code
```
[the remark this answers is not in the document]
BoxiLi: It runs the snippets, if any of the code produces an error, it will fail, as shown in many of my previous attempts, e.g., https://github.com/qutip/qutip-qip/runs/2614682544?check_suite_focus=true 

The `--suppress-no-test-exit-code` option exists because `pytest` doesn't treat any of the code snippets as a formally defined tests (There is no assert anyway). It just runs through the code and ends if everything is fine.

PR #52, opened 2021-06-11
Adds single qubit decomposition functions

purva-thakre:
WIP for 4 single qubit decomposition functions. 

**To do before marking PR ready for review :**

- [ ] Check format --> black
- [ ] Check style --> PEP8


tests/test_general_decompositions.py
```diff
@@ -0,0 +1,70 @@
...
+    """When an empty array is provided as input."""
+    with pytest.raises(ValueError, match="An empty array was provided as input."):
+        normalize_matrix(np.array([]))
+
+
+def test_normalize_matrix_non_square_matrix():
+    """Test error is raised when row number and column number of Matrix
+    are mismatched."""
+    with pytest.raises(ValueError, match="Input must be a square matrix to be a valid gate."):
+                    normalize_matrix(np.array([[1,4,3]]))
```
purva-thakre, line 63: There is an issue with using `@pytest.mark.parametrize` for this test because pytest reads a list of numpy arrays as an array itself. Understand how pytest could access each array separately and change code accordingly.

tests/test_general_decompositions.py
```diff
@@ -0,0 +1,70 @@
...
+                    normalize_matrix(np.array([[1]]))
+                    normalize_matrix(np.array([[1, 2], [3, 4],[5,6]]))
```
purva-thakre, line 65: There is an issue with using `@pytest.mark.parametrize` for this test because pytest reads a list of numpy arrays as a list. Understand how pytest could iterate over each array and change code accordingly.

BoxiLi: The following seems to work for me:
``` python
@pytest.mark.parametrize("state",
    [np.array([[1,4,3]]),
        np.array([[1]]),
        np.array([[1, 2], [3, 4],[5,6]])
    ])
def test_normalize_matrix_non_square_matrix(state):
    """Test error is raised when row number and column number of Matrix
    are mismatched."""
    with pytest.raises(ValueError, match="Input must be a square matrix to be a valid gate."):
        normalize_matrix(state)
```

Notice that the test in the current version is actually not working in an ideal way. It will succeed if *any* of the call raises a `ValueError`, not *all* of them. And the array `np.array([[1]])` is actually not raising an error, which fails the test in the above test.

purva-thakre: Thanks ! I was actually creating a list of arrays via `()` instead of an array of arrays (like what you did). Most of the examples in [documentation](https://docs.pytest.org/en/6.2.x/parametrize.html) are using `[]` instead of `()`. I guess I just got lucky with my other examples. 😅

src/qutip_qip/decompositions/general_decompositions.py
```diff
@@ -0,0 +1,134 @@
...
+    try:
+        input_array = normalize_matrix(input_array)
+    except ZeroDivisionError:
+        input_array = input_array
+    identity_matrix = np.eye(input_array.shape[0])
+    input_array_dagger = input_array.conj().T
+    check_unitary_left = np.allclose(identity_matrix, np.dot(input_array_dagger,input_array))
+    check_unitary_right = np.allclose(identity_matrix, np.dot(input_array,input_array_dagger))
+    if check_unitary_left != check_unitary_right:
+        raise ArithmeticError("Unitary product assertions do not match.")
```
purva-thakre, line 89: Line not covered by tests. Was added as an extra check to make sure the calculated products match.  
Any special cases ? Or use `no cover ` ?

src/qutip_qip/decompositions/general_decompositions.py
```diff
@@ -0,0 +1,134 @@
...
+        if input_matrix_rows !=input_matrix_columns:
+            raise ValueError("Input must be a square matrix to be a valid gate.")
+
+        if np.linalg.det(input_array) != 1:
+            if np.linalg.det(input_array) ==0:
+                raise ZeroDivisionError("Determinant of matrix =0.")
+            norm_factor = float(1/np.abs(np.linalg.det(input_array)))**0.5
+            input_array = np.around(norm_factor*input_array,5)
+        else:
+            input_array = input_array
```
purva-thakre, line 68: This line is currently not covered by tests according to code coverage. It is making sure the input to the function is assigned to the same variable when the input matrix does not have to be normalized. There's actually [a test](https://github.com/qutip/qutip-qip/pull/52/files#diff-2574ccab83f2a5c033e37ed87e8dc34d29ccb2bd78baebc32fd21bfda9b2b0a8R81) for this scenario with a normalized Pauli Z matrix but is still highlighted in red in code coverage report.

src/qutip_qip/decompositions/general_decompositions.py
```diff
@@ -0,0 +1,134 @@
...
+def check_unitary(input_array)-> bool:
+    """Checks if the input matrix is unitary or not.
+
+    Args:
+       input_array : Matrix of a gate in Numpy array form.
+    """
+    try:
+        input_array = normalize_matrix(input_array)
+    except ZeroDivisionError:
+        input_array = input_array
```
purva-thakre, line 83: Don't think this exception has to be added. If this function's goal is to check if the matrix is unitary or not then a matrix whose determinant is zero is not a unitary matrix.

src/qutip_qip/decompositions/general_decompositions.py
```diff
@@ -0,0 +1,134 @@
...
+    and some other special unitary array).
+
+    Args:
+       input_array : Matrix of a gate in Numpy array form.
+    """
+    if check_unitary(input_array) is True:
+        try:
+            input_array = normalize_matrix(input_array)
+        except ZeroDivisionError:
+            input_array = input_array
```
purva-thakre, line 105: Same as above. If the determinant is zero then do not need to make this exception. `det(U) = 1`

src/qutip_qip/decompositions/general_decompositions.py
BoxiLi: Yeah you can return `False` already.

However, I'm not sure if normalizing is needed. It seems very unlikely to me that a not normalized gate will be unitary after the normalization. If someone provides something with `|det| != 1`, he/she already missed the point of providing a unitary.

Or is there a special reason?

src/qutip_qip/decompositions/verify_decompositions.py
```diff
@@ -0,0 +1,34 @@
...
+#    HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+#    SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+#    LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+#    DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+#    THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+#    (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+#    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+###############################################################################
+
+# add functions to check the calculated decomposition is equivalent to input array.
```
BoxiLi, line 34: If you mean checking whether the decomposed list of matrices is correct after the decomposition, I think it should be in the tests and not here. If we provide something for the user, we should make sure in advance that it is correct by testing it ourselves.

Unless there are some round-off error or special reasons that we cannot tell ahead if the decomposition will succeed?

src/qutip_qip/decompositions/general_decompositions.py
```diff
@@ -0,0 +1,161 @@
...
+    number_of_qubits: int
+        Returns an integer value for the number of qubits
+    """
+    input_check_bool = check_input(input_gate)
+    if input_check_bool == True:
+        input_shape = input_gate.shape
+        if input_shape[0]%2==0.0: # won't be 2 for higher qudits
+        # TO DO : if d=4 (qu4it) then it will still appear as a d=2 i.e. qubit gate
+        # change to powers of 2
+            number_of_qubits = np.log(input_shape[0])/np.log(2)
```
purva-thakre, line 96: @hodgestar @BoxiLi Is there a better way to find the number of qubits based on the size of the input gate matrix ? Here, with what I have chosen to do, if the qudits are of dimensions that are power of 2 then my condition is valid for both qubit and said qudit.

purva-thakre: @BoxiLi @hodgestar I edited my previous comment. I think I have a better idea now. It's possible my understanding of `dims` is wrong.

BoxiLi: Is there a reason that we need to consider qudits/qutrits? I thought the whole decomposition module assumes that we are working with qubits? Or is there anything also useful for qudits?

This kind of information is much easier to extract from the `Qobj`, before taking the full numpy matrix. A 2-qubit ket state has `dims=[[2, 2], [1, 1]]`, while a qudit ket `dims=[[4], [1]]`.

BoxiLi: For unitaries acting on them: `dims=[[2, 2], [2, 2]]` and `dims=[[4], [4]]`

purva-thakre: > Is there a reason that we need to consider qudits/qutrits?

Not really. As the user can input any n x n unitary, I added this extra check to make sure the unitary acts only on qubits and not on other dimensions.

BoxiLi: Ok, then this is just a validity check. Since the input is `Qobj`, using `dims` is easier. We can check if it is a list of `2`.
```
input_gate.dims[0] == [2] * len(input_gate.dims[0])
```

purva-thakre: Yeah, that's my problem. `dims` cannot distinguish between a 4 x 4 input. It could be an object acting on 2 qubits or a qudit of d = 4. 

```
q = Qobj([[0., 0, 0, 0.],[0., 0, 0, 0.],[0., 0, 0, 0.],[0., 0, 0, 0.]])
q.dims = [[4], [4]]
```
I think I am going to change the input to the function. In addition to the input Qobj, they will also need to input number of qubits. This could help avoid the ambiguity.

BoxiLi: Well, technically speaking, the example you provided is interpreted in QuTiP as a unitary acting on a *qudit* system, not two qubits.
E.g.
```
q = Qobj([[0., 0, 0, 0.],[0., 0, 0, 0.],[0., 0, 0, 0.],[0., 0, 0, 0.]], dims = [[4], [4]])
q * basis([2,2],[0,0])
```
will raise an error because the dimension does not match. This is how QuTiP distinguish them. If the user is meant to provide an operator acting on qubits, the dimension should match.

BoxiLi: That said, in decomposition we can also be lenient and allow the user to just produce a 2^n times 2^n `Qobj`.

So I see two choices:
- We make it strict and all the input must have the dimension `[[2,2,2,...],[2,2,2,...]`
- We allow it to be ambiguous a bit and require that `input.shape[0]` is a power of 2, by some smart bit manipulation: https://stackoverflow.com/questions/57025836/how-to-check-if-a-given-number-is-a-power-of-two

purva-thakre:
Closing this PR because I want to start on a new branch. When trying to rebase this branch, there were a lot of conflicts.

PR #53, opened 2021-06-13
Updates `readme.md`

purva-thakre:
Updates `readme` with information about formatting and style. I couldn't find any information about this in qutip's documentation or was looking at the wrong pages.  

This instructs a user to install [`Black`](https://black.readthedocs.io/en/stable/) and ways to format their code. The defaults set by `Black` have not been changed but could be edited.If another package is preferred, then I can change the lines for preferred package. 

I do think there needs to be additional information regarding the use of `Black` in documentation and/or test files. Will add this after I get a feedback on this initial PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,14 @@ pip install -e .
 ```
 which makes sure that you are up to date with the latest `pip` version.
 
+In order to check if your code in `some_file.py` follows `PEP8` style guidelines, [*Black*](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html) has to be installed.
+```
+pip install black
+python -m black some_file.py --diff --color
+python -m black some_file.py
+```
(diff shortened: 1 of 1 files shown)


purva-thakre:
Actually, I think it's better to close this for now. 

There's a lot of files that need to be re-formatted. It's better to create a separate issue to handle the errors one by one + verify these changes are needed. Then maybe this information could be added after `black` doesn't give any errors in existing files.

BoxiLi:
Yeah, there are many files that are not in the ideal pip8 format at the moment. We can add it as a suggestion for contributors, not a requirement.

If we slowly require all our new code to be pip8 formated. Things will get better.

And if we were to add this, it is better to add it to the `contributing to code` in the doc, I think, rather than in readme.

PR #58, opened 2021-06-14
Update contribution guidelines

purva-thakre:
Adds some information about running local tests, generating local code coverage and making sure [new code is PEP8 compliant](https://github.com/qutip/qutip-qip/pull/53#issuecomment-860247855). 

Wanted to add generating local code coverage because this helps avoid pushing to github every time there's a need to check for code coverage.


BoxiLi:
Thanks for this contribution guide on style and coverage! @purva-thakre

Issue #60, opened 2021-06-17
Make changes to installation order for circuit diagrams


purva-thakre:
As stated [here](https://qutip-qip.readthedocs.io/en/latest/qip-basics.html#plotting-a-quantum-circuit), ImageMagick and latex have to be installed in order to get the circuit diagrams.  

When trying to install ImageMagick, conda gets stuck trying to solve the environment (force stopped the process after +6 hours). The environment has all of the development dependencies for `qutip-qip` as well as the ones for documentation.

BoxiLi:
Ok... I'm actually using a Windows installer so never tried that myself. Are you using `doc\requrements.txt` which fixes all the versions? If you are, you could try to start from a fresh environment and install packages like `numpydoc sphinx_rtd_theme` for doc without using `doc\requrements.txt`, as stated [here](https://qutip-qip.readthedocs.io/en/latest/installation.html#prerequisites). Then hopefully conda will find a resolved environment and update `doc\requrements.txt` accordingly.

I'm not using `pdflatex` in conda but a separate TeX library manager. Could that be the problem?

purva-thakre:
I think `pdflatex` might be the problem. The package is old and has not been updated in quite a [while](https://pypi.org/project/pdflatex/). The dependencies for this are older and they clash with newer dependencies installed for the development environment. 

One of the possible workarounds for it is to install[ full TeX package](https://tex.stackexchange.com/q/112734/203959). It's downloading now and will check later if this works. I think we might need to edit the instructions to at least have a lighter version of `texlive` or something else.

purva-thakre:
`pdflatex` is actually not the issue. With a full `texlive` installation, there were no dependency errors. 

There's some incompatibility with `imagemagick` dependencies. Yesterday, the installation for `imagemagick` got stuck at below before I force stopped the process. 
```
Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: failed with repodata from current_repodata.json, will retry with next repodata source.
Collecting package metadata (repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: - 

```
Today, it's trying to resolve the errors as shown below : 
```
Solving environment: - 
Found conflicts! Looking for incompatible packages.
This can take several minutes.  Press CTRL-C to abort.
Examining conflict for tornado pygments matplotlib pyparsing python_abi sqlite

```

[...]

PR #61, opened 2021-06-18
Remove vendored qcircuit.tex

jakelishman:
This file included GPL'd code, which is incompatible with QuTiP's 3-clause BSD licence.  Instead, we rely on the user to have a functional install of the file (it's been included in TeXLive since at least 2014).

This also attempts to maintain patches made to the LaTeX code.

See qutip/qutip#1579

This is the same PR as qutip/qutip#1580, effectively.

### DIFF
--- a/src/qutip_qip/circuit_latex.py
+++ b/src/qutip_qip/circuit_latex.py
@@ -41,7 +41,8 @@ import warnings
 
 _latex_template = r"""
 \documentclass{standalone}
-%s
+\usepackage[braket]{qcircuit}
+\renewcommand{\qswap}{*=<0em>{\times}}
 \begin{document}
 \Qcircuit @C=1cm @R=1cm {
 %s}
@@ -227,9 +228,17 @@ if _pdflatex is not None:
             try:
                 os.chdir(temporary_dir)
                 with open(filename + ".tex", "w") as file:
-                    file.write(_latex_template % (_qcircuit_latex_min, code))
-                _run_command((_pdflatex, '-interaction', 'batchmode',
-                              filename))
+                    file.write(_latex_template % code)
+                try:
+                    _run_command((_pdflatex, '-interaction', 'batchmode',
+                                  filename))
+                except RuntimeError as e:
+                    message = (
+                        "pdflatex failed."
+                        " Perhaps you do not have it installed, or you are"
+                        " missing the LaTeX package 'qcircuit'."
+                    )
+                    raise RuntimeError(message) from e
                 _crop_pdf(filename + ".pdf")
                 if file_type in _MISSING_CONVERTERS:
                     dependency = _MISSING_CONVERTERS[file_type]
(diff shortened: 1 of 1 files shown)


src/qutip_qip/circuit_latex.py
```diff
@@ -227,9 +228,17 @@ def image_from_latex(code, file_type="png"):
...
-                              filename))
+                    file.write(_latex_template % code)
+                try:
+                    _run_command((_pdflatex, '-interaction', 'batchmode',
+                                  filename))
+                except RuntimeError as e:
+                    message = (
+                        "pdflatex failed."
+                        " Perhaps you do not have it installed, or you are"
+                        " missing the LaTeX package 'qcircuit'."
```
purva-thakre, line 239: @jakelishman Thanks for this ! `qcircuit` not installed was one of the issues I was facing in #60.

jakelishman: Just having a flick through #60, it looks like you might be looking for too much on `conda` - `imagemagick` is something you'd typically expect to find in a system package manager, rather than `conda`.  On Debian-like Linuxes that would be likely be `apt` (`apt[-get] install imagemagick`), or `brew install imagemagick` on Mac.

`qcircuit` itself should be present in any fully featured TeXLive installation.  If you've just got a minimal TeXLive it might not have it, but you can use `tlmgr` to install it.

jakelishman:
This shouldn't logically affect coverage by a significant amount, because I don't believe this part of the production is tested anyway.  Getting a proper installation of TeXLive running on GitHub Actions CI will take quite a long time - TeXLive Full 2021 is something like 5GB, or I suppose you could install TeXLive core, and then `tlmgr` install the various bits of `amsmath`, `braket` and `qcircuit` as necessary?

@BoxiLi: did you consider setting a threshold on Coveralls' reports of failures?  There's an option somewhere around to have it only report failures if they're greater than a given percent.

src/qutip_qip/circuit_latex.py
purva-thakre: > qcircuit itself should be present in any fully featured TeXLive installation

I had a minimal install with `qcircuit` but not all of it's dependencies were installed. I used [this](https://pypi.org/project/latextools/). 

> imagemagick is something you'd typically expect to find in a system package manager

How would you typically know if some dependency could be found in a package manager instead of `conda` ?

jakelishman: That `latextools` package seems to be more about building and manipulating LaTeX source in Python, rather than actually providing the LaTeX distribution - it still needs TeXLive or something.

Conda often has more than I expect it to, but generally command-line tools unrelated to Python would be installed from the system package manager (because they're system packages). Conda does blur the lines a little bit (especially when it comes to BLAS), but `imagemagick` and `pdflatex` would normally be system commands.

jakelishman: TeXLive comes with `tlmgr` (its own package manager for LaTeX files), so a minimal installation for your purposes here would probably be
```bash
$ apt-get install imagemagick pdfcrop pdf2svg
$ apt-get install texlive  # fairly minimal version, but I think it has amsmath and amsfonts
$ tlmgr install braket qcircuit
```
(granted I'm guessing a little, but that's the gist)

purva-thakre: Thanks !  I will test this in a new virtual environment when I update the documentation. 

I used `texlive` as well but I had to go on a search for a couple of more `latex` packages based on the errors.

BoxiLi (approved):
Tested with the Jupyter notebook of https://nbviewer.ipython.org/github/qutip/qutip-notebooks/blob/master/examples/quantum-gates.ipynb.

LGTM. I shall merge if @purva-thakre find the same on Ubuntu

BoxiLi:
> did you consider setting a threshold on Coveralls' reports of failures? 

Yes, I remember you mentioned this once somewhere during the unitary hack. Haven't implement it yet. That is a good one to include. 0.1% decrease is no failure. It is just because the total number of code is decreasing. Is that option somewhere already implemented in `qutip`?

Testing with whole tex library is too heavy indeed.

jakelishman:
Since this file isn't covered, I assume it's actually that the total number of lines went _up_ rather than down.  These tools are usually smart enough to call a multiline string one "logical line", but in removing it, I added an extra try/catch block which adds 3 or 4 more lines in total.

purva-thakre:
> LGTM. I shall merge if @purva-thakre find the same on Ubuntu

@BoxiLi Once #63 is merged, the linked notebook is able run all the code blocks properly.

jakelishman:
By the way, given that SVGs are just text-based, there's probably an approximate way to test some of this functionality - I could imagine a test that constructs a circuit injecting some dummy text into a node somewhere, then calls `circuit._repr_svg()`, and asserts that `dummy_text in svg and "svg" in svg`.  That's harder to do for the PNG file, but plausible.  Of course, that relies on getting a decent LaTeX distribution installed on a CI machine.

PR #63, opened 2021-06-21
Fixes gate label import

purva-thakre:
Fixes import of gate labels when gates were moved to operations folder. See related [comment.](https://github.com/qutip/qutip-qip/issues/60#issuecomment-863828529) 

Tested with [examples in this notebook](https://nbviewer.ipython.org/github/qutip/qutip-notebooks/blob/master/examples/quantum-gates.ipynb#Controlled-PHASE). 

The notebook also needs a separate PR to change how gates are imported + deprecation warning in `In [62]`. This will be added later. 

Following code block is for future reference in relation to `In [62]` : 

``` python

<ipython-input-60-0c4e722f5dae>:3: DeprecationWarning: `np.complex` is a deprecated alias for the builtin `complex`. To silence this warning, use `complex` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.complex128` here.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
  mat = np.zeros((4, 4), dtype=np.complex)
```

### DIFF
--- a/src/qutip_qip/operations/gates.py
+++ b/src/qutip_qip/operations/gates.py
@@ -57,7 +57,7 @@ __all__ = ['Gate', 'rx', 'ry', 'rz', 'sqrtnot', 'snot', 'phasegate', 'qrot',
            'gate_expand_1toN', 'gate_expand_2toN', 'gate_expand_3toN',
            'qubit_clifford_group', 'expand_operator', '_single_qubit_gates',
            '_para_gates', '_ctrl_gates','_swap_like','_toffoli_like',
-           '_fredkin_like']
+           '_fredkin_like','_GATE_NAME_TO_LABEL','_gate_label']
 
 
 _single_qubit_gates = ["RX", "RY", "RZ", "SNOT", "SQRTNOT", "PHASEGATE",
--- a/src/qutip_qip/circuit.py
+++ b/src/qutip_qip/circuit.py
@@ -49,7 +49,7 @@ from .operations import (Gate, rx, ry, rz, sqrtnot, snot, phasegate,
                                         iswap, sqrtswap, sqrtiswap, fredkin,
                                         toffoli, controlled_gate, globalphase,
                                         expand_operator, gate_sequence_product,
-                                        ,'_GATE_NAME_TO_LABEL','_gate_label')
+                                        _GATE_NAME_TO_LABEL,_gate_label)
 from qutip import tensor, basis, identity, ket2dm, qeye
 from qutip.qobj import Qobj
 from qutip.measurement import measurement_statistics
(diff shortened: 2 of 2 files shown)


src/qutip_qip/operations/gates.py
```diff
@@ -57,7 +57,7 @@
            'gate_expand_1toN', 'gate_expand_2toN', 'gate_expand_3toN',
            'qubit_clifford_group', 'expand_operator', '_single_qubit_gates',
            '_para_gates', '_ctrl_gates','_swap_like','_toffoli_like',
-           '_fredkin_like']
+           '_fredkin_like','_GATE_NAME_TO_LABEL','_gate_label']
```
hodgestar, line 60: We shouldn't add these `_` constants and functions to `__all__` I don't think and we don't need to import `_GATE_NAME_TO_LABEL` in `circuit.py` at all.

hodgestar:
@purva-thakre Thank you for noticing this. I came across it independently and addressed it in a slightly different in #66. If #66 is merged, I'll close this PR. Apologies for not finding this PR before I fixed the bug in mine.

hodgestar:
#66 also adds a test for `.latex_code()`.

src/qutip_qip/operations/gates.py
purva-thakre: > we don't need to import _GATE_NAME_TO_LABEL in circuit.py at all

We do actually. It's used in `circuit.py`. I tried running one of the notebooks with `_gate_label` and that caused some errors.

purva-thakre:
> @purva-thakre Thank you for noticing this. I came across it independently and addressed it in a slightly different in #66. If #66 is merged, I'll close this PR. Apologies for not finding this PR before I fixed the bug in mine.

No problem @hodgestar

src/qutip_qip/operations/gates.py
hodgestar:
```
$ ag _GATE_NAME_TO_LABEL .
src/qutip_qip/operations/gates.py
171:            if (name in _GATE_NAME_TO_LABEL) and (arg_value is not None):
226:_GATE_NAME_TO_LABEL = {
263:    if name in _GATE_NAME_TO_LABEL:
264:        gate_label = _GATE_NAME_TO_LABEL[name]
```
I'm not seeing it used anywhere other than in `gates.py`? Could you provide a link to where else it is used?

purva-thakre: Sorry, I was confused between `_gate_label` and `_GATE_NAME_TO_LABEL`. You are right. Only `_gate_label` has to be imported. 

Although, for this issue[ Boxi wanted a separate PR](https://github.com/qutip/qutip-qip/issues/60#issuecomment-864374124). Feel free to close this PR if that's changed.

BoxiLi: Oh, I was asking for a separate PR mainly because the decomposition is a new feature and this is a bug fix. Mixing them might cause some confusion if someone checks the history one day. As #66 is all about bug fixing and cleaning, I can just merge that one and saves you the trouble of updating it for `_GATE_NAME_TO_LABEL`. But thank you anyway!

PR #67, opened 2021-06-28
Adds single qubit decomposition functions

### LINKED ISSUE #73: Understand issue with global phase factor function
The output of Pauli Y is off by a global phase of -1 when decomposed with either of the single qubit decomposition functions. Understand what needs to be fixed for this - missing minus sign ?, not  use determinant of input to find the global phase ?

If there's a common multiplication factor for each element of the matrix, the function fails to remove it.

purva-thakre:
Adds general single qubit decomposition functions. 

- There's only 2 combinations of `decompose_to_rotation_matrices` because the calculations for others need to be re-calculated. The outputs were not agreeing with input gates.   Will come back to these after a general method for n qubits has been added. 
- Same as above, there's only 1 combinations for `ABC_decomposition` and others will be added towards the end. 
- The issue with `extract_global_phase` function is still giving an output off by a phase. In the example notebook, all the outputs are the same as input except for `sigmay`. As I was able to use `average_gate_fidelity` to compare the outputs even when they are off by a phase, I plan to come back to this towards the end. Too many days were spent trying to fix this. 

Initially, most of the outputs except for sigmay` were off by a phase_factor of -1. The problem was solved by changing[ the global phase angle. ](https://github.com/qutip/qutip-qip/pull/67/files#diff-6367c4af96a402daf1a20673bc8e9eb34b0390fd18dc119373485b5105f3b9d4R31)

decomposition_pauliy_off_

Imports, style etc will be corrected after final approval. 

Fixes #73


src/qutip_qip/decompositions/decompositions_extras.py
```diff
@@ -0,0 +1,26 @@
+from qutip_qip.decompositions.general_decompositions import (check_input,
+check_input_shape, convert_qobj_gate_to_array)
+
+from qutip_qip.decompositions.single_decompositions import (_ZYZ_rotation, ABC_decomposition)
+from qutip import *
```
BoxiLi, line 5: In general, one should avoid star importation in a package. `qutip` is a very large namespace. It includes a lot of things in the namespace that you may not be aware of.

Personally, I find it ok to do this at the early stage of development while exploring different things. But should be removed before merging.

src/qutip_qip/decompositions/general_decompositions.py
```diff
@@ -0,0 +1,113 @@
...
+
+    """
+    input_check_bool = check_input(input_gate)
+    if input_check_bool == True:
+        input_shape = input_gate.shape
+        try:
+            assert input_shape[0] == 2**num_of_qubits
+        except AssertionError:
+            return(False)
+        return(True)
```
BoxiLi, line 67: Why not just `return assert input_shape[0] == 2**num_of_qubits`?

src/qutip_qip/decompositions/general_decompositions.py
```diff
@@ -0,0 +1,113 @@
...
+    ----------
+    input_gate : :class:`qutip.Qobj`
+        The matrix that's supposed to be decomposed should be a Qobj.
+
+    """
+    input_array = convert_qobj_gate_to_array(input_gate)
+    determinant_of_input = np.linalg.det(input_array)
+    y = np.imag(determinant_of_input)
+    x = np.real(determinant_of_input)
+    global_phase_angle = np.arctan2(y,x)
```
BoxiLi, line 111: Is it possible that the problem is here? Since I remember you mentioned a few times about using `cmath.phase` instead of `arctan`.

src/qutip_qip/decompositions/single_decompositions.py
```diff
@@ -0,0 +1,191 @@
...
+        Rz_C = Gate("RZ",targets=[0], arg_value=(-alpha+beta)/2, arg_label=r'{:0.2f} \times \pi'.format((-alpha_string+beta_string)/2))
+
+        return(Rz_A, Ry_A, Pauli_X, Ry_B, Rz_B, Pauli_X, Rz_C, Phase_gate)
+
+    else:
+        raise GateError("Shape of input does not match with the number of qubits.")
+
+_rotation_matrices_dictionary ={"ZYZ": _ZYZ_rotation}
+
+def product_of_rotation_matrices(input_gate, num_of_qubits, method):
```
BoxiLi, line 127: Maybe `decompose_to_rotation_matrices`?

src/qutip_qip/decompositions/single_decompositions.py
```diff
@@ -0,0 +1,191 @@
...
+        Ry_A = Gate("RY",targets=[0], arg_value=theta/2, arg_label=r'{:0.2f} \times \pi'.format(theta_string/2))
+        Pauli_X = Gate("X",targets=[0])
+        Ry_B = Gate("RY",targets=[0], arg_value=-theta/2, arg_label=r'{:0.2f} \times \pi'.format(-theta_string/2))
+        Rz_B = Gate("RZ",targets=[0], arg_value=-(alpha+beta)/2, arg_label=r'{:0.2f} \times \pi'.format(-(alpha_string+beta_string)/2))
+        Rz_C = Gate("RZ",targets=[0], arg_value=(-alpha+beta)/2, arg_label=r'{:0.2f} \times \pi'.format((-alpha_string+beta_string)/2))
+
+        return(Rz_A, Ry_A, Pauli_X, Ry_B, Rz_B, Pauli_X, Rz_C, Phase_gate)
+
+    else:
+        raise GateError("Shape of input does not match with the number of qubits.")
```
BoxiLi, line 123: It would be nice to keep things in one place. You already raise many errors in check functions, only this one is raised here. I think it would be nice to either raise all errors there or only return `False` or `True`, never raise.

src/qutip_qip/decompositions/single_decompositions.py
```diff
@@ -0,0 +1,191 @@
...
+convert_qobj_gate_to_array, extract_global_phase)
+
+from qutip_qip.operations import *
+from qutip_qip.circuit import QubitCircuit, Gate
+
+class MethodError():
+    pass
+
+class GateError():
+    pass
```
BoxiLi, line 15: Inherit from Python `Exception`?

src/qutip_qip/decompositions/general_decompositions.py
purva-thakre: I think this was because it was not returning `False` when `AssertionError` occurred. I might need to double check.

src/qutip_qip/decompositions/general_decompositions.py
purva-thakre: Not for `arctan2`. `arctan` and `arctan2` are two different functions in numpy. 

I had made a mistake by trying to use `arctan`. 

I think the issue with some of the outputs being off by some global phase could be the lack of a normalization constant (or division by absolute value). I did think of this initially but since I check if the matrix is unitary before extracting the phase, I didn't think I had to have it.

src/qutip_qip/decompositions/general_decompositions.py
BoxiLi: Oh my mistake, I meant `return input_shape[0] == 2**num_of_qubits`, no `assert`.

src/qutip_qip/decompositions/single_decompositions.py
purva-thakre: I have tried to do so in the new changes. Let me know if I did not do what I thought was being done.

src/qutip_qip/decompositions/single_decompositions.py
```diff
@@ -0,0 +1,226 @@
...
+    Rz_alpha = Gate("RZ",targets=[target], arg_value=alpha, arg_label=r'{:0.2f} \times \pi'.format(alpha_string))
+    Rx_theta = Gate("RX",targets=[target], arg_value=theta, arg_label=r'{:0.2f} \times \pi'.format(theta_string))
+    Rz_beta = Gate("RZ",targets=[target], arg_value=beta, arg_label=r'{:0.2f} \times \pi'.format(beta_string))
+
+    return(Rz_alpha, Rx_theta, Rz_beta, Phase_gate)
+
+
+_rotation_matrices_dictionary ={"ZYZ": _ZYZ_rotation,
+                                "ZXZ": _ZXZ_rotation,
+                                }
```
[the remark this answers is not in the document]
BoxiLi: The routine could be like this (just my thought, not that you have to):
- Regardless of the scheme, always use ZYZ decomposition, get the 3 angles and the global phase.
- For each scheme, there is a private function to decide what is the 3 rotation gates to use (RZ, RY or RZ) as well as the global phase correction. They always use the same rotation angles calculated for ZYZ (maybe add a minus sign here and there if needed). This can be manually derived as I sketched above. 

In this way one only need one function to calculate the ZYZ angles, the rest is just tranformation.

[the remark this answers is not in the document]
purva-thakre: Yes, that's what I attempted to do by creating [`_angles_for_ZYZ`](https://github.com/qutip/qutip-qip/pull/67/files#diff-6367c4af96a402daf1a20673bc8e9eb34b0390fd18dc119373485b5105f3b9d4R14). 

I had calculated the angles for the new combinations by using output of ZYZ and transforming Rz as product of RxRyRx and so on..  I made a mistake somewhere in these transformations and the outputs don't match. 

Adding the combinations later down the line is just a matter of re-checking these calculations.

src/qutip_qip/decompositions/single_decompositions.py
```diff
@@ -0,0 +1,226 @@
...
+from qutip.qobj import Qobj
+from qutip_qip.decompositions.general_decompositions import (check_input, check_input_shape,
+convert_qobj_gate_to_array, extract_global_phase, MethodError, GateError)
+
+from qutip_qip.operations import *
+from qutip_qip.circuit import QubitCircuit, Gate
+
+
+# Functions for decompose_to_rotation_matrices
+single_qubit = 1
```
hodgestar, line 13: Should this line be here? Not sure what it is doing. :)

src/qutip_qip/decompositions/decompositions_extras.py
```diff
@@ -0,0 +1,35 @@
...
+        q_circuit = QubitCircuit(num_of_qubits, reverse_states=False)
+        for i in decomposed_gate:
+            q_circuit.add_gate(i)
+        return(q_circuit)
+    else:
+        raise TypeError("Input is not a tuple of gates.")
+
+def matrix_of_decomposed_gates(quantum_circuit):
+    """Evaluates all the gates in the circuit. 
+    """
```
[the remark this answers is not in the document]
BoxiLi: On second thought, maybe not `get_unitary` as get is often used in class attributes, which are easy to obtain, something like `compute_unitary` would be better.

src/qutip_qip/decompositions/__init__.py
```diff
@@ -0,0 +1,7 @@
+from qutip_qip.decompositions.general_decompositions import (check_input,
+check_input_shape, convert_qobj_gate_to_array, extract_global_phase)
+
+from qutip_qip.decompositions.single_decompositions import (_ZYZ_rotation, _ZXZ_rotation, _rotation_matrices_dictionary,
```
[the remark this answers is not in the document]
purva-thakre: I had done this to test the outputs of these particular functions. I will remove it once I no longer have to check their outputs.

src/qutip_qip/decompositions/general_decompositions.py
```diff
@@ -0,0 +1,122 @@
...
+        The matrix that's supposed to be decomposed should be a Qobj.
+
+    Returns
+    -------
+    input_gate : `np.array`
+        The input is returned as a converted numpy array.
+    """
+    input_check_bool = check_input(input_gate)
+    if input_check_bool == True:
+        input_to_array = Qobj.full(input_gate)
```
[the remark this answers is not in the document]
purva-thakre: What's the difference between the two ? I can see both ways work. Is it something to do with the methods of `Qobj` class ?

src/qutip_qip/decompositions/single_decompositions.py
purva-thakre: This was added to differentiate the inputs to a decomposition function when the target and number of qubits was the same. I think, I can try to define this in every functions definition with a default value (like Boxi discusses[ here](https://github.com/qutip/qutip-qip/pull/67#discussion_r664576271)).

src/qutip_qip/decompositions/decompositions_extras.py
```diff
@@ -0,0 +1,35 @@
...
+from qutip.qobj import Qobj
+from qutip_qip.operations import *
+from qutip_qip.circuit import QubitCircuit, Gate
+from numpy import pi
+import numpy as np
+
+def decomposed_gates_to_circuit(decomposed_gate,num_of_qubits):
+    """This function takes the input from a decomposition function and creates
+    a circuit diagram.
+    """
```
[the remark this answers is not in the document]
purva-thakre: Well, no. It does return a `QubitCircuit` object. 

Perhaps the line `creates a circuit diagram.` ought to be removed.

src/qutip_qip/decompositions/general_decompositions.py
```diff
@@ -0,0 +1,122 @@
...
+        try:
+            isinstance(input_to_array, np.ndarray)
+        except: # Not sure if this error has to be included. If an error occurs,
+        # then it ought to covered in the main qutip module.
+            raise ConversionError("Input Qobj could not be converted to a numpy array.")
```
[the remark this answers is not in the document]
purva-thakre: Yes, I know it's not converting it. It's just checking the object was properly converted to a numpy array. 

I was not sure if this extra check was needed. Based on your last statement, I don't think it is needed.

src/qutip_qip/decompositions/general_decompositions.py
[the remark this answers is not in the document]
BoxiLi: Actually, I'm not sure either. Both seems to work indeed. Maybe @hodgestar knows something?

src/qutip_qip/decompositions/decompositions_extras.py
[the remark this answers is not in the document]
BoxiLi: Oh, I think I misread `decomposed` as `decompose`. Anyway, I understand that you add them here just for your convenience in testing. In principle, this is very general can be added to `QubitCircuit`, as in our discussion.

src/qutip_qip/decompositions/general_decompositions.py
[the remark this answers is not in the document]
purva-thakre: There's another example of this in my code [here](https://github.com/qutip/qutip-qip/pull/67/files/196be453208d3ac49753b515055c897a71a695f8#diff-2543f9f156f3ca948a15c356cece94f773845236a0513e2b02efbd3a1643282cR46) : ` unitary_check = Qobj.check_isunitary(input_gate)`

src/qutip_qip/decompositions/decompositions_extras.py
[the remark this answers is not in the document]
purva-thakre: Just to check - do you want these methods to be defined as methods of `QubitCircuit` ? 

Looking over my notes from the meeting, I have only made a note to move them to `circuit.py` 😅

purva-thakre:
@hodgestar @BoxiLi 

Going back to this Monday's discussion of renaming `decompositions` to `decompose` - what's your opinion of renaming `single_decompositions` to `single_qubit_decompose` or `decompose_single_qubit_gate` ?

src/qutip_qip/decompositions/single_decompositions.py
```diff
@@ -0,0 +1,226 @@
...
+
+    Here, :math:`i \neq j` and :math:`i, j \in {x, y, z}`
+
+    .. math::
+
+        U = \begin{bmatrix}
+            a       & b \\
+            -b^*       & a^*  \\
+            \end{bmatrix} = \textrm{R}_i(\alpha)  \textrm{R}_j(\theta)  \textrm{R}_i(\beta)
+
```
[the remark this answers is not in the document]
purva-thakre: > num_of_qubits is actually not needed here because it is by definition single-qubit gates

Yes, it is needed to check if a target is provided for larger number of qubits. Here's the check :

```python
try:
        assert num_of_qubits == 1
except AssertionError:
        if target is None and num_of_qubits >1:
            raise GateError("This method is valid for single qubit gates only. Provide a target qubit for single qubit gate.")
```

BoxiLi:
`decompose_single_qubit_gate` sounds nice to me. We could have unified names for this module:
- `decompose_single_qubit_gate(gate, method, ...)`
- `decompose_two_qubit_gate(gate, method, ...)`
- `decompose_gate(gate, method, ...)`

purva-thakre:
> `decompose_single_qubit_gate` sounds nice to me. We could have unified names for this module:
> 
> * `decompose_single_qubit_gate(gate, method, ...)`
> * `decompose_two_qubit_gate(gate, method, ...)`
> * `decompose_gate(gate, method, ...)`

I think this seems like a good idea. Define a dictionary for all the methods specific to the number of qubits. 
Would you prefer to make functions like `decompose_to_rotation_matrices` and `ABC_Decomposition` be made private ?

doc/source/_apidoc/qutip_qip.decompositions.rst
```diff
@@ -0,0 +1,37 @@
+qutip\_qip.decompositions package
```
purva-thakre, line 1: HTML build failing due to renamed modules. Change before requesting a review.

BoxiLi: Yeah, it should be fairly easy to update this using the command https://qutip-qip.readthedocs.io/en/latest/contribution-docs.html#generate-api-documentation.

purva-thakre:
Changed decomposition functions being a subpackage to a submodule. All other functions in the subpackage have been made private with only `from qutip_qip.decompose import decompose_one_qubit_gate` being publicly available.

The functions `decompose_to_rotation_matrices` and `ABC_decomposition` can be called via strings in `decompose_one_qubit_gate`. 

image


Here's an example screenshot :

image

If it's preferred to keep the functions separate, I will undo last commit.

BoxiLi:
> Changed decomposition functions being a subpackage to a submodule.

Maybe you two discussed this already, but what is the motivation behind this change?

It doesn't (shouldn't) make a difference in the user interface I think because we only expose those public APIs in `decompose\__init__.py`. However, now we have a separate private subpackage and a module that calls that package. Seems a bit confusing to someone looking at the source code.

purva-thakre:
@BoxiLi We discussed some style changes as well as removing target and num_qubits as input to single qubit functions. 

> but what is the motivation behind this change?

I wanted to use an easier way to import the main decomposition functions if they were all in one file. 

Before yesterday's changes, if there was a need to import some function it would be like below : 
```python
from qutip_qip.decompose.main_decomposition_functions_folder_name import decompose_n_qubit gate 
```

After creating `decompose.py` outside the folder containing functions defined for decomposing the gates, the import changes as 
```python
from qutip_qip.decompose import decompose_n_qubit gate 
```

I could keep all of these in the same folder if you would prefer the import path be changed in `__init__.py`. I was unsure if this was a good idea.

BoxiLi:
Ok, I think this is the key point here:
> Before yesterday's changes, if there was a need to import some function it would be like below :
> from qutip_qip.decompose.some_decomposition_functions_folder_name import decompose_n_qubit gate 

If I understand correctly, you were saying that if we use a subpackage structure like this:
```
- qutip_qip
    - decompose
        - __init__.py
        - single_qubit_decompose.py
```
The user can only use `decompose_single_qubit_gate ` by
```
from qutip_qip.decompose.single_qubit_decompose import `decompose_single_qubit_gate`
```
Is that correct?

I don't think the above is true because if you import `decompose_single_qubit_gate` in `__init__.py`, you can already use
```
from qutip_qip.decompose import decompose_single_qubit_gate
```
That is what happening in e.g. `devices` subpackage.

Or did I miss understand you?

purva-thakre:
> Or did I miss understand you?

Not really. There was an issue with my local system path which could only import things as long as the entire path was specified exactly. I will switch back to what it was before. 

Thanks !

Alt Text

purva-thakre:
@hodgestar  @BoxiLi 

I have pushed the new discussed changes. 

Looking back, I don't think we discussed [`decomposed_gates_to_circuit`](https://github.com/qutip/qutip-qip/pull/67/files#diff-489b56a1ae827181e6e743555d12a3c9dca7b6791cd00c4e47c8452cc0fd01ceR2136) and [`compute_unitary`](https://github.com/qutip/qutip-qip/pull/67/files#diff-489b56a1ae827181e6e743555d12a3c9dca7b6791cd00c4e47c8452cc0fd01ceR2155). These functions were defined for convenience. Anyone can also use already defined things in `qip` to create a circuit and calculate the matrix of the gates. Let me know if you prefer to  discard these functions and show the steps to take for both in a tutorial.

src/qutip_qip/decompose/_utility.py
```diff
@@ -0,0 +1,31 @@
...
+
+    pass
+
+
+def check_gate(gate, num_qubits):
+    """Verifies input is a valid quantum gate.
+
+    Parameters
+    ----------
+    gate : :class:`qutip.Qobj`
```
purva-thakre, line 15: remove qutip - only :class:`.Qobj`

src/qutip_qip/decompose/decompose_single_qubit_gate.py
```diff
@@ -0,0 +1,256 @@
...
+            -b^*       & a^*  \\
+            \end{bmatrix} = \textrm{R}_i(\alpha)  \textrm{R}_j(\theta) \textrm{R}_i(\beta) = \textrm{A} \sigma_k \textrm{B} \sigma_k \textrm{C}
+
+    Here,
+
+    * :math:`\textrm{A} = \textrm{R}_i(\alpha) \textrm{R}_j\left(\frac{\theta}{2} \right)`
+
+    * :math:`\textrm{B} = \textrm{R}_j \left(\frac{-\theta}{2} \right)\textrm{R}_i \left(\frac{- \left(\alpha + \beta \right)}{2} \right)`
+
+    * :math:`\textrm{C} = \textrm{R}_i \left(\frac{\left(-\alpha + \beta\right)}{2} \right)`
```
purva-thakre, line 204: Lines too long. Split in math mode.

src/qutip_qip/decompose/decompose_single_qubit_gate.py
```diff
@@ -0,0 +1,256 @@
...
+            :header-rows: 1
+
+            * - Method Key
+              - Method
+            * - ZYZ
+              - :math:`\textrm{R}_z(\alpha)  \textrm{R}_y(\theta) \textrm{R}_z(\beta)`
+            * - ZXZ
+              - :math:`\textrm{R}_z(\alpha)  \textrm{R}_x(\theta) \textrm{R}_z(\beta)`
+            * - ZYZ_PauliX
+              - :math:`\textrm{A} \sigma_k \textrm{B} \sigma_k \textrm{C}` :math:`\forall k =x, i =z, j=y`
```
purva-thakre, line 227: Lines too long. Split in math mode.

BoxiLi:
> Looking back, I don't think we discussed `decomposed_gates_to_circuit` and `compute_unitary`. These functions were defined for convenience. Anyone can also use already defined things in qip to create a circuit and calculate the matrix of the gates. Let me know if you prefer to discard these functions and show the steps to take for both in a tutorial.

They are indeed useful functions, we can surely keep them in the code. However, we can make them a bit more general:
- The name `decomposed_gates_to_circuit` is appropriate in the application here indeed, but the function itself is much more general than this. So we should use a more general name.
- The two functions are closely related to `QubitCircuit`. Following the convention of `QubitCircuit`, we can add them as class methods. This will also save us some input checks.

I made these two minor changes in a commit, feel free to criticize.

purva-thakre:
@BoxiLi Looks good to me !

I could have also made the changes if you would have replied to an earlier comment ! Thanks ! 😸

src/qutip_qip/decompose/_utility.py
BoxiLi: Is there a reason?I'm not sure but there seems to be a way to link  `:class:`qutip.Qobj` to the qutip doc, with some extension. I tried it once but didn't work. That is still on my to-do list.

src/qutip_qip/decompose/decompose_single_qubit_gate.py
BoxiLi: If you find the math looks better this way, we can also keep it, although it avoids PEP8. Personally, I think no one will read it in the source code. It is for the sphinx anyway. Maybe we can later move equations to the main docs if we want to have a proper doc for decomposition.

src/qutip_qip/decompose/decompose_single_qubit_gate.py
```diff
@@ -0,0 +1,256 @@
...
+
+    * :math:`\textrm{B} = \textrm{R}_j \left(\frac{-\theta}{2} \right)\textrm{R}_i \left(\frac{- \left(\alpha + \beta \right)}{2} \right)`
+
+    * :math:`\textrm{C} = \textrm{R}_i \left(\frac{\left(-\alpha + \beta\right)}{2} \right)`
+
+
+    Parameters
+    ----------
+    input_gate : :class:`qutip.Qobj`
+        The matrix that's supposed to be decomposed should be a Qobj.
```
BoxiLi, line 210:
```suggestion
        The matrix to be decomposed.
```
It is clear from the previous line that the matrix should be a Qobj.

BoxiLi:
> I could have also made the changes if you would have replied to an earlier comment ! Thanks ! 😸

Yeah, I definitely believe that you would have made the same changes. I missed that question so failed to reply on time. The PR is in good shape to me. I was considering that we could merge it on Monday if @hodgestar has no further suggestions.

src/qutip_qip/decompose/_utility.py
purva-thakre: I think I need to use `:class:~qutip.Qobj` but I will double check soon.

src/qutip_qip/decompose/decompose_single_qubit_gate.py
purva-thakre: It's up to you. I was trying to keep my code consistent with PEP8 but if you would prefer to ignore the doctsrings, it's ok with me. 

> Maybe we can later move equations to the main docs if we want to have a proper doc for decomposition.

This could work as well.

BoxiLi: Oh, in general, we should keep also docstrings PEP8 compatible. If you find a way to make the math compatible with PEP8 without destroying the readability then please surely do it. I was just saying that if there is no easy way then we can also keep it like this.

purva-thakre: > If you find a way to make the math compatible with PEP8 without destroying the readability then please surely do it.

I don't know an easy way yet.

src/qutip_qip/circuit.py
```diff
@@ -349,6 +348,18 @@ def add_gate(self, gate, targets=None, controls=None, arg_value=None,
                             control_value=control_value)
                 self.gates.insert(position, gate)
 
+    def add_gates(self, gates):
+        """
+        Add a sequence of gates to the circuit.
```
hodgestar, line 353: We should perhaps document the order in which the gates are added -- i.e. that this is equivalent to `for g in gates: self.add_gate(g)`.

src/qutip_qip/circuit.py
```diff
@@ -1486,6 +1497,13 @@ def _propagators_no_expand(self):
 
         return U_list
 
+    def compute_unitary(self):
+        """Evaluates all the gates in the quantum circuit of decomposed gates.
+        """
```
hodgestar, line 1502: We should document what is returned (i.e. the type of the return and a one line description).

hodgestar (approved):
Looking good to me!

src/qutip_qip/circuit.py
purva-thakre: @hodgestar Is there a better way to describe this besides stating `gates are added from a tuple sequentially i.e. from left to right`.

BoxiLi: I think there is no need to specify it to be a tuple because anything iterable will work here. Maybe we can say `The gates will be applied in positive order, i.e., the first gate in the sequence will be applied first to the state.`

BoxiLi (approved):

BoxiLi:
Good job! You can now rebase your changes for two qubits gate decomposition on this. @purva-thakre

PR #75, opened 2021-07-11
Update license information

BoxiLi:
Same as https://github.com/qutip/qutip/pull/1579, remove the license in every single file and update the LICENSE file accordingly. I'll merge this after we come to an agreement with that PR.


jakelishman:
As far as I'm aware, we are in agreement in the main PR, since I got Nathan's tick. I meant to merge it then, but I wanted to wait til I'd finished work for the day, and then I just forgot. I can do it now.

BoxiLi:
Ok, great, thanks! I'll still wait for a few approvals on this PR.

LICENSE
```diff
@@ -1,6 +1,4 @@
-BSD 3-Clause License
-
-Copyright (c) 2020, QuTiP
+Copyright (c) 2011 to 2021 inclusive, QuTiP developers and contributors.
```
jakelishman, line 1: Since qutip-qip is new, I'm not sure you can assert copyright from past years. Also, since a lot of code is migrated from QuTiP main, I think to be in compliance with the QuTiP licence, we may need a statement of attribution about that as well.

jakelishman (commented):
I think there may be some slightly different issues that need addressing for the licensing here  because qutip-qip is built from QuTiP code, and the admin team don't own the copyright entirely, so I think we may actually need to include a copyright notice acknowledging the original contributions of other QuTiP contributors.

jakelishman:
I'm not certain on any of what I said, but that's my rough impression of how licensing works.

LICENSE
BoxiLi: Yeah, maybe this can only be `Copyright (c) 2020 to 2021 inclusive, QuTiP developers and contributors.`

BoxiLi:
> admin team don't own the copyright entirely, so I think we may actually need to include a copyright notice acknowledging the original contributions of other QuTiP contributors.

All the credits were kept during the separation of this repo, so their contributions are still there.

LICENSE
BoxiLi: But indeed I'm not sure if we need to keep a different licence for those files migrated from `qutip` and those created new in `qutip-qip`.

hodgestar: Since this is a QuTiP family package and this is a "fork" of a subset of qutip, I think it can keep the "2011 to 2021 inclusive" and the "QuTiP developers and contributors".

If the contributors were not the same we would write:

```
Copyright (c) 2021, QuTiP QIP developers and contributors.
Copyright (c) 2011 to 2021 inclusive, QuTiP developers and contributors.
```

Because we need to retain the old copyright years and owners and add the new ones. Since they're the same, we can just merge them.

hodgestar (approved):

PR #76, opened 2021-07-11
Rename get_noisy_dynamics to get_noisy_pulses

BoxiLi:
**Description**
In the noise module, users are allowed to define custom noise for their own physical model. The `Noise` class uses a hook method `get_noisy_dynamics` to realize this. This method will be called when including noise in the physical model.

It was pointed out by @quantshah that the name `get_noisy_dynamics` was wrong because the noise object returns a list of `Pulse`, instead of `QobjEvo`. This PR fixed this issue and use the term `get_noisy_pulses`. The old API is still usable but will issue a `PendingDeprecationWarning`.

I also copied the docstring of `get_noisy_pulses` to those in the subclass for easier reference. I originally thought that if the docstrings are missing, they will be inherited automatically, but this is not the case.

I also pinned @hodgestar because I briefly mentioned this and asked for your advice in one of our discussions.

### DIFF
--- a/src/qutip_qip/noise.py
+++ b/src/qutip_qip/noise.py
@@ -223,7 +223,7 @@ class DecoherenceNoise(Noise):
         ----------
         dims: list, optional
             The dimension of the components system, the default value is
-            [2,2...,2] for qubits system.
+            [2, 2, ..., 2] for a system of qubits.
         pulses : list of :class:`.Pulse`
             The input pulses. The noise will be added to pulses in this list.
         systematic_noise : :class:`.Pulse`
--- a/src/qutip_qip/noise.py
+++ b/src/qutip_qip/noise.py
@@ -108,7 +108,7 @@ class Noise(object):
         """
         Return the input pulses list with noise added and
         the pulse independent noise in a dummy :class:`.Pulse` object.
-        This is a template method, a method with the same name and signitures
+        This is a template method, a method with the same name and signatures
         needs to be defined in the subclasses.
 
         Parameters
--- a/src/qutip_qip/noise.py
+++ b/src/qutip_qip/noise.py
@@ -131,19 +131,17 @@ class Noise(object):
         systematic_noise : :class:`.Pulse`
             The dummy pulse representing pulse-independent noise.
         """
-        try:
-            # Try to check if the old API is defined.
-            result = self.get_noisy_dynamics(dims, pulses, systematic_noise)
+        get_noisy_dynamics = getattr(self, "get_noisy_dynamics", None)
+        if get_noisy_dynamics is not None:
             warnings.warn(
                 "Using get_noisy_dynamics as the hook function for custom "
                 "noise will be deprecated, "
                 "please use get_noisy_pulses instead.",
                 PendingDeprecationWarning)
-            return result
-        except AttributeError:
-            raise NotImplementedError(
-                "Subclass error needs a method"
-                "`get_noisy_pulses` to process the noise.") from None
+            return self.get_noisy_dynamics(dims, pulses, systematic_noise)
+        raise NotImplementedError(
+            "Subclass error needs a method"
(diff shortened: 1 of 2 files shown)


src/qutip_qip/noise.py
```diff
@@ -129,18 +129,28 @@ def get_noisy_dynamics(self, dims, pulses, systematic_noise):
...
+            warnings.warn(
+                "Using get_noisy_dynamics as the hook function for custom "
+                "noise will be deprecated, "
+                "please use get_noisy_pulses instead.",
+                PendingDeprecationWarning)
+            return result
+        except AttributeError:
+            raise NotImplementedError(
+                "Subclass error needs a method"
+                "`get_noisy_pulses` to process the noise.") from None
```
hodgestar, line 146: I'm not sure how important this is, but if `.get_noisy_dynamics` exists and raises an `AttributeError` then this code will transform that into a `NotImplementedError` which could be a bit confusing. We could instead do: 
```suggestion
        get_noisy_dynamics = getattr(self, "get_noisy_dynamics", None)
        if get_noisy_dynamics is not None:
            warnings.warn(
                "Using get_noisy_dynamics as the hook function for custom "
                "noise will be deprecated, "
                "please use get_noisy_pulses instead.",
                PendingDeprecationWarning)
            return self.get_noisy_dynamics(dims, pulses, systematic_noise)
        raise NotImplementedError(
            "Subclass error needs a method"
            "`get_noisy_pulses` to process the noise.")
```

tests/test_noise.py
```diff
@@ -187,19 +178,6 @@ def test_random_noise(self):
             noisy_pulses[2].coherent_noise[0].tlist,
             np.linspace(1, 6, int(5/0.2) + 1))
 
-    def test_user_defined_noise(self):
```
hodgestar, line 190: We should add a test that falling back to `.get_noisy_dynamics()` works as expected.

hodgestar: Ah, we do already! Woot.

hodgestar (approved):

hodgestar (commented):
Woot. That should have been a comment / request changes.

hodgestar:
The duplication of the docstrings is not great, but definitely better to have the documentation appear in the apidocs than not.

src/qutip_qip/noise.py
```diff
@@ -103,13 +104,12 @@ class Noise(object):
...
-    def __init__(self):
-        pass
-
-    def get_noisy_dynamics(self, dims, pulses, systematic_noise):
+    def get_noisy_pulses(self, dims=None, pulses=None, systematic_noise=None):
         """
         Return the input pulses list with noise added and
-        the pulse independent noise in a dummy Pulse object.
+        the pulse independent noise in a dummy :class:`.Pulse` object.
+        This is a template method, a method with the same name and signitures
```
hodgestar, line 111:
```suggestion
        This is a template method, a method with the same name and signatures
```

src/qutip_qip/noise.py
```diff
@@ -205,8 +213,29 @@ def __init__(self, c_ops, targets=None, coeff=None, tlist=None,
...
             self, dims=None, pulses=None, systematic_noise=None):
+        """
+        Return the input pulses list with noise added and
+        the pulse independent noise in a dummy :class:`.Pulse` object.
+
+        Parameters
+        ----------
+        dims: list, optional
+            The dimension of the components system, the default value is
+            [2,2...,2] for qubits system.
```
hodgestar, line 226:
```suggestion
            [2, 2, ..., 2] for a system of qubits.
```

hodgestar (approved):
Left a few small docstring suggestions, but looks good to me. Happy for this to be merged whenever you think it is ready @BoxiLi.

BoxiLi:
Thanks for correcting the typos!

PR #70, opened 2021-07-06
Adds two qubit functions

purva-thakre:
**Circuit of Two-Level Unitaries Example**

The output of the function is a tuple of the gate dictionary and two-level gates in order. The gate labels are arbitrary numbered U's which will hopefully change when a two-level unitary gate is added. 

```python
np.set_printoptions(precision=3)
two_qubit = rand_unitary(4,dims=[[2,2],[2,2]])

# Output from the function
dict_and_gate_list = decompose_two_qubit_to_two_level_unitary(two_qubit)

# appending the user_gates attribute to the quantum circuit of interest
quantum_circuit = QubitCircuit(2,reverse_states=False )
user_gates_from_output = dict_and_gate_list[0]
quantum_circuit.user_gates = user_gates_from_output

# add gates to the circuit and get circuit diagram
quantum_circuit.add_gates(dict_and_gate_list[-1])
quantum_circuit.png
```
download

```python
# Comparison to the input
calculatedu = quantum_circuit.compute_unitary()
print(average_gate_fidelity(calculatedu,two_qubit))
```

PR #82, opened 2021-07-22
Adds two qubit decomposition functions

purva-thakre:
**Circuit of Two-Level Unitaries Example**

The output of the function is a tuple of the gate dictionary and two-level gates in order. The gate labels are arbitrary numbered U's which will hopefully change when a two-level unitary gate is added. 

```python
np.set_printoptions(precision=3)
two_qubit = rand_unitary(4,dims=[[2,2],[2,2]])

# Output from the function
dict_and_gate_list = decompose_two_qubit_to_two_level_unitary(two_qubit)

# appending the user_gates attribute to the quantum circuit of interest
quantum_circuit = QubitCircuit(2,reverse_states=False )
user_gates_from_output = dict_and_gate_list[0]
quantum_circuit.user_gates = user_gates_from_output

# add gates to the circuit and get circuit diagram
quantum_circuit.add_gates(dict_and_gate_list[-1])
quantum_circuit.png
```
download

```python
# Comparison to the input
calculatedu = quantum_circuit.compute_unitary()
print(average_gate_fidelity(calculatedu,two_qubit))
```

### DIFF
--- a/src/qutip_qip/decompose/decompose_two_qubit_gate.py
+++ b/src/qutip_qip/decompose/decompose_two_qubit_gate.py
@@ -1,0 +1,293 @@
+import numpy as np
+import cmath
+
+from qutip import Qobj
+from qutip_qip.decompose._utility import (
+    check_gate,
+    MethodError,
+)
+
+import warnings
+from qutip_qip.circuit import Gate
+from qutip_qip.operations import controlled_gate
+
+from .decompose_single_qubit_gate import decompose_one_qubit_gate
+
+# for unknown labels in two level gates
+warnings.filterwarnings("ignore", category=UserWarning)
+
+
+def _U6(single_qubit_U, control=0, target=1, control_value=1):
+    """Defines a two-level unitary whose decomposition is known.
+    Lemma 5.1 of https://arxiv.org/abs/quant-ph/9503016
+    """
+    check_gate(single_qubit_U, num_qubits=1)
+    two_level_U6 = controlled_gate(
+        single_qubit_U, N=2, control=0, target=1, control_value=1
+    )
+    return two_level_U6
+
+
+def _U5(single_qubit_U, control=1, target=0, control_value=1):
+    """Defines a two-level unitary whose decomposition is known.
+    U6 but with the control and target qubits switched.
+    """
+    check_gate(single_qubit_U, num_qubits=1)
+    two_level_U5 = controlled_gate(
+        single_qubit_U, N=2, control=1, target=0, control_value=1
+    )
+    return two_level_U5
+
+
+def _decompose_U6(input_array):
+    """Decomposes a two-level unitary into ABC and CNOT when target =1 and
+    the control value is 1.
+    Based on Lemma 5.1 of arXiv:quant-ph/9503016v1
+    """
+    U = np.array(
+        [[input_array[2][2], input_array[2][3]],
+            [input_array[3][2], input_array[3][3]]]
+    )
+    check_gate(Qobj(U), num_qubits=1)
+    gate_list = decompose_one_qubit_gate(Qobj(U), "ZYZ_PauliX")
+
+    # Change the gate targets for ABC and X gates from default = 0
+    # First 2 indices are for A, next 2 are for B and last is for C
+    # Note that we have avoided changing the Pauli X indices because it is
+    # easier to define a gate with the target qubit.
+    target_indices_to_change = [0, 1, 3, 4, 6]
+    for i in target_indices_to_change:
+        gate_list[i].targets = [1]
+
+    # last is phase whose target is not changed based on Lemma 5.2 of
+    # arXiv:quant-ph/9503016v1 but the gate type is changed
+    phase_angle = gate_list[-1].arg_value
+    phase_gate = Gate(
+        "PHASEGATE",
+        targets=[0],
+        arg_value=phase_angle,
+        arg_label=r"{:0.2f} \times \pi".format(phase_angle / np.pi),
+    )
+    Pauli_X = Gate("X", targets=[1], classical_controls=[0])
+
+    CNOT_ctrl_0 = Gate("CNOT", controls=0, targets=1)
+    return (
+        gate_list[0],
+        gate_list[1],
+        Pauli_X,
+        CNOT_ctrl_0,
+        Pauli_X,
+        gate_list[3],
+        gate_list[4],
+        Pauli_X,
+        CNOT_ctrl_0,
+        Pauli_X,
+        gate_list[6],
+        phase_gate,
+    )
+
+
+def _decompose_U5(input_array):
+    """Decomposes a two-level unitary into ABC and CNOT when target = 0 and
+    control value is 1.
+    Based on Lemma 5.1 of arXiv:quant-ph/9503016v1 with flipped targets.
+    """
+    U = np.array(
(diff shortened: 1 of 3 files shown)


src/qutip_qip/decompose/decompose_two_qubit_gate.py
```diff
@@ -0,0 +1,293 @@
...
+    U4_gate = Gate("U4", targets=[0, 1])
+    U5_gate = Gate("U5", targets=[0, 1])
+    U6_gate = Gate("U6", targets=[0, 1])
+
+    gate_list = (U6_gate, U5_gate, U4_gate, U3_gate, U2_gate, U1_gate)
+
+    return(user_gates, gate_list)
+
+
+def decompose_two_qubit_to_CNOT_and_single_qubit_gates(input_gate):
```
purva-thakre, line 206: Output of this function is not what it is expected to be. For this reason, all of the smaller gates for U6,...,U1 have been defined separately. A for loop will be created for this once the output matches exactly to the input. In addition, all the two qubit decomposition methods will be made private after a general two qubit decomposition function isalso be added - provides choice for circuit output. 

Need to recalculate the target and control qubits because they were different from what was expected for the [two level circuit output function](https://github.com/qutip/qutip-qip/pull/70/files#diff-4fd62396e554b500c5f57c03c2e6cf0123b85fb4ecb549a57b97f8d120a25f62R174). The output of the two level output matches [exactly to the input](https://github.com/qutip/qutip-qip/pull/70/files#diff-5b0436268788ff97a67b84a89049cd3120d5e7f7013de9df7073097ed2046adfR13) with new values for controls and targets. 

Just because I know the correct control and targets does not imply how the controls and targets for CNOT and Pauli X of each gate should be corrected. 

The hope is to check and compare my calculations with the targets and controls from gray code function's output. 

Tentative new values for the target/controls - [cnot_gates_2_qubits.pdf](https://github.com/qutip/qutip-qip/files/6859753/cnot_gates_2_qubits.pdf)

src/qutip_qip/decompose/decompose_two_qubit_gate.py
```diff
@@ -0,0 +1,293 @@
...
+        gate_list[4],
+        Pauli_X,
+        CNOT_ctrl_1,
+        Pauli_X,
+        gate_list[6],
+        phase_gate,
+    )
+
+
+def _decompose_to_two_level_arrays(input_gate):
```
purva-thakre, line 131: This function will be discarded after a general function decomposing a n qubit gate to two level arrays is defined. 

The plan is to iterate over the row and column indices in a for loop to avoid manually [defining indices for the decomposition. ](https://github.com/qutip/qutip-qip/pull/70/files#diff-4fd62396e554b500c5f57c03c2e6cf0123b85fb4ecb549a57b97f8d120a25f62R141)

src/qutip_qip/decompose/decompose_two_qubit_gate.py
```diff
@@ -0,0 +1,293 @@
...
+        U__two_level_dagger = Qobj(U__two_level_dagger, dims=[[2, 2], [2, 2]])
+        array_list.append(U__two_level_dagger)
+
+    # for U6 - multiply input array by U5 and take dagger
+    U6_dagger = input_array
+    array_list.append(Qobj(U6_dagger, dims=[[2, 2], [2, 2]]))
+    return(array_list)
+
+
+def decompose_two_qubit_to_two_level_unitary(input_gate):
```
purva-thakre, line 172: Same as above. Once a general method for decomposition to two level unitaries exists, this function will be discarded. 

Defined here for convenience of checking the output of a two qubit gate.

src/qutip_qip/decompose/decompose_two_qubit_gate.py
```diff
@@ -0,0 +1,293 @@
...
+    user_gates = {
+        "U1": array_list[0],
+        "U2": array_list[1],
+        "U3": array_list[2],
+        "U4": array_list[3],
+        "U5": array_list[4],
+        "U6": array_list[5],
+        }
+
+    U1_gate = Gate("U1", targets=[0, 1])
```
purva-thakre, line 194: Think of better names for two-level unitaries.

src/qutip_qip/decompose/decompose_two_qubit_gate.py
BoxiLi: > Issue with gray code - There's ambiguity as to how you map from 1 state to another as long as only 1 bit is changed at each step.

I think if all qubits are the same, the physical resources shouldn't make a difference. Of course in real hardware qubits are not the same and gate implementation is not always symmetric so one way might be better than the other. But I think that is a very concrete scenario and we don't need to care about it now. It would be good enough to just define a convention like *always go with the route for smaller binary number (01 < 10)*.

>The binary sequence for 2 qubits is binary_sequence= [00, 01, 10, 11] which can be re-written in terms of array indices as [0, 1, 2, 3].

Not sure if this is helpful here, `numpy` has `ravel_multi_index` and `unravel_index` to compute between binary representation and matrix row/column index.

BoxiLi: E.g. 
```python
>>> np.ravel_multi_index([0,1,0],[2,2,2])
2
```

purva-thakre: > the physical resources shouldn't make a difference

But the order that you map in does matter because the sequence has to be cyclic. When I did my calculations, I had tried to use the shortest method to go from one state another in the hope that the total gate number would decrease. As long as you choose a consistent method ( for this project, I have chosen [reflect and prefix method](http://www.cs.williams.edu/~jannen/teaching/f16/cs135/lectures/lecture.11.pdf)). 

I now think my error is not related to the gray code sequence but instead about the order of gates and/or controls, targets (see example below). I might have missed some information from the reference papers. Now, my plan is to at least have a working code for what was planned this week and then come back to correcting a few lines after I have figured out my error. This should also give you a chance to review the code logic etc.  

Suppose I want to decompose the following matrix :

image

The non-trivial basis vectors are 01 and 10. Using the gray code generator, 01 -> 11 -> 10. 

- For 01 -> 11 : When qubit 2 is 1, qubit 1 is flipped. Thus, there's a CNOT gate added to the circuit with `controls=[1], targets=[0], control_value=1`
- For 11 -> 10 : When qubit 1 is 1, qubit 2 is flipped. This is a controlled gate on target 0 with control value 1. But because this is the last steps in the sequence, the non-trivial sub-matrix has to be added to the circuit. The form of this is U6 whose decomposition is known in Lemma 5.1.  
- The last gate is to go back to state 01. 

image

image


The circuit's unitary is calculated as :

image

purva-thakre: > E.g.
> 
> ```python
> >>> np.ravel_multi_index([0,1,0],[2,2,2])
> 2
> ```

Thanks ! I defined those functions yesterday. I'll keep them for now. If it's preferred to use numpy, I'll make the changes then.

BoxiLi: Looks good.

image

This circuit makes sense. This is the same way how the SWAP gate was decomposed. Only with a unitary in the middle.

purva-thakre:
Closing this because the code has been added locally in the gray code ordering scheme.

PR #83, opened 2021-07-24
Separate the gate definition from the QubitCircuit.propagators method

BoxiLi:
**Changes**
- Merge the duplicated `QubitCircuit.propagators` methods.
- Move the definition for gate operators to the `Gate` class: one method defining the compact gate operator as Qobj, while the other one expands it to the full Hilbert space. This makes it easier to define subclasses for each gate later.
- Rename a previously undocumented `QubitCircuit.get_inds` method to `get_all_qubits` and add documentation. It was added recently but never documented.

There are two new method for `Gate`: `get_qobj` and `get_compact_qobj`. Although they could be merged to one, I made them two different methods because if for custom gate, I wish that the user only need to define `get_compact_qobj`, without having to consider permutation and expanding etc. And in `get_qobj`, we take care of that.

The special case is the global phase gate. Currently, in `propagators` it always gives a matrix with the full dimension. However, it should be treated as just a complex number, which is more efficient. I suspect that in the circuit simulation with precomputed unitaries the global phase gate is not properly handled. But I will leave this for another PR.

### DIFF
--- a/src/qutip_qip/circuit.py
+++ b/src/qutip_qip/circuit.py
@@ -1294,7 +1294,7 @@ class QubitCircuit:
             if gate.name in self.user_gates:
                 if gate.controls is not None:
                     raise ValueError("A user defined gate {} takes only  "
-                                    "`targets` variable.".format(gate.name))
+                                     "`targets` variable.".format(gate.name))
                 func_or_oper = self.user_gates[gate.name]
                 if inspect.isfunction(func_or_oper):
                     func = func_or_oper
--- a/src/qutip_qip/circuit.py
+++ b/src/qutip_qip/circuit.py
@@ -33,7 +33,7 @@
 from collections.abc import Iterable
 from itertools import product
 import numbers
-
+import warnings
 import inspect
 
 import numpy as np
@@ -202,7 +202,7 @@ class QubitCircuit:
         self.N = N
         self.reverse_states = reverse_states
         self.gates = []
-        self.dims = dims
+        self.dims = dims if dims is not None else [2] * N
         self.num_cbits = num_cbits
 
         if input_states:
@@ -1247,7 +1247,7 @@ class QubitCircuit:
 
         return temp
 
-    def propagators(self, expand=True):
+    def propagators(self, expand=True, ignore_measurement=False):
         """
         Propagator matrix calculator returning the individual
         steps as unitary matrices operating from left to right.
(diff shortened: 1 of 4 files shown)


src/qutip_qip/circuit.py
```diff
@@ -202,7 +202,7 @@ def __init__(self, N, input_states=None, output_states=None,
         self.N = N
         self.reverse_states = reverse_states
         self.gates = []
-        self.dims = dims
+        self.dims = dims if dims is not None else [2] * N
```
hodgestar, line 205: Question for my own understanding: What are the cases where it would be useful to have dims not be `[2] * N`?

src/qutip_qip/circuit.py
```diff
@@ -1262,240 +1262,64 @@ def propagators(self, expand=True):
...
+            if gate.name == "GLOBALPHASE":
+                qobj = gate.get_qobj(self.N)
+                U_list.append(qobj)
+                continue
+
+            if gate.name in self.user_gates:
                 if gate.controls is not None:
                     raise ValueError("A user defined gate {} takes only  "
-                                     "`targets` variable.".format(gate.name))
+                                    "`targets` variable.".format(gate.name))
```
hodgestar, line 1297: Not sure why the indentation was changed here.

src/qutip_qip/circuit.py
```diff
@@ -1262,240 +1262,64 @@ def propagators(self, expand=True):
...
-                U_list.append(globalphase(gate.arg_value, self.N))
-            elif gate.name == "IDLE":
-                U_list.append(qeye(2))
-            elif gate.name in self.user_gates:
+            if gate.name == "GLOBALPHASE":
+                qobj = gate.get_qobj(self.N)
+                U_list.append(qobj)
+                continue
+
+            if gate.name in self.user_gates:
```
hodgestar, line 1294: Perhaps now that we have nicer gate classes we could get rid of `user_gates` in future and just add some helpers for turning unitaries or functions into proper gates? Thought for the future though, not for now.

hodgestar (approved):
I left some questions, but this looks great as a first step.

src/qutip_qip/circuit.py
BoxiLi: In principle, there is nothing preventing one from defining a qutrit/qudit/multi_level circuit. Although there is no predefined gate for that. It is not very popular at the moment but it is also not very hard to support it.

src/qutip_qip/circuit.py
BoxiLi: Hmm, don't know. Removed.

src/qutip_qip/circuit.py
BoxiLi: Indeed, this will make that also much easier.

BoxiLi:
Hmmm, I understand that code climate does not like 20 if-else, neither do I. But let's leave it to another PR.

src/qutip_qip/circuit.py
hodgestar: Woot. Maybe we should write some tests or examples with qutrit+ gates at some point so that we can be sure it all works and so that maybe users can find it and try it out.

hodgestar:
Still approved!

src/qutip_qip/circuit.py
BoxiLi: There is one :)

https://github.com/qutip/qutip-qip/blob/8b49a89405183eb3e264abfc50aab3bf47fe2ef8/tests/test_circuit.py#L513-L532

BoxiLi: It is just that the *default* `dims` was not used at all before because we don't have predefined qutrits gate anyway. But now I'm using it in expanding the circuit.

But you are right that it is not tested if part of the circuit is qubits using the predefined gate, while the rest is using custom qutrits gate.

PR #84, opened 2021-07-25
Create .codeclimate.yml

BoxiLi:
Copy the code climate settings from qutip

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,46 @@
+version: "2"
+exclude_patterns:
+- "doc/"
+- "dist/"
+- "**/tests/"
+checks:
+  argument-count:
+    config:
+      threshold: 10
+  complex-logic:
+    config:
+      threshold: 20
+  file-lines:
+    enabled: false
+  method-complexity:
+    config:
+      threshold: 20
+  method-count:
+    config:
+      threshold: 75
+  method-lines:
+    config:
+      threshold: 100
+  nested-control-flow:
+    config:
+      threshold: 4
+  return-statements:
+    config:
+      threshold: 4
+  similar-code:
+    config:
+      threshold: 64
+  identical-code:
+    config:
+      threshold: 64
+plugins:
+  fixme:
+    enabled: true
+  pep8:
+    enabled: true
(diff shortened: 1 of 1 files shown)


BoxiLi:
As this is the initial commit of codeclimate, I think we can merge this as-is and solve those climate issues later step by step?

.codeclimate.yml
```diff
@@ -0,0 +1,46 @@
+version: "2"
+exclude_patterns:
+- "doc/"
+- "dist/"
+- "**/tests/"
```
hodgestar, line 5: Should we include tests in CodeClimate? Tests are code too.

hodgestar:
I would personally prefer not to use a tool run by a third party like CodeClimate and rather to just run these tools ourselves in CI (e.g. flake8), but qutip itself already uses CodeClimate and it's better to have something than nothing, so let's go for it.

Aside: Reasons for not liking a third party tool here -- it's a real pain to run CodeClimate checks locally and CodeClimate maintains state about which "bad" bits of code should be ignored that is completely outside of our control and invisible in the repository.

hodgestar (approved):
I left a question about using CodeClimate on tests and some thoughts about CodeClimate itself, but I am also +1 on merging this as is.

BoxiLi:
>  it's a real pain to run CodeClimate checks locally

That is true. I don't know a way to run it locally either. Codeclimate checks a few more things as flake8, like similar code block and too complicated logic etc. We could find something that works both for qutip and qutip-qip.

BoxiLi:
Interestingly, now you can choose in codeclimate to *approve* the issues for this PR so that the tests will all pass.

.codeclimate.yml
BoxiLi: Yeah I agree.

hodgestar:
> That is true. I don't know a way to run it locally either. Codeclimate checks a few more things as flake8, like similar code block and too complicated logic etc. We could find something that works both for qutip and qutip-qip.

There is a radon plugin for flake8 that will check code complexity -- https://github.com/rubik/radon -- and which I usually use with flake8.

PR #85, opened 2021-07-25
Support different pulse shapes

BoxiLi:
- Different pulse shapes are generated using window function from scipy.signal.windows.
- A generate_pulse_shape method is defined and can be used to generate pulse shape with the arbitrary area and required maximum.
- A default single qubit compiler is defined and simplifies the various compiler subclasses.

### DIFF
--- a/src/qutip_qip/compiler/cavityqedcompiler.py
+++ b/src/qutip_qip/compiler/cavityqedcompiler.py
@@ -83,7 +83,7 @@ class CavityQEDCompiler(GateCompiler):
         coeff, tlist = self.generate_pulse_shape(
             args["shape"], args["num_samples"],
             maximum=self.params["sx"][targets[0]],
-            area=gate.arg_value / 2. / np.pi * 0.5)  # operator is= X, not X/2
+            area=gate.arg_value / 2. / np.pi * 0.5)  # operator is X, not X/2
         pulse_info = [("sx" + str(targets[0]), coeff)]
         return [Instruction(gate, tlist, pulse_info)]
 
--- a/src/qutip_qip/compiler/cavityqedcompiler.py
+++ b/src/qutip_qip/compiler/cavityqedcompiler.py
@@ -1,4 +1,6 @@
+from functools import partial
 import numpy as np
+import scipy
 
 from ..circuit import QubitCircuit
 from ..operations import Gate
@@ -10,8 +12,22 @@ __all__ = ['CavityQEDCompiler']
 
 class CavityQEDCompiler(GateCompiler):
     """
-    Decompose a :class:`.QubitCircuit` into
-    the pulse sequence for the processor.
+    Compiler for :obj:`.DispersiveCavityQED`.
+    Compiled pulse strength is in the unit of GHz.
+
+    Supported native gates: "RX", "RY", "RZ", "ISWAP", "SQRTISWAP",
+    "GLOBALPHASE".
+
+    Default configuration (see :obj:`.GateCompiler.args` and
+    :obj:`.GateCompiler.compile`):
+
+        +-----------------+--------------------+
+        | key             | value              |
+        +=================+====================+
+        | ``shape``       | ``rectangular``    |
+        +-----------------+--------------------+
+        |``params``       | Hardware Parameters|
+        +-----------------+--------------------+
 
     Parameters
     ----------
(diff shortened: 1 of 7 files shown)


BoxiLi:
@hodgestar Would you like to have a look when you get time, thanks :)

hodgestar (commented):
@BoxiLi Hello! Apologies for the delay. Woot on adding the different pulse shapes.

This is not a complete review, but I have a strong preference for the simpler one-compiler-method-or-function-per-gate model that we had before. The new generic-one-qubit and generic-two-qubit methods make this less flexible and harder to read and understand.

If we feel a that there is a lot of commonality in such compiler functions, we should instead write smaller helper functions that they can use -- i.e. a sort of "mini-language" for writing the compiler functions.

E.g. For rz_compiler one might write:

```python
def rz_compiler(self, gate, args):
    tlist, pulse_info = self.pauli_pulse("sz", gate.arg_value, gate.targets[0])
    return [Instruction(gate, list, pulse_info)]

def pauli_pulse(self, op_name, phi, target):
    g = self.params[op_name][target]
    coeff = np.sign(phi) * g
    tlist = abs(phi) * (2 * g) / np.pi
    pulse_info = [(op_name + str(target), coeff)]
    return tlist, pulse_info
```

And then `rx_compile` is also quite short and clear. I'm not sure this is exactly the write way to break it down, but hopefully it illustrates the idea -- break functions down into meaningful steps that correspond to known self-contained concepts, don't generalize them and add lots of `if ...:` statements.

src/qutip_qip/compiler/cavityqedcompiler.py
```diff
@@ -95,28 +111,20 @@ def rx_compiler(self, gate, args):
         Compiler for the RX gate
         """
         targets = gate.targets
-        g = self.params["sx"][targets[0]]
-        coeff = np.sign(gate.arg_value) * g
-        tlist = abs(gate.arg_value) / (2*g) / np.pi / 2
+        coeff, tlist = self.generate_pulse_shape(
+            args["shape"], args["num_samples"],
+            maximum=self.params["sx"][targets[0]],
+            area=gate.arg_value / 2. / np.pi * 0.5)  # operator is= X, not X/2
```
hodgestar, line 117:
```suggestion
            area=gate.arg_value / 2. / np.pi * 0.5)  # operator is X, not X/2
```

src/qutip_qip/compiler/cavityqedcompiler.py
```diff
@@ -127,55 +135,75 @@ def sqrtiswap_compiler(self, gate, args):
...
-        tlist = (4 * np.pi / abs(J)) / 8 / np.pi / 2
+            1. / self.Delta[q1] + 1. / self.Delta[q2]) / 2.
+        if gate_name == "ISWAP":
+            area = 1. / 2.
+            correction_angle = -np.pi / 2.
+        elif gate_name == "SQRTISWAP":
+            area = 1. / 4.
+            correction_angle = -np.pi / 4.
+        else:
+            raise ValueError(f"Gate {gate.name} cannot not be compiled.")
```
hodgestar, line 146: If we add area and correction_angle as parameters, we can remove the if statement and try-except.

src/qutip_qip/compiler/cavityqedcompiler.py
```diff
@@ -95,28 +111,20 @@ def rx_compiler(self, gate, args):
...
-        """
-        Compiler for the SQRTISWAP gate
-
-        Notes
-        -----
-        This version of sqrtiswap_compiler has very low fidelity, please use
-        iswap
-        """
-        # FIXME This decomposition has poor behaviour
+    def _two_qubit_compiler(self, gate_name, gate, args):
```
hodgestar, line 121:
```suggestion
    def _swap_compiler(self, gate_name, gate, args):
```

src/qutip_qip/compiler/spinchaincompiler.py
```diff
@@ -118,43 +133,72 @@ def rx_compiler(self, gate, args):
...
         g = self.params["sxsy"][q1]
-        coeff = -g
-        tlist = np.pi / (4 * g) / np.pi / 2
+        maximum = g
+        if gate_name == "ISWAP":
+            area = -1 / 8
+        elif gate_name == "SQRTISWAP":
+            area = -1 / 16
+        else:
+            raise ValueError(f"Gate {gate.name} cannot not be compiled.")
```
hodgestar, line 153: Same comment about removing the `if` blocks here by adding area as a parameter.

tests/test_compiler.py
```diff
@@ -183,3 +184,21 @@ def test_compiler_result_format():
     processor.set_all_tlist(tlist)
     assert_array_equal(processor.pulses[0].coeff, coeffs[0])
     assert_array_equal(processor.pulses[0].tlist, tlist[0])
+
+
+@pytest.mark.parametrize("shape", _default_window_t_max.keys())
```
hodgestar, line 189: Let's add a test for the rectangular shape too somewhere.

src/qutip_qip/compiler/gatecompiler.py
```diff
@@ -68,8 +70,18 @@ class GateCompiler(object):
...
+        The compilation configurations.
+        It will be passed to each compiling functions.
+        Available arguments:
+
+        * ``shape``: The compiled pulse shape. ``rectangular`` or
+          one of the `SciPy window functions
+          <https://docs.scipy.org/doc/scipy/reference/signal.windows.html>`_.
+        * ``num_samples``:
+          Number of samples in continuous pulse.
+          It has not effect in rectangular pulse.
```
hodgestar, line 82:
```suggestion
          Number of samples for continuous pulses.
          It has no effect for rectangular pulses.
```

src/qutip_qip/compiler/gatecompiler.py
```diff
@@ -80,8 +92,11 @@ def __init__(self, num_qubits=None, params=None, pulse_dict=None, N=None):
             "GLOBALPHASE": self.globalphase_compiler,
             "IDLE": self.idle_compiler
         }
-        self.args = {}
-        self.args.update({"params": self.params})
+        self.args = {  # Default configuration
+            "shape": "rectangular",
+            "num_samples": None,
+            "params": self.params,
```
hodgestar, line 98: I think we should think about whether `shape`, `num_samples` and `params` belong in `self.args` or not. Quite possibly how it is is currently fine, but we should have a clear idea of why some attributes are in `args` and others not.

src/qutip_qip/compiler/circuitqedcompiler.py
```diff
@@ -78,6 +119,9 @@ def cnot_compiler(self, gate, args):
         sign = np.sign(amplitude) * np.sign(area)
         tlist = tlist / amplitude * area * sign
         coeff = coeff * amplitude * sign
+        from scipy.integrate import simps
```
hodgestar, line 122:
```suggestion
```
Removing debug import?

src/qutip_qip/compiler/circuitqedcompiler.py
```diff
@@ -32,37 +52,58 @@ def _normalized_gauss_pulse(self):
...
-        coeff = (np.exp(-(tlist-td/2)**2/2/sigma**2)
-            - np.exp(-(0-td/2)**2/2/sigma**2)) / max_pulse
+        coeff = (
+            np.exp(-(tlist-td/2)**2/2/sigma**2)
+            - np.exp(-(0-td/2)**2/2/sigma**2)
+            ) / max_pulse
         return tlist, coeff
 
-    def single_qubit_compiler(self, gate, args):
+    def ry_compiler(self, gate, args):
```
hodgestar, line 63: Maybe we can write an `_rotation_compiler` the same way we have `_swap_compiler` elsewhere now.

hodgestar (changes requested):
Reviewed with Boxi on a call.

BoxiLi:
Some of the changes are not tracked correctly on GitHub because I rebased against master, but you should be able to see the new changes following the list of commits. Thanks!

src/qutip_qip/compiler/cavityqedcompiler.py
```diff
@@ -47,44 +63,85 @@ def __init__(
         self.Delta = self.wq - self.params["w0"]
         self.global_phase = global_phase
 
-    def rz_compiler(self, gate, args):
+    def _rotation_compiler(self, gate, op_label, param_label, args):
         """
-        Compiler for the RZ gate
+        Single qubit gate compiler.
```
hodgestar, line 68:
```suggestion
        Single qubit rotation compiler.
```

src/qutip_qip/compiler/cavityqedcompiler.py
```diff
@@ -95,55 +152,69 @@ def sqrtiswap_compiler(self, gate, args):
...
+        compiled_gate1 = self.gate_compiler["RZ"](gate1, args)
         instruction_list += compiled_gate1
-        gate2 = Gate("RZ", [q2], None, arg_value=-np.pi/4)
-        compiled_gate2 = self.rz_compiler(gate2, args)
+        gate2 = Gate("RZ", [q2], None, arg_value=correction_angle)
+        compiled_gate2 = self.gate_compiler["RZ"](gate2, args)
         instruction_list += compiled_gate2
-        gate3 = Gate("GLOBALPHASE", None, None, arg_value=-np.pi/4)
+        gate3 = Gate("GLOBALPHASE", None, None, arg_value=correction_angle)
         self.globalphase_compiler(gate3, args)
```
hodgestar, line 168: Should this globalphase gate be added to the instruction list somewhere?

src/qutip_qip/compiler/gatecompiler.py
hodgestar: If `args` is "what is passed to the compiler functions", then I think these are fine here.

src/qutip_qip/compiler/gatecompiler.py
```diff
@@ -287,3 +302,87 @@ def _process_idling_tlist(
...
+        coeff *= np.abs(maximum) * sign
+        tlist *= abs(area) / np.abs(maximum)
+        return coeff, tlist
+
+
+"""
+Normalized scipy window functions.
+The scipy implementation only makes sure that it is maximum is 1.
+Here, we save a default t_max so that the integral is always 1.
+"""
```
hodgestar, line 357: I forgot to mention that this should be in a comment, not a string. A string is only a docstring if it occurs as the first statement in a module, class or function body. Otherwise it is just a string. So this is a bit like write `"a"` on a line by itself in the middle of a program (except more informative, but still a bit strange).

hodgestar (changes requested):
@BoxiLi Largely looks good to me now.

I think we might also need some more tests of the compilers with a full set of gates, because, e.g., we seemed to not add the globalphase gate to the end of the `_swap_compiler` but no tests failed. Doesn't necessarily have to happen in this PR, but maybe would be good to have later (unless I missed something silly).

src/qutip_qip/compiler/cavityqedcompiler.py
BoxiLi: The global phase is treated differently than the others, it is directly added to an attribute.
```python
    def globalphase_compiler(self, gate, args):
        """
        Compiler for the GLOBALPHASE gate
        """
        self.global_phase += gate.arg_value
```

This is because it is a virtual gate that has no corresponding physical instruction. In principle, we can add an empty `Instruction` here but we still need to filter it out later because the scheduler will be confused by something with 0 time duration.

I do agree that the treatment of the global phase gate is not really consistent with other gates. The same problem happens in `QubitCircuit` (because it does not have a compact Qobj representation).

hodgestar (approved):

hodgestar:
@BoxiLi I approved.

The "test-qutip-support" test failed but the logs of the test run were missing for some unknown reason, so I clicked the re-run button.

There are two codeclimate issues about the duplicated _rotation_compiler method that we should ignore (but which I don't think I have permission to sort out).

BoxiLi:
This is because Coveralls is still under maintenance. https://status.coveralls.io/?utm_source=embed

hodgestar:
> This is because Coveralls is still under maintenance. https://status.coveralls.io/?utm_source=embed

Eek. Let's merge this then. It won't hurt to be missing coverage from one PR.

BoxiLi:
> but which I don't think I have permission to sort out

Oh you don't? I thought if you are an org admin you can do this automatically. I approved them.

PR #86, opened 2021-07-26
Adds gray code ordering method

purva-thakre:
**Examples**

1. **Plotting the gray code grid**
Suppose the non-trivial state indices of a two-level unitary are 10 and 31. The gray code sequence of this mapping can be obtained via

```python
plot_gray_code_grid(10,31,5)
plt.show()
``` 
image

2.

### DIFF
--- a/src/qutip_qip/decompose/_utility.py
+++ b/src/qutip_qip/decompose/_utility.py
@@ -136,26 +136,20 @@ def _gray_code_steps(index_of_state_1, index_of_state_2, num_qubits):
     gray_code_sequence = _gray_code_sequence(num_qubits)
     state_1_gray_code_index = gray_code_sequence.index(state_1_binary)
     state_2_gray_code_index = gray_code_sequence.index(state_2_binary)
-    num_steps_gray_code = state_2_gray_code_index - state_1_gray_code_index
+    num_steps_gray_code = np.abs(
+        state_2_gray_code_index - state_1_gray_code_index)
 
     # create a smaller gray code sequence between the two states of interest
-    gray_code = _gray_code_sequence(num_qubits)[
-        state_1_gray_code_index:state_2_gray_code_index+1]
+    if state_1_gray_code_index < state_2_gray_code_index:
+        gray_code = _gray_code_sequence(num_qubits)[
+            state_1_gray_code_index:state_2_gray_code_index+1]
+    else:  # check math for reversed order
+        # what is the difference between using gates from left to right vs
+        # right to left - with current code, the order is mixed.
+        gray_code = _gray_code_sequence(num_qubits)[
+            state_2_gray_code_index:state_1_gray_code_index+1]
+
     return(gray_code, num_steps_gray_code)
-
-
-def gray_code_plot(index_of_state_1, index_of_state_2, num_qubits):
-    """ Plots the difference between each step of a gray code sequence.
-
-    Parameters
-    -----------
-    index_of_state_i: int
-        The non-trivial indices in a two-level unitary matrix gate.
-
-    num_qubits:
-        Number of qubits being acted upon by the quantum gate.
-    """
-    # This function is still incomplete.
 
 
 def gray_code_gate_info(index_of_state_1, index_of_state_2, num_qubits):
@@ -181,7 +175,7 @@ def gray_code_gate_info(index_of_state_1, index_of_state_2, num_qubits):
     for i in range(num_steps+1):
         bit_array_at_i = []
         a = gray_code[i]
-        for j in range(3):
+        for j in range(num_qubits):
             bit_array_at_i.append(a[j])
 
         input_binary_values[i] = bit_array_at_i
@@ -195,7 +189,7 @@ def gray_code_gate_info(index_of_state_1, index_of_state_2, num_qubits):
         control_value = []
         target = []
         all_together = {}
-        for j in range(3):
+        for j in range(num_qubits):
             if a[j] == b[j]:
                 controls.append(j)
                 control_value.append(a[j])
--- a/src/qutip_qip/decompose/_utility.py
+++ b/src/qutip_qip/decompose/_utility.py
@@ -139,6 +139,10 @@ def _gray_code_steps(index_of_state_1, index_of_state_2, num_qubits):
     num_steps_gray_code = np.abs(
         state_2_gray_code_index - state_1_gray_code_index)
 
+    if num_steps_gray_code == 1:
+        num_steps_gray_code = num_steps_gray_code
+    else:  # repeat the mapping back to initial index
+        num_steps_gray_code = 2*num_steps_gray_code - 1
     # create a smaller gray code sequence between the two states of interest
     if state_1_gray_code_index < state_2_gray_code_index:
         gray_code = _gray_code_sequence(num_qubits)[
@@ -161,12 +165,13 @@ def gray_code_gate_info(index_of_state_1, index_of_state_2, num_qubits):
     gray_code = gray_code_info[0]
     num_steps = gray_code_info[1]
 
-    # find number of ones in each gray code step, this will be used
-    # to keep track of control values.
-    # one_bit_count = []
-    # for i in range(num_steps+1):
-    #    check_bit = gray_code[i].count("1")
-    #    one_bit_count.append(check_bit)
+    # repeat the mapping from last index to first initial index
+    if num_steps == 1:
+        gray_code = gray_code
+    else:  # skip last index
+        for i in reversed(range(len(gray_code)-1)):
+            gray_code_repeat = gray_code[i]
+            gray_code.append(gray_code_repeat)
 
     # make a dictionary of an array of binary values
     # find position of 1's
--- a/src/qutip_qip/decompose/_utility.py
+++ b/src/qutip_qip/decompose/_utility.py
@@ -1,5 +1,6 @@
 from qutip import Qobj
-
+import matplotlib.pyplot as plot
+import numpy as np
 
 class MethodError(Exception):
     """When invalid method is chosen, this error is raised."""
@@ -29,3 +30,166 @@ def check_gate(gate, num_qubits):
         raise ValueError("Input is not unitary.")
     if gate.dims != [[2] * num_qubits] * 2:
         raise ValueError(f"Input is not a unitary on {num_qubits} qubits.")
+
+
+def _binary_sequence(num_qubits):
+    """ Defines the binary sequence list for basis vectors of a n-qubit gate.
+
+    The string at index `i` is also the row/column index for a basis vector in
+    a numpy array.
+    """
+    old_sequence = ['0', '1']
+    full_binary_sequence = []
+
+    if num_qubits == 1:
+        full_binary_sequence = old_sequence
+    else:
+        for x in range(num_qubits-1):
+            full_binary_sequence = []
+            zero_append_sequence = ['0' + x for x in old_sequence]
+            full_binary_sequence.extend(zero_append_sequence)
+            one_append_sequence = ['1' + x for x in old_sequence]
+            full_binary_sequence.extend(one_append_sequence)
+            old_sequence = full_binary_sequence
+
+    return(full_binary_sequence)
+
+
+def _gray_code_sequence(num_qubits, output_form=None):
+    """ Finds the sequence of gray codes for basis vectors by using logical
+    XOR operator of Python.
+
+    For print( _gray_code_sequence(2)), the output is [0, 1, 3, 2] i.e. in
+    terms of the binary sequence, the output is ['00', '01', '11', '10'].
+
+    https://docs.python.org/3/library/operator.html#operator.xor'
+
+    Parameters
+    ----------
+    num_qubits
+        Number of qubits in the circuit
+
+    output_form : :"index_values" or None
+        The format of output list. If a string "index_values" is provided then
+    the function's output is in terms of array indices of the binary sequence.
+    The default is a list of binary strings.
+
+    Returns
+    --------
+    list
+        List of the gray code sequence in terms of array indices or binary
+        sequence positions.
+    """
+    gray_code_sequence_as_array_indices = []
+
+    for x in range(2**num_qubits):
+        gray_code_at_x = x ^ x // 2  # floor operator to shift bits by 1
+        # when the shift is done, the new spot is filled with a new value.
+        gray_code_sequence_as_array_indices.append(gray_code_at_x)
+        if output_form == "index_values":
+            output = gray_code_sequence_as_array_indices
+        else:
+            gray_code_as_binary = []
+            binary_sequence_list = _binary_sequence(num_qubits)
+            for i in gray_code_sequence_as_array_indices:
+                gray_code_as_binary.append(binary_sequence_list[i])
+            output = gray_code_as_binary
+    return(output)
+
+
+def _gray_code_steps(index_of_state_1, index_of_state_2, num_qubits):
+    """ Finds the sequence mapping from state 1 to state 2.
+
+    State 1 and 2 define the basis vectors of non-trivial values in a two-level
+    unitary matrix. Here, the inputs are their respective indices in a quantum
+    gate's array.
+
+    This function finds the number of steps between both states when only 1 bit
+    can be changed at each step.
+    """
+    # check array values are not the same
+    try:
+        assert (index_of_state_1 != index_of_state_2)
+    except AssertionError:
+        raise IndexError("Both indices need to be different.")
+
+    # array index for original binary sequence
+    indices_of_array = []
+    for i in range(2**num_qubits):
+        indices_of_array.append(i)
+
+    # Check both indices could be indices for num_qubits array
+    try:
+        assert all(x in indices_of_array for x in [
+                index_of_state_1, index_of_state_2])
+    except AssertionError:
+        raise IndexError(
+            "At least one of the input indices is invalid.")
+
+    # get the basis vector strings
+    binary_sequence_list = _binary_sequence(num_qubits)
+    state_1_binary = binary_sequence_list[index_of_state_1]
+    state_2_binary = binary_sequence_list[index_of_state_2]
+
+    # compare binary sequence positions to gray code sequence
+    # finds number of steps neeeded to go from one state to another
+    gray_code_sequence = _gray_code_sequence(num_qubits)
+    state_1_gray_code_index = gray_code_sequence.index(state_1_binary)
+    state_2_gray_code_index = gray_code_sequence.index(state_2_binary)
+    num_steps_gray_code = state_2_gray_code_index - state_1_gray_code_index
+
+    # create a smaller gray code sequence between the two states of interest
+    gray_code = _gray_code_sequence(num_qubits)[
+        state_1_gray_code_index:state_2_gray_code_index+1]
+    return(gray_code, num_steps_gray_code)
+
+
+def gray_code_plot(index_of_state_1, index_of_state_2, num_qubits):
+    """ Plots the difference between each step of a gray code sequence.
+
+    Parameters
+    -----------
+    index_of_state_i: int
+        The non-trivial indices in a two-level unitary matrix gate.
+
+    num_qubits:
+        Number of qubits being acted upon by the quantum gate.
+    """
+    # This function is still incomplete.
+
+
+def gray_code_gate_info(index_of_state_1, index_of_state_2, num_qubits):
+    """ Finds information about control and targets for CNOT gate in a gray
+    code sequence.
+    """
+    gray_code_info = _gray_code_steps(
+        index_of_state_1, index_of_state_2, num_qubits)
+    gray_code = gray_code_info[0]
+    num_steps = gray_code_info[1]
+
+    # find number of ones in each gray code step, this will be used
+    # to keep track of control values.
+    one_bit_count = []
+    for i in range(num_steps+1):
+        check_bit = gray_code[i].count("1")
+        one_bit_count.append(check_bit)
+
+    gate_dictionary = {}
+    for i in range(num_steps):
+        a = gray_code[i]
+        b = gray_code[i+1]
+        for j in range(num_qubits):
+            gate_controls = []
+            gate_target = []
+            gate_control_value = []
+            while a[j] == b[j]:
+                partial_gate_control = j
+                gate_controls.append(partial_gate_control)
+                gate_control_value.append(a[j])
+            else:
+                partial_gate_target = j
+                gate_target.append(partial_gate_target)
+
+        gate_dictionary[i] = [gate_controls, gate_control_value, gate_target]
+
+    return(gate_dictionary)
--- a/src/qutip_qip/decompose/_utility.py
+++ b/src/qutip_qip/decompose/_utility.py
@@ -169,27 +169,41 @@ def gray_code_gate_info(index_of_state_1, index_of_state_2, num_qubits):
 
     # find number of ones in each gray code step, this will be used
     # to keep track of control values.
-    one_bit_count = []
+    # one_bit_count = []
+    # for i in range(num_steps+1):
+    #    check_bit = gray_code[i].count("1")
+    #    one_bit_count.append(check_bit)
+
+    # make a dictionary of an array of binary values
+    # find position of 1's
+    input_binary_values = {}
+
     for i in range(num_steps+1):
-        check_bit = gray_code[i].count("1")
-        one_bit_count.append(check_bit)
-
-    gate_dictionary = {}
-    for i in range(num_steps):
+        bit_array_at_i = []
         a = gray_code[i]
-        b = gray_code[i+1]
-        for j in range(num_qubits):
-            gate_controls = []
-            gate_target = []
-            gate_control_value = []
-            while a[j] == b[j]:
-                partial_gate_control = j
-                gate_controls.append(partial_gate_control)
-                gate_control_value.append(a[j])
+        for j in range(3):
+            bit_array_at_i.append(a[j])
+
+        input_binary_values[i] = bit_array_at_i
+
+    # compare the array values
+    step_iteration_dictionary = {}
+    for i in range(num_steps):
+        a = input_binary_values[i]
+        b = input_binary_values[i+1]
+        controls = []
+        control_value = []
+        target = []
+        all_together = {}
+        for j in range(3):
+            if a[j] == b[j]:
+                controls.append(j)
+                control_value.append(a[j])
             else:
-                partial_gate_target = j
-                gate_target.append(partial_gate_target)
+                target.append(j)
+            all_together['controls ='] = controls
+            all_together['control_value ='] = control_value
+            all_together['targets ='] = target
+        step_iteration_dictionary[i] = all_together
 
-        gate_dictionary[i] = [gate_controls, gate_control_value, gate_target]
-
-    return(gate_dictionary)
+    return(step_iteration_dictionary)
--- a/src/qutip_qip/decompose/decompose_general_qubit_gate.py
+++ b/src/qutip_qip/decompose/decompose_general_qubit_gate.py
@@ -4,9 +4,9 @@ import cmath
 from qutip import Qobj
 from qutip_qip.decompose._utility import (
     check_gate,
-    MethodError,
+    MethodError, _gray_code_steps, gray_code_gate_info
 )
-
+import matplotlib.pyplot as plt
 import warnings
 from qutip_qip.circuit import Gate
 from qutip_qip.operations import controlled_gate
@@ -17,8 +17,91 @@ from .decompose_single_qubit_gate import decompose_one_qubit_gate
 warnings.filterwarnings("ignore", category=UserWarning)
 
 
-def _decompose_to_two_level_arrays(input_gate, num_qubits):
+def plot_gray_code_grid(index_of_state_1, index_of_state_2, num_qubits):
+    """ Plots the difference between each step of a gray code sequence.
+
+    .. note::
+
+        If you would prefer to plot the full gray code sequence, specify the
+        first and last indices of full sequence. For a n qubit state, the
+        indices of interest will be 0 and :math:`n^2-1` respectively.
+
+    In order to create a plot for a 5 qubit two-level unitary with nontrivial
+    values 00101 and 10001,
+
+    .. code-block:: python
+
+        plot_gray_code_grid(5,17,5)
+        plt.show()
+
+    Parameters
+    -----------
+    index_of_state_i: int
+        The non-trivial indices in a two-level unitary matrix gate.
+
+    num_qubits:
+        Number of qubits being acted upon by the quantum gate.
+
+    Returns
+    -------
+    matplotlib.image.AxesImage
+        The preferred plot for mapping from one binary state to another in a
+    gray code sequence.
+    """
+    # create a gray code sequence
+    gray_code_and_num_steps = _gray_code_steps(
+        index_of_state_1, index_of_state_2, num_qubits)
+    gray_code_sequence = gray_code_and_num_steps[0]
+    num_steps = gray_code_and_num_steps[-1]
+
+    # create a dictionary for creating an array later
+    input_binary_values = {}
+    for i in range(num_steps+1):
+        bit_array_at_i = []
+        a = gray_code_sequence[i]
+        for j in range(num_qubits):
+            bit_array_at_i.append(int(a[j]))
+            input_binary_values[i] = bit_array_at_i
+
+    # array is created for imshow
+    input_binary_array = np.full([num_steps+1, num_qubits], None, dtype=float)
+    for i in range(num_steps+1):
+        input_binary_array[i] = np.array([input_binary_values[i]])
+
+    # size_for_plot = 2*num_qubits  # figsize=(size_for_plot, size_for_plot)
+    # figsize could be used to adjust the length of plot based on the size
+    # of input
+
+    # information for plot
+    fig, ax = plt.subplots(1)
+    y_axis = list(range(len(gray_code_sequence)))
+    x_axis = list(range(num_qubits))
+    ax.set_xticks(x_axis)
+    ax.set_yticks(y_axis)
+    ax.set_yticklabels(list(reversed(gray_code_sequence)))
+    plt.xlabel("Number of Qubits")
+    plt.ylabel("Binary")
+    plt.title('Zero Code Sequence')
+    im = ax.imshow(
+        input_binary_array, cmap='binary', extent=[
+            x_axis[0], x_axis[-1], y_axis[0], y_axis[-1]], aspect='auto')
+
+    return(im)
+
+
+def _decompose_to_two_level_arrays(input_gate, num_qubits, expand=True):
     """Decompose a general qubit gate to two-level arrays.
+
+    Parameters
+    -----------
+    input_gate : :class:`qutip.Qobj`
+        The gate matrix to be decomposed.
+    num_qubits : int
+        Number of qubits being acted upon by the input_gate
+    expand : True
+        Default parameter to return the output as full two-level Qobj. If
+    `expand = False` then the function returns a tuple of index information
+    and a 2 x 2 Qobj for each gate. The Qobj are returned in reversed order.
     """
     check_gate(input_gate, num_qubits)
     input_array = input_gate.full()
@@ -31,7 +114,6 @@ def _decompose_to_two_level_arrays(input_gate, num_qubits):
             new_index = [i, j]
             index_list.append(new_index)
 
-    # index_list = [[0, 1], [0, 2], [0, 3], [1, 2], [1, 3]]
     for i in range(len(index_list)-1):
         index_1, index_2 = index_list[i]
 
@@ -40,10 +122,11 @@ def _decompose_to_two_level_arrays(input_gate, num_qubits):
         a_star = np.conj(a)
         b = input_array[index_2][index_1]
         b_star = np.conj(b)
-        norm_constant = cmath.sqrt(np.absolute(a*a_star)+np.absolute(b*b_star))
+        norm_constant = cmath.sqrt(
+            np.absolute(a*a_star)+np.absolute(b*b_star))
 
-        # Create identity array and then replace with above values for index_1
-        # and index_2
+        # Create identity array and then replace with above values for
+        # index_1 and index_2
         U_two_level = np.identity(2**num_qubits, dtype=complex)
         U_two_level[index_1][index_1] = a_star/norm_constant
         U_two_level[index_2][index_1] = b/norm_constant
@@ -55,10 +138,68 @@ def _decompose_to_two_level_arrays(input_gate, num_qubits):
 
         # U dagger to calculate the gates
         U__two_level_dagger = np.transpose(np.conjugate(U_two_level))
-        U__two_level_dagger = Qobj(U__two_level_dagger, dims=[[2] * num_qubits] * 2)
         array_list.append(U__two_level_dagger)
 
     # for U6 - multiply input array by U5 and take dagger
     U_last_dagger = input_array
-    array_list.append(Qobj(U_last_dagger, dims=[[2] * num_qubits] * 2))
-    return(array_list)
+    array_list.append(U_last_dagger)
+
+    if expand is True:
+        array_list_with_qobj = []
+        for i in reversed(range(len(index_list))):
+            U_two_level_array = array_list[i]
+            array_list_with_qobj.append(Qobj(
+                U_two_level_array, dims=[[2] * num_qubits] * 2))
+        return(array_list_with_qobj)
+    else:
+        compact_U_information = []
+        for i in reversed(range(len(index_list))):
+            U_non_trivial = np.full([2, 2], None, dtype=complex)
+            index_info = []
+            U_index_together = []
+
+            # create index list
+            index_1, index_2 = index_list[i]
+            index_info = [index_1, index_2]
+            U_index_together.append(index_info)
+
+            # create 2 x 2 arrays
+            U_two_level = array_list[i]
+            U_non_trivial[0][0] = U_two_level[index_1][index_1]
+            U_non_trivial[1][0] = U_two_level[index_2][index_1]
+            U_non_trivial[0][1] = U_two_level[index_1][index_2]
+            U_non_trivial[1][1] = U_two_level[index_2][index_2]
+            U_index_together.append(
+                Qobj(U_non_trivial, dims=[[2] * 1] * 2))
+
+            compact_U_information.append(U_index_together)
+
+        return(compact_U_information)
+
+
+def _two_level_gate_info(input_gate, num_qubits):
+    """ From the output of two level arrays, create a tuple of controlled gate
+    information.
+    """
+    array_decompose = _decompose_to_two_level_arrays(
+        input_gate, num_qubits, expand=False)
+
+    full_index_list = []
+    for i in range(len(array_decompose)):
+        index_list = array_decompose[i][0]
+        full_index_list.append(index_list)
+
+    gate_keys = []
+    for i in reversed(range(len(array_decompose))):
+        gate_string = 'gate'
+        gate_string = gate_string + str(i)
+        gate_keys.append(gate_string)
+
+    gate_dictionary = {}
+    for i in range(len(array_decompose)):
+        index_1, index_2 = full_index_list[i]
+        gate_i_info = gray_code_gate_info(index_1, index_2, num_qubits)
+        gate_key = gate_keys[i]
+        gate_dictionary[gate_key] = gate_i_info
+
+    return(gate_dictionary)
--- a/src/qutip_qip/decompose/decompose_general_qubit_gate.py
+++ b/src/qutip_qip/decompose/decompose_general_qubit_gate.py
@@ -203,13 +203,24 @@ def _two_level_gate_info(input_gate, num_qubits):
         gate_key = gate_keys[i]
         gate_dictionary[gate_key] = gate_i_info
 
+    # this function output is not what's expected because the output of
+    # gray_code_gate_info is not what's expected.
+    # for this reason, there's no test designed for it, once the error in
+    # gray_code_gate_info is corrected, a test will be added
     return(gate_dictionary)
 
 
 def _sqrt_of_1_qubit_array(input_array):
     """ Finds the square root of a 1 qubit gate for decomposing a multi-qubit
     array into smaller controlled CNOT and 2-qubit two-level unitary.
+
+    # based on Lemma 7.5 of https://arxiv.org/abs/quant-ph/9503016
     """
+    # final lemma 7.5 function is not defined - will be dfined if needed
+
+    # This method is for when 1 qubit is not special unitary but the single
+    # qubit decomposition functions defined already make a non-special unitary
+    # output into a su(2) by the determinant.
     check_gate(Qobj(input_array, dims=[[2] * 1] * 2), 1)
     # method taken
     # from https://en.wikipedia.org/wiki/Square_root_of_a_2_by_2_matrix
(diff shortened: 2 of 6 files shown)


src/qutip_qip/decompose/_utility.py
```diff
@@ -29,3 +30,166 @@ def check_gate(gate, num_qubits):
...
+            gate_control_value = []
+            while a[j] == b[j]:
+                partial_gate_control = j
+                gate_controls.append(partial_gate_control)
+                gate_control_value.append(a[j])
+            else:
+                partial_gate_target = j
+                gate_target.append(partial_gate_target)
+
+        gate_dictionary[i] = [gate_controls, gate_control_value, gate_target]
```
purva-thakre, line 193: @hodgestar @BoxiLi  This function fails to return an output because it gets stuck in an infinite loop.  Not sure what might be causing it.

src/qutip_qip/decompose/decompose_general_qubit_gate.py
```diff
@@ -0,0 +1,64 @@
...
+from qutip_qip.circuit import Gate
+from qutip_qip.operations import controlled_gate
+
+from .decompose_single_qubit_gate import decompose_one_qubit_gate
+
+# for unknown labels in two level gates
+warnings.filterwarnings("ignore", category=UserWarning)
+
+
+def _decompose_to_two_level_arrays(input_gate, num_qubits):
```
purva-thakre, line 20: The output of this has been checked for two qubits in a previous PR. The change here has been to find the `index_list` first. 

The output has not been fully checked. The length of the output was checked and not their product. 

The output will be checked once an easier way to define gate objects for these arrays has been added. 

- 2 qubits - array length is 6
- 3 qubits - array length is 28
- 4 qubits  - array length is 120

src/qutip_qip/decompose/_utility.py
BoxiLi: If it runs infinitely then there is likely a bug in the `while` loop.

BoxiLi: Could you provide an example input for this function?

BoxiLi: It might be wrong but in the `while a[j] == b[j]` loop you never change `a` and `b` so `a[j] == b[j]` is always true and the while loop never stops?

BoxiLi:
By the way, don't pay attention to the Cognitive Complexity requirement in the codeclimate at the moment, I'll raise that bar in a separate PR.

src/qutip_qip/decompose/_utility.py
purva-thakre: > Could you provide an example input for this function?

Apologies for the delay. I have managed to correct my function definition. You should be able to see the code for below tomorrow morning (I haven't finished adding tests yet). 

Consider some example 3 qubit two-level unitary acting non-trivially on `010` and `101`.

The full binary sequence can be found by using  `_binary_sequence(3)` which will output `['000', '001', '010', '011', '100', '101', '110', '111']`

The full gray code sequence is `['000', '001', '011', '010', '110', '111', '101', '100']` by using ` _gray_code_sequence(3)` or in terms of index positions of binary sequence as `[0, 1, 3, 2, 6, 7, 5, 4]`. 

Thus, it can be seen that our examples indices `010` and `101` can be translated to binary indices as 2 and 5. The partial gray code sequence and the total number of steps can be found via `_gray_code_steps(2, 5, 3)` which outputs a tuple of the partial gray code sequence and number of steps. 

```
(['010', '110', '111', '101'], 3)
```

Now, our goal is to find the information about CNOT gates for mapping from  `010` to `101` via `gray_code_gate_info(2, 5, 3)`. 

The output of this function is in the form of a dictionary with key as iteration value. 
```
{0: {'controls =': [1, 2], 'control_value =': ['1', '0'], 'targets =': [0]}, 1: {'controls =': [0, 1], 'control_value =': ['1', '1'], 'targets =': [2]}, 2: {'controls =': [0, 2], 'control_value =': ['1', '1'], 'targets =': [1]}}
```

purva-thakre:
@BoxiLi Haven't pushed the tests yet. But all the functions do output what they are supposed to (except for the plot).

src/qutip_qip/decompose/_utility.py
```diff
@@ -29,3 +30,180 @@ def check_gate(gate, num_qubits):
...
     if gate.dims != [[2] * num_qubits] * 2:
         raise ValueError(f"Input is not a unitary on {num_qubits} qubits.")
+
+
+def _binary_sequence(num_qubits):
+    """ Defines the binary sequence list for basis vectors of a n-qubit gate.
+
+    The string at index `i` is also the row/column index for a basis vector in
+    a numpy array.
+    """
```
BoxiLi, line 40: There is a compact python function that can generate binary combinations. https://stackoverflow.com/questions/14931769/how-to-get-all-combination-of-n-binary-value
But I think yours is more memory friendly as there is no nested list. Actually, a recursive function looks neater: https://stackoverflow.com/questions/64890117/what-is-the-best-way-to-generate-all-binary-strings-of-the-given-length-in-pytho. 

Just FYI, no need to change unless there is a big difference in efficiency.

BoxiLi: However, if this function is going to be called many times with the same value, it might be worth to cache the result.

Actually when you compute it for `num_qubits=N`, you already calculated all result for `num_qubits<N`. That's why a recursive function here might be more useful to cache the result.

But maybe not of a very high priority.

src/qutip_qip/decompose/_utility.py
```diff
@@ -29,3 +30,180 @@ def check_gate(gate, num_qubits):
...
+            zero_append_sequence = ['0' + x for x in old_sequence]
+            full_binary_sequence.extend(zero_append_sequence)
+            one_append_sequence = ['1' + x for x in old_sequence]
+            full_binary_sequence.extend(one_append_sequence)
+            old_sequence = full_binary_sequence
+
+    return(full_binary_sequence)
+
+
+def _gray_code_sequence(num_qubits, output_form=None):
```
BoxiLi, line 58: I feel like there is a more efficient way to generate the sequence without first compute a list of all binaries and then copy. But the overhead compared to the overall decomposition is not that big. So let's keep that in mind with a comparably low priority.

src/qutip_qip/decompose/_utility.py
```diff
@@ -29,3 +30,180 @@ def check_gate(gate, num_qubits):
...
+    num_qubits:
+        Number of qubits being acted upon by the quantum gate.
+    """
+    # This function is still incomplete.
+
+
+def gray_code_gate_info(index_of_state_1, index_of_state_2, num_qubits):
+    """ Finds information about control and targets for CNOT gate in a gray
+    code sequence.
+    """
```
BoxiLi, line 164: Could you give the parameters or an example of a valid input? I tried a few but always get an error 
```At least one of the input indices is invalid.```

Comment (BoxiLi) [reply src/qutip_qip/decompose/_utility.py:164]: Ah, you already give one above!

Comment (purva-thakre) [reply src/qutip_qip/decompose/_utility.py:164]: I resolved a previous conversation but[ here's a linked description ](https://github.com/qutip/qutip-qip/pull/86#discussion_r677942416)(don't know if you saw it yet).

File: src/qutip_qip/decompose/_utility.py (hunk 3 above)
Comment (BoxiLi) [reply src/qutip_qip/decompose/_utility.py:40]: Ok, it seems that for each unitary decomposition, you only need to call this once. So the overhead is ok.

File: src/qutip_qip/decompose/_utility.py (hunk 5 above)
Comment (purva-thakre) [reply src/qutip_qip/decompose/_utility.py:164]: Here's a 4 qubit example : 
image

**Comment by purva-thakre** [comment]
@BoxiLi Here's an example output for decomposing a random 3 qubit unitary into two-level arrays. Earlier when I was referring to finding the non-trivial indices, I was talking about the output arrays from `_decompose_to_two_level_arrays`. Right now, I am comparing these to an identity matrix to find the indices that make them two-level unitaries. 
```python
np.set_printoptions(precision=3)
np.set_printoptions(suppress=True)
two_qubit = rand_unitary(8,dims=[[2] * 3] * 2)
array_decompose = _decompose_to_two_level_arrays(two_qubit, 3)
print(*array_decompose, sep='\n\n')
```
```
Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = False
Qobj data =
[[ 0.964-0.254j  0.067+0.033j  0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.067-0.033j -0.964-0.254j  0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 0.759+0.j     0.   +0.j    -0.127-0.638j  0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [-0.127+0.638j  0.   +0.j    -0.759+0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 0.868+0.j     0.   +0.j     0.   +0.j     0.485+0.105j  0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.485-0.105j  0.   +0.j     0.   +0.j    -0.868+0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 0.797+0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.304-0.522j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.304+0.522j  0.   +0.j     0.   +0.j     0.   +0.j    -0.797+0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 0.953-0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
  -0.087+0.292j  0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [-0.087-0.292j  0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
  -0.953-0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 0.967+0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j    -0.226+0.114j  0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [-0.226-0.114j  0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j    -0.967+0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 0.84 +0.j    0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j
   0.   +0.j    0.   +0.j   -0.185-0.51j]
 [ 0.   +0.j    1.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j
   0.   +0.j    0.   +0.j    0.   +0.j  ]
 [ 0.   +0.j    0.   +0.j    1.   +0.j    0.   +0.j    0.   +0.j
   0.   +0.j    0.   +0.j    0.   +0.j  ]
 [ 0.   +0.j    0.   +0.j    0.   +0.j    1.   +0.j    0.   +0.j
   0.   +0.j    0.   +0.j    0.   +0.j  ]
 [ 0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j    1.   +0.j
   0.   +0.j    0.   +0.j    0.   +0.j  ]
 [ 0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j
   1.   +0.j    0.   +0.j    0.   +0.j  ]
 [ 0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j
   0.   +0.j    1.   +0.j    0.   +0.j  ]
 [-0.185+0.51j  0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j
   0.   +0.j    0.   +0.j   -0.84 +0.j  ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = False
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j    -0.053+0.748j  0.375-0.545j  0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.375+0.545j  0.053+0.748j  0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.941+0.j     0.   +0.j     0.298-0.163j  0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.298+0.163j  0.   +0.j    -0.941+0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.895+0.j     0.   +0.j     0.   +0.j     0.161-0.416j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.161+0.416j  0.   +0.j     0.   +0.j    -0.895+0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.997+0.j     0.   +0.j     0.   +0.j     0.   +0.j
  -0.   -0.077j  0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j    -0.   +0.077j  0.   +0.j     0.   +0.j     0.   +0.j
  -0.997+0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.967+0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.231+0.107j  0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.231-0.107j  0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j    -0.967+0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.979+0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j    -0.1  +0.177j]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j    -0.1  -0.177j  0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j    -0.979+0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = False
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j    -0.599+0.088j  0.657+0.449j  0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.657-0.449j  0.599+0.088j  0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.608-0.j     0.   +0.j     0.794+0.003j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.794-0.003j  0.   +0.j    -0.608-0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.724+0.j     0.   +0.j     0.   +0.j
  -0.637+0.265j  0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j    -0.637-0.265j  0.   +0.j     0.   +0.j
  -0.724+0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.015+0.019j  0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.015-0.019j  0.   +0.j     0.   +0.j
   0.   +0.j    -1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.978+0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j    -0.192-0.087j]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j    -0.192+0.087j  0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j    -0.978+0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = False
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j    -0.02 -0.863j -0.323+0.389j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j    -0.323-0.389j  0.02 -0.863j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.995+0.j     0.   +0.j
   0.102+0.016j  0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.102-0.016j  0.   +0.j
  -0.995+0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.873-0.j     0.   +0.j
   0.   +0.j    -0.094-0.478j  0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j    -0.094+0.478j  0.   +0.j
   0.   +0.j    -0.873-0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.827+0.j     0.   +0.j
   0.   +0.j     0.   +0.j    -0.237+0.509j]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   1.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j    -0.237-0.509j  0.   +0.j
   0.   +0.j     0.   +0.j    -0.827+0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = False
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j    -0.494-0.809j
  -0.319+0.015j  0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j    -0.319-0.015j
   0.494-0.809j  0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 1.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j
   0.   +0.j    0.   +0.j    0.   +0.j  ]
 [ 0.   +0.j    1.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j
   0.   +0.j    0.   +0.j    0.   +0.j  ]
 [ 0.   +0.j    0.   +0.j    1.   +0.j    0.   +0.j    0.   +0.j
   0.   +0.j    0.   +0.j    0.   +0.j  ]
 [ 0.   +0.j    0.   +0.j    0.   +0.j    1.   +0.j    0.   +0.j
   0.   +0.j    0.   +0.j    0.   +0.j  ]
 [ 0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j    0.842+0.j
   0.   +0.j    0.327-0.43j  0.   +0.j  ]
 [ 0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j
   1.   +0.j    0.   +0.j    0.   +0.j  ]
 [ 0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j    0.327+0.43j
   0.   +0.j   -0.842+0.j    0.   +0.j  ]
 [ 0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j
   0.   +0.j    0.   +0.j    1.   +0.j  ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 1.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j
   0.   +0.j    0.   +0.j    0.   +0.j  ]
 [ 0.   +0.j    1.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j
   0.   +0.j    0.   +0.j    0.   +0.j  ]
 [ 0.   +0.j    0.   +0.j    1.   +0.j    0.   +0.j    0.   +0.j
   0.   +0.j    0.   +0.j    0.   +0.j  ]
 [ 0.   +0.j    0.   +0.j    0.   +0.j    1.   +0.j    0.   +0.j
   0.   +0.j    0.   +0.j    0.   +0.j  ]
 [ 0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j    0.966+0.j
   0.   +0.j    0.   +0.j   -0.148+0.21j]
 [ 0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j
   1.   +0.j    0.   +0.j    0.   +0.j  ]
 [ 0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j
   0.   +0.j    1.   +0.j    0.   +0.j  ]
 [ 0.   +0.j    0.   +0.j    0.   +0.j    0.   +0.j   -0.148-0.21j
   0.   +0.j    0.   +0.j   -0.966+0.j  ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = False
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.13 +0.76j  -0.247-0.587j  0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
  -0.247+0.587j -0.13 +0.76j   0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     1.   +0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = True
Qobj data =
[[ 1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j     0.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     1.   +0.j
   0.   +0.j     0.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.964+0.j     0.   +0.j     0.125-0.233j]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.   +0.j     1.   +0.j     0.   +0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j
   0.125+0.233j  0.   +0.j    -0.964+0.j   ]]

Quantum object: dims = [[2, 2, 2], [2, 2, 2]], shape = (8, 8), type = oper, isherm = False
Qobj data =
[[ 1.   +0.j    -0.   +0.j     0.   -0.j     0.   -0.j    -0.   -0.j
  -0.   -0.j     0.   -0.j     0.   -0.j   ]
 [-0.   +0.j     1.   +0.j    -0.   +0.j     0.   +0.j     0.   +0.j
  -0.   -0.j    -0.   +0.j    -0.   -0.j   ]
 [ 0.   +0.j     0.   -0.j     1.   +0.j    -0.   -0.j    -0.   +0.j
   0.   -0.j    -0.   -0.j    -0.   +0.j   ]
 [-0.   +0.j    -0.   +0.j     0.   -0.j     1.   +0.j    -0.   -0.j
  -0.   +0.j     0.   -0.j     0.   -0.j   ]
 [ 0.   +0.j     0.   +0.j     0.   +0.j    -0.   -0.j     1.   -0.j
  -0.   -0.j    -0.   +0.j    -0.   +0.j   ]
 [ 0.   -0.j     0.   +0.j    -0.   -0.j     0.   +0.j    -0.   +0.j
   1.   +0.j    -0.   +0.j     0.   -0.j   ]
 [-0.   +0.j     0.   +0.j     0.   +0.j     0.   +0.j    -0.   +0.j
   0.   +0.j    -0.745+0.14j  -0.016-0.652j]
 [-0.   +0.j     0.   +0.j    -0.   +0.j     0.   -0.j    -0.   -0.j
   0.   +0.j     0.622+0.196j  0.341-0.677j]]
```

BoxiLi:
My suggestion is that if we can represent all the information in a compact form, like  `(single_qubit_unitary_matrix, control_info)`, then we should never generate the full matrix. Either save them in this compact form or create a `Gate` class to save them. Generating the full matrix and then identify the useful elements are both very expensive. They should be avoided whenever possible.

It is like entropy in physics, it is easy to go from low entropy to high entropy (Gate->full matrix) but "very hard" to go the other way around :)

Of course, when you want to test the result you will like them to be in the matrix or `Qobj` form. This should be done byan external function outside the decomposition routine, like the two functions we added to `QubitCircuit` in the single-qubit gate PR.

src/qutip_qip/decompose/_utility.py
purva-thakre: @BoxiLi Would it be better if we push this later down the line ? I can create an issue for this to remember to work on it. This is because I will need to look over a couple of references to find the most efficient method. 

If it's preferred, I could also add some of the referenced methods below and provide a choice for how the gray codes are generated. 

- References on page 8-9 of https://arxiv.org/pdf/1706.00852.pdf
- References on page 6 of https://arxiv.org/pdf/1608.06001.pdf
- Overview of gray codes - [Ref 1](https://tel.archives-ouvertes.fr/tel-01203339/document), [Ref 2](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.109.6592&rep=rep1&type=pdf)

purva-thakre:
> Either save them in this compact form or create a Gate class to save them. Generating the full matrix and then identify the useful elements are both very expensive. 

I think I am going to keep the array output as is for the decomposition to two level arrays. I will try to define a function to generate  a compact form output `(single_qubit_unitary_matrix, control_info)` like you suggested. 

I will need to look over how a `Gate` class could be created to save these. I haven't worked a lot on defining classes, sub-classes and their respective methods. A google search does point me towards the right references for it.

BoxiLi:
> I think I am going to keep the array output as is for the decomposition to two level arrays because I tend to check their output product quite often for larger number of qubits. I will try to define a function to generate a compact form output (single_qubit_unitary_matrix, control_info) like you suggested. 

Surely you can go whichever way you find easier before we finalize things. My experience is that, during the calculation, you will very likely first get the compact information. Like in the following code.

https://github.com/purva-thakre/qutip-qip/blob/cf96bc0b30393bf377f79eadf3e7b841ffabfa8a/src/qutip_qip/decompose/decompose_general_qubit_gate.py#L45-L59

You first get the single-qubit-unitary consisting of ` a_star/norm_constant` etc and `index_1`, `index_2`. So you can just save them in a compact form:
```
compact_info = (
np.array([[a_star/norm_constant, b/norm_constant],[b_star/norm_constant, -a/norm_constant]]),
index_1, 
index_2
)
``` 
And let the function return it if only the compact form is needed. For your convenience, you can add a parameter `expand` to your function 
```_decompose_to_two_level_arrays(input_gate, num_qubits, expand=True)```
If it is true, you don't return the compact form but continue computing the array as you did so that you can compare the result. I would say it is easier than iterating the array to extract compact information.

**Comment by BoxiLi** [comment]
For your purpose, defining a class is equivalent to saving the compact info. If you want to focus on the decomposition algorithm, the compact info representation is sufficient.

The most basic use case of class is to save a bunch of information together under the same name. You can understand it same as a list. Just instead of access each element by the index. you can access them by an attribute name, like `Gate.controls`, `Gate.targets`. So in this sense, it is the same as a compact information carrier.

We don't need to support two-level gates in `QubitCircuit` right now.  Adding a class definition is not difficult and pretty easy to add later if you prefer.

**Comment by purva-thakre** [comment]
> The most basic use case of class is to save a bunch of information together under the same name

Yep, I am aware of this. My issue is not knowing which attributes would be helpful and then going back and changing them too many times. 

For the array output, I will create a `expand=True` parameter like you suggested. 

But for the `Gate` object, there will need to be a sub-class for two-level unitaries so that there's an easy way to label these controlled gates, provide target/control and the control values.

**Comment by BoxiLi** [comment]
> But for the Gate object, there will need to be a sub-class for two-level unitaries so that there's an easy way to label these controlled gates, provide target/control and the control values.

That also makes sense. Do you need to show them in the circuit plotting? If that is the case, then indeed you have to define a class object. I have to admit that currently define a gate subclass is not very straightforward as  Gate is still a bit entangled with QubitCircuit. Let me know if you have any trouble.

Basically, you can always define a method in the subclass that overwrite the method in the parent class. So you have plenty of flexibility there. That is the magic of class.

File: src/qutip_qip/decompose/decompose_general_qubit_gate.py [hunk 6]
```diff
@@ -0,0 +1,466 @@
...
+
+
+
+
+def _decompose_multi_cnot_further():
+    """ Decomposes output of lemma 7.3 even further by using only 1 particular
+    cnot with different controls.
+
+    # based on lemma 7.2 of https://arxiv.org/abs/quant-ph/9503016
+    """
```
purva-thakre, line 403: Some of these functions have incomplete code because locally their output is incorrect. Want to correct the error before committing.

tests/decomposition_functions/test_utility.py
```diff
@@ -48,3 +53,114 @@ def test_check_gate_unitary_input(unitary):
...
+    expected_num_steps = 9
+    expected_gray_code = ['000', '001', '011', '010', '110', '111']
+    calc_num_steps_sequence = _gray_code_steps(0, 7, 3)
+    calc_num_steps = calc_num_steps_sequence[1]
+    calc_gray_code = calc_num_steps_sequence[0]
+    assert np.equal(calc_num_steps, expected_num_steps)
+    assert expected_gray_code == calc_gray_code
+
+
+@pytest.mark.xfail
```
purva-thakre, line 127: This test fails. See comment for more info

purva-thakre:
closing this for now. 

Will add a new general decomposition method PR later this week.

PR #88, opened 2021-07-30
Support qutip v5

BoxiLi:
Add support for qutip v5 and the corresponding tests. This will keep the development compatible with `dev.major`.

Some tests are skipped mainly because of the change in `QobjEvo`. The interpretation of coefficients needs to be reworked. But with the new functionality, `Pulse` can now be defined in a simpler way. Those will be addressed in a different PR. This PR includes only minor fixes for the renaming in qutip such as `Options`->`SolverOptions` and the use of `np.testing.assert_all_close` with `Qobj`.

### DIFF
--- a/src/qutip_qip/device/cavityqed.py
+++ b/src/qutip_qip/device/cavityqed.py
@@ -144,10 +144,16 @@ class DispersiveCavityQED(ModelProcessor):
         Eliminate the auxillary modes like the cavity modes in cqed.
         """
         psi_proj = tensor(
-            [basis(self.num_levels, 0)]
-            + [identity(2) for n in range(self.num_qubits)]
-        )
-        return psi_proj.dag() * U * psi_proj
+            [basis(self.num_levels, 0)] +
+            [identity(2) for n in range(self.num_qubits)])
+        result = psi_proj.dag() * U * psi_proj
+        # In qutip 5 multiplication of matrices
+        # with dims [[1, 2], [2, 2]] and [[2, 2], [1, 2]]
+        # will give a result of
+        # dims [[1, 2], [1, 2]] instead of [[2], [2]].
+        if result.dims[0][0] == 1:
+            result = result.ptrace(list(range(len(self.dims)))[1:])
+        return result
 
     def load_circuit(self, qc, schedule_mode="ASAP", compiler=None):
         if compiler is None:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -54,13 +54,13 @@ jobs:
 
     - name: Install QuTiP from PyPI
       if: ${{ ! startsWith( matrix.qutip-version, '@') }}
-      run: pip install 'qutip${{ matrix.qutip-version }}'
+      run: python -m pip install 'qutip${{ matrix.qutip-version }}'
 
     - name: Install QuTiP from GitHub
+      if: ${{ startsWith( matrix.qutip-version, '@') }}
       run: |
         python -m pip install numpy scipy cython
-      if: ${{ startsWith( matrix.qutip-version, '@') }}
-      run: pip install 'git+https://github.com/qutip/qutip.git${{ matrix.qutip-version }}'
+        python -m pip install 'git+https://github.com/qutip/qutip.git${{ matrix.qutip-version }}'
 
     - name: Install qutip-qip
       # Installing in-place so that coveralls can locate the source code.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,31 +10,37 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.6, 3.9]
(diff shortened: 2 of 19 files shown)


src/qutip_qip/device/cavityqed.py
```diff
@@ -248,7 +248,14 @@ def eliminate_auxillary_modes(self, U):
...
             [identity(2) for n in range(self.num_qubits)])
-        return psi_proj.dag() * U * psi_proj
+        result = psi_proj.dag() * U * psi_proj
+        # In qutip 5 multiplication of matrices
+        # with dims [[1, 2], [2, 2]] and [[2, 2], [1, 2]]
+        # will give a result of
+        # dims [[1, 2], [1, 2]] instead of [[2], [2]].
+        if result.dims[0][0] == 1:
+            result = result.ptrace(list(range(len(self.dims)))[1:])
+        return result
```
BoxiLi, line 258: @jakelishman A quick question if you get time, is this the expected behaviour. The two examples below seem inconsistent on how dimensions are handled.
```python
>>> basis(2,0).dag() * basis(2,0)
(1+0j)

>>> tensor(basis(2,0).dag(), qeye(2)) * tensor(basis(2,0),qeye(2)) 
Quantum object: dims=[[1, 2], [1, 2]], shape=(2, 2), type='oper', isherm=True
Qobj data =
[[1. 0.]
 [0. 1.]]
```

jakelishman: This was intended. I think we also added a method called `Qobj.contract_subspaces` or something as well to handle the second behaviour if you wanted it?  (Not certain it actually got put in.)  At any rate, you certainly don't need to do `ptrace`; you can just modify the dimensions.

The point was that automatic Hilbert-space contraction in most cases caused more bugs down the line when reusing the objects, or when dealing with explicit Hilbert spaces of dimension 1 (that may appear programmatically - they're pretty useless in practice).  The contraction to a scalar depends on the operands being of `Qobj` type `bra` and `ket`.  Your example has type `oper`, so you get an `oper` back, and this doesn't contract any Hilbert spaces, because it couldn't tell if the first Hilbert space was `oper * oper` or `bra * ket` or even the forbidden `ket * ket`.

BoxiLi: > you either get a scalar back if the objects were exactly bra and ket, or you get a Qobj with the same rule for dimensions that everything always uses ([left.dims[0], right.dims[1]]). We had too many problems with attempts to guess when a dimension should be contracted that we just made it exactly consistent in all cases.

Ok, this rule also makes sense. Thanks!

BoxiLi:
@Ericgig I have a question about the interpolation in the new `coefficient` function. What is the difference between the scipy cubic interpolation and the new interpolation in `coeficient` of qutip5?

My goal is to generate a coefficient that 
- time region1: no pulse, idling for a fixed amount of time
- time region2: a non-zero pulse.

My understanding is that this can be achieved if I choose a zero coeff array for the first and last time steps in time region 1 (step number N=5 in the example below) . E.g. in `coeff_zero` and `tlist_zero` below. It works for qutip 4.6.

However, when I now switch to qutip5, it fails and I have to increase `N`. I'm perfectly fine with this but just wondering, is there any reason behind it or any caveat?

The code below is a small example showing the difference between the two interpolation functions.

```
from qutip import coefficient
import numpy as np
from scipy import interpolate, signal
import matplotlib.pyplot as plt

## idling
# an example that works only for scipy but not qutip 5
coeff_zero = np.array([0., 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,])
tlist_zero = np.array([0., 0.1, 0.2, 0.3, 0.4, 0.5, 4.5, 4.6, 4.7, 4.8, 4.9])

# an example that works for both
# coeff_zero = np.array([0., 0.,  0.,  0.,  0.,  0.,  0.,  0. ,  0.,  0.  , 0.,  0. ,  0.,  0. ,  0.,  0. ,])
# tlist_zero = np.array([0., 0.1, 0.2, 0.3, 0.4, 0.5, 4.5, 4.55, 4.6, 4.65, 4.7, 4.75, 4.8, 4.85, 4.9, 4.95])

## pulse
tlist_pulse = np.linspace(5, 7, 200)
coeff_pulse = signal.windows.get_window("hamming", len(tlist_pulse))

## full coefficient
coeff = np.concatenate([coeff_zero, coeff_pulse])
tlist = np.concatenate([tlist_zero, tlist_pulse])

## qutip 5 interpolation
inter_func = coefficient(coeff, tlist=tlist)
xline = np.linspace(0, 6.2, 200)
yline = [inter_func(t) for t in xline]
plt.plot(xline, yline, label="qutip.dev")

## scipy cubic interpolation
inter_func = interpolate.interp1d(
    tlist, coeff, kind="cubic", bounds_error=False, fill_value=0.0)
xline = np.linspace(0, 6.2, 200)
yline = [inter_func(t) for t in xline]
plt.plot(xline, yline, label="scipy")
plt.legend()
plt.show()
```

image

PR #89, opened 2021-08-01
Fix the leaking noise objects

BoxiLi:
A list of noise objects defined by the user is passed to the function. However, when adding T1 and T2 noise, the generated objects were added directly to the list. This changes the `noise_list` defined by the user and is a problem if the same `Processor` instance is used to perform more than one simulation. The T1,T2 noise objects from the previous run will accumulate in the `noise_list`. Therefore, we copy the list before add T1 T2 noise.


hodgestar (approved):
Good catch!

PR #90, opened 2021-08-11
General decomposition method

purva-thakre:
[GSoC Notebook - Deliverable.pdf](https://github.com/qutip/qutip-qip/files/7031142/GSoC.Notebook.-.Deliverable.pdf)

Tutorial of partially working code 
- full final output cannot be checked right now
- Last output in the notebook is not correct due to Toffoli gates being inserted improperly - the circuit should not have introduced ancillas - see figure in [`Decomposition scheme for qubits greater than 2`](https://purva-thakre.github.io/purva-blog/gsoc/qutip/final-gsoc/)

**New To-Do :**

- [ ] Change output of [`_gray_code_gate_info` ](https://github.com/qutip/qutip-qip/pull/90/files#diff-9b635d4decbf8dade1e4507b1a75a5954ab381fa8b3ebc606dc9d2737874e979R161)
        - do not combine into forward gray code and backward gray code
        - return gate info for gates in forward gray code info upto the two-level gate
        - information for two-level gate is returned in another list
- [ ] After above change, [`_paulix_for_control_0`](https://github.com/qutip/qutip-qip/pull/90/files#diff-9b635d4decbf8dade1e4507b1a75a5954ab381fa8b3ebc606dc9d2737874e979R295) will not need 2 separate loops depending on if there's only 1 controlled gate or it's a forward/backward code. 
       - it is important to keep in mind whatever gates are determined for forward gray code list, the backward gray code applies the same gates in a reverse order. (this could be done by checking if the forward_gray_code list is empty or not) 
- [ ] Add a function to iterate over full output of `_paulix_for_control_0` and add CNOT gates for n = 2
- [ ] After the output of all possible gates is a list, introduce a function to iterate over said list of forward gray code and add CNOT gates. 
       - [ ] If the number of qubits is n > 2, choose to stop decomposition at toffoli or go further (provide option)
       - [ ] If the chosen decomposition method is via using ancilla qubits then the gate target can only be last qubit - if we allow the first qubit to be a target, the number of ancillas being introduced doubles.
- [ ] The output of [`_two_q_CNOT_gates`](https://github.com/qutip/qutip-qip/pull/90/files#diff-4add9955651bf88c9e275a01f0238c95ade46f084379817c7d077227f85266f8R524) should be a dictionary of lists only - not nested dictionaries and lists
- [ ] Provide option [of decomposition scheme](https://github.com/qutip/qutip-qip/pull/90/files#diff-4add9955651bf88c9e275a01f0238c95ade46f084379817c7d077227f85266f8R915) for n = 1
- [ ] Provide option [of decomposition scheme](https://github.com/qutip/qutip-qip/pull/90/files#diff-4add9955651bf88c9e275a01f0238c95ade46f084379817c7d077227f85266f8R915) for n = 3 - keep toffoli undecomposed or not
- [ ] Provide option [of decomposition scheme](https://github.com/qutip/qutip-qip/pull/90/files#diff-4add9955651bf88c9e275a01f0238c95ade46f084379817c7d077227f85266f8R915) for n > 3 after the recursion scheme and ancilla outputs are same as input gate - another option to further decompose toffoli
- [ ] Define recursion scheme to decompose without ancillas
- [ ] Use recursion scheme to decompose n-qubit toffoli without ancillas (n>3)

### DIFF
--- a/src/qutip_qip/decompose/_utility.py
+++ b/src/qutip_qip/decompose/_utility.py
@@ -1,5 +1,8 @@
 from qutip import Qobj
 import numpy as np
+import copy
+from qutip_qip.circuit import Gate
+
 
 class MethodError(Exception):
     """When invalid method is chosen, this error is raised."""
@@ -96,7 +99,10 @@ def _gray_code_steps(index_of_state_1, index_of_state_2, num_qubits):
     unitary matrix. Here, the inputs are their respective indices in a quantum
     gate's array.
     This function finds the number of steps between both states when only 1 bit
-    can be changed at each step.
+    can be changed at each step. In addition, a partial sequence going from
+    basis state 1 to state 2 and another partial sequence mapping back from
+    state 2 to state 1. The last gray sequence value in forward mapping is used
+    to define the two-level unitary.
     """
     # check array values are not the same
     try:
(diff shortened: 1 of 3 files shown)


src/qutip_qip/decompose/decompose_general_qubit_gate.py
```diff
@@ -0,0 +1,305 @@
...
+            U_non_trivial[1][1] = U_two_level[index_2][index_2]
+            U_index_together.append(
+                Qobj(U_non_trivial, dims=[[2] * 1] * 2))
+
+            compact_U_information.append(U_index_together)
+
+        return(compact_U_information)
+
+
+def _two_qubit_lastq_target(last_two_level_array_info):
```
purva-thakre, line 105: There's a general function which recursively decomposes upto when this function can be used.

src/qutip_qip/decompose/decompose_general_qubit_gate.py
```diff
@@ -0,0 +1,305 @@
...
+        B2,
+        X_gate,
+        CNOT_gate,
+        X_gate,
+        C,
+        phase_gate]
+    return(final_gate_list)
+
+
+def _two_qubit_firstq_target(last_two_level_array_info):
```
purva-thakre, line 180: Same as above Lemma 7.9 method but with controls and targets flipped - first qubit is target and all others are controls.

src/qutip_qip/decompose/decompose_general_qubit_gate.py
```diff
@@ -0,0 +1,305 @@
...
+        B2,
+        X_gate,
+        CNOT_gate,
+        X_gate,
+        C,
+        phase_gate]
+    return(final_gate_list)
+
+
+def _find_index_for_firstq_target(two_level_array_output, num_qubits):
```
purva-thakre, line 255: This is for the case when the two-level array does not have first or last qubit as target.

src/qutip_qip/decompose/decompose_general_qubit_gate.py
```diff
@@ -0,0 +1,305 @@
...
+    ind_for_0_target = _find_index_for_firstq_target(
+        two_level_array, num_qubits)
+    first_target_qubit_array = two_level_array[ind_for_0_target]
+    first_target_qubit_array_index = first_target_qubit_array[0]
+
+    # create a list of all two-level arrays without first or last qubit as
+    # targets
+    index_list = list(range(len(two_level_array)))
+    index_list.remove(0)
+    index_list.remove(ind_for_0_target)
```
purva-thakre, line 304: For the other two-level arrays - SWAP gates were used to map to either last qubit as target or first qubit as target (whichever has least number of gates) and then SWAP back to the original two-level array.

src/qutip_qip/decompose/_utility.py
```diff
@@ -29,3 +32,469 @@ def check_gate(gate, num_qubits):
...
+                all_together['targets ='] = target
+            backward_step_iteration_dictionary[
+                int((num_steps+1)/2)+i] = all_together
+
+        step_iteration_dictionary.append(
+            copy.deepcopy(backward_step_iteration_dictionary))
+        return(step_iteration_dictionary)
+
+
+def _paulix_for_control_0(step_iteration_dictionary):
```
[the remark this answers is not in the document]
purva-thakre: Decided to re-define this function after submitting final evaluation.

src/qutip_qip/decompose/decompose_general_qubit_gate.py
purva-thakre: Re-defining the function

src/qutip_qip/decompose/decompose_general_qubit_gate.py
purva-thakre: Re-defining the function and deleting two separate functions for first qubit and last qubit

src/qutip_qip/decompose/_utility.py
```diff
@@ -29,3 +32,469 @@ def check_gate(gate, num_qubits):
...
+        # right to left - with current code, the order is mixed.
+        gray_code = _gray_code_sequence(num_qubits)[
+            state_2_gray_code_index:state_1_gray_code_index+1]
+    gray_code_forward = gray_code
+    gray_code_backward = gray_code_forward[0:len(gray_code_forward)-1][::-1]
+
+    if num_steps_gray_code == 1:
+        return(gray_code_forward, num_steps_gray_code)
+    else:
+        return(gray_code_forward, gray_code_backward, num_steps_gray_code)
```
purva-thakre, line 158: The issue with creating dictionaries began here. The output was made dependent on if the gray code maps directly from one state to another or there's a couple of gates needed before implementing a two-level gate decomposition.

The new plan is to separately return the ctrl-target info for two-level gate itself and a list mapping to the two-level gate. In the case of there just being 1 step in the gray code, it would be easier to just check if list of gates is `None` or not.

src/qutip_qip/decompose/_utility.py
```diff
@@ -29,3 +32,469 @@ def check_gate(gate, num_qubits):
...
+                    gray_code_with_ctrl_1 = backward_gray_code[key]
+                    back_full_gate_i_info = [
+                        left_pauli, gray_code_with_ctrl_1, right_pauli]
+                    back_full_gate_dict[key] = back_full_gate_i_info
+                else:
+                    back_full_gate_dict[key] = backward_gray_code[key]
+
+        # calculated lengths
+        len_forward_gate_info = len(full_gate_dict)
+        len_backward_gate_info = len(back_full_gate_dict)
```
purva-thakre, line 491: Same as above comment - if there's more than one step then the list of gates can be reversed for the backward gray code gates. There's no need to iterate over forward gray code info to find the last gate object - a new function will output it separately.

src/qutip_qip/decompose/decompose_general_qubit_gate.py
```diff
@@ -0,0 +1,932 @@
...
+            reversed_list = full_toffoli_list[::-1]
+            full_gate_list.append(reversed_list[i])
+        return(full_gate_list)
+
+
+def _total_num_qubits_with_ancilla(num_qubits):
+    """Finds total number of qubits when ancilla qubits are added to create a
+    qubit circuit.
+    """
+    if num_qubits > 3:
```
purva-thakre, line 342: This needs to be changed to number of ctrl qubits - if n is the total number of qubits then total number of possible ctrl qubits are n-1.

src/qutip_qip/decompose/decompose_general_qubit_gate.py
```diff
@@ -0,0 +1,932 @@
...
+        gray_code_target = gray_code[0]['targets =']
+        if gray_code_target[0] == 0:
+            return(ind)
+
+
+def _full_CNOT_to_Toffoli(num_qubits):
+    """Decomposes a multi-controlled CNOT with last qubit as target into a
+    circuit composed of toffoli and ancilla qubits.
+    """
+    if num_qubits >= 3:
```
purva-thakre, line 286: if n = 3 : skip adding ancilla qubit

src/qutip_qip/decompose/decompose_general_qubit_gate.py
```diff
@@ -0,0 +1,932 @@
...
+        return(total_num_qubits)
+    else:
+        return(num_qubits)
+
+
+def _two_level_CNOT_to_Toffoli(num_qubits):
+    """When a 2-level unitary with last target is supposed to be decomposed,
+    4 multi-controlled CNOT will act on num_qubits-1 qubits.
+    """
+    return(_full_CNOT_to_Toffoli(num_qubits-1))
```
purva-thakre, line 355: add condition for n = 3, the current output is for when n > 3

src/qutip_qip/decompose/decompose_general_qubit_gate.py
```diff
@@ -0,0 +1,932 @@
...
+    for i in range(6):
+        if len(full_dict[i]) == 2:
+            cnot_ind.append(i)
+    forw_ind_list = {}
+    backw_ind_list = {}
+    full_ind_list = []
+    for i in cnot_ind:
+        gray_code_info = full_dict[i]
+        forw = gray_code_info[0]
+        backw = gray_code_info[1]
```
purva-thakre, line 594: This will be changed after forward gray code is separated from the main two-level gate. Then backward gray code is just reversed order of forward gray code.

src/qutip_qip/decompose/decompose_general_qubit_gate.py
```diff
@@ -0,0 +1,932 @@
...
+        ind_list.append(ind_at_i)
+    return(ind_list)
+
+
+def _shift_last_qubit_ancilla(two_level_arrays_index_list, num_qubits):
+    """in the gray code gate information, if a gate is either controlled on
+    last qubit or last qubit is a target,
+    shift the last qubit by number of ancilla qubits added.
+    """
+    if num_qubits > 3:
```
purva-thakre, line 779: add condition for n = 3 : return input

src/qutip_qip/decompose/decompose_general_qubit_gate.py
```diff
@@ -0,0 +1,932 @@
...
+    sub_lists = [
+            A1_gate_list,
+            A2_gate_list,
+            toffoli_CNOT_decomposition, CNOT1, toffoli_CNOT_decomposition,
+            B1_gate_list,  B2_gate_list, toffoli_CNOT_decomposition, CNOT1,
+            toffoli_CNOT_decomposition, C_gate_list, first_gate_list[-1]]
+
+    full_gate_list = list(flatten(sub_lists))
+
+    return(full_gate_list, total_qubits_for_circ)
```
purva-thakre, line 910: output of this function is not correct since ancilla qubits are not added correctly.

BoxiLi:
@purva-thakre Just for your interest. Here is a small example for flattening a nested list/dictionary:
``` python
def flatten(input, result):
    if not isinstance(input, (dict, list, tuple)):
        result.append(input)
        return
    for i in range(len(input)):
        flatten(input[i], result)

input = {0: [{0: "gate1", 1: "gate2"}, ["gate3"]], 1: "gate4"}
result = []
flatten(input, result)
print(result)
```

It basically uses recursion to reduce the problem into smaller ones and eventually solve the simplest cases. In this case, the smallest case is when the input is just one gate. It can also be understood as a deep-first search.

We don't really want to include this in the code, but it will probably help you test your result when you unify the output of the different results. I would suggest that you always have some (incomplete) tests running when you change the output format. In case changes break the correctness of the current output.

purva-thakre:
Closing this for now because there's [a new PR](https://github.com/qutip/qutip-qip/pull/106) to start with (hopefully) a better design of nested lists.

PR #91, opened 2021-08-22
Add qubit state generation, truncation and expansion.

BoxiLi:
**Changes**
- Add truncation and expansion with qubit state
  - Truncate multi-level state into a qubit state discarding the non-computational basis.
  - Expanding a qubit state multi-level state
- Support a more general way of definition:
  - use a list or a string for zero/one/plus/minus state
  - use a list of integers for defining arbitrary disentangled quantum states (up to a global phase).

**Question**
Should the name be `qubit_states` or `qubit_state`?


BoxiLi:
@purva-thakre If you have time, would you like to take a review on this PR? Maybe the `qubit_states` function is interesting to you when you write tests etc further for the decomposition. You could try the functions a bit and ask for explanations/changes on whatever is not clear or you think is wrong. No need to push it if you are busy :)

purva-thakre:
@BoxiLi Sure, I could take a look (might take me a couple of days).

src/qutip_qip/qubits.py
```diff
@@ -31,36 +31,229 @@
...
+            raise TypeError(f"Invalid input {s}.")
+    return qutip.tensor(states_list)
+
+
+def _find_reduced_indices(dims):
+    """Find the forresponding indices for a given qu"""
+    if len(dims) == 1:
+        return np.array([0, 1], dtype=np.intp)
+    else:
+        rest_reduced_indices = _find_reduced_indices(dims[1:])
```
purva-thakre, line 129: @BoxiLi I might be unfamiliar with how you defined the function here. 

How can you use a function being defined in its own definition ?

src/qutip_qip/qubits.py
```diff
@@ -31,36 +31,229 @@
...
+    [[0.70710678+0.j        ]
+     [0.        +0.70710678j]]
+    """
+    states_map = {
+        0: qutip.basis(2, 0),
+        1: qutip.basis(2, 1),
+        "0": qutip.basis(2, 0),
+        "1": qutip.basis(2, 1),
+        "+": (qutip.basis(2, 0) + qutip.basis(2, 1)).unit(),
+        "-": (qutip.basis(2, 0) - qutip.basis(2, 1)).unit(),
```
purva-thakre, line 107: Should there be a `Warning` or an `Exception` for the scenario when `+` and `-` are used without string quotations ? 

I think there will be a `SyntaxError` raised in either case. But the exception could change the input to string if this particular error is raised. 

Not sure if this is a good idea for built-in functions. I could give it a try if it's fine.

src/qutip_qip/qubits.py
```diff
@@ -31,36 +31,229 @@
...
+                s * qutip.basis(2, 1) +
+                np.sqrt(1 - abs(s)**2) * qutip.basis(2, 0)
+            )
+        else:
+            raise TypeError(f"Invalid input {s}.")
+    return qutip.tensor(states_list)
+
+
+def _find_reduced_indices(dims):
+    """Find the forresponding indices for a given qu"""
```
BoxiLi, line 125:
```suggestion
    """Find the corresponding indices for a given qu"""
```

src/qutip_qip/qubits.py
BoxiLi: Good question, this is called "recursion". A function calls itself indeed but every time the problem gets simpler. And there is always a "leaf" that stops this (the first `if` branch), otherwise, it will run forever.  In this code, every time I call the function, the dims becomes shorter. and eventually, it becomes length 1 and the first `if` branch terminates the program.

BoxiLi: The best example is always the Fibonacci numbers in Python that you can find everywhere.

``` python
def fibonacci(n):
    elif n == 0:
        return 0
    elif n == 1 or n == 2:
        return 1
    else:
        return fibonacci(n-1) + fibonacci(n-2)
```

src/qutip_qip/qubits.py
BoxiLi:
```suggestion
    """Find all the qubit indices for a given dimension, e.g. ``[3, 4]``"""
```

src/qutip_qip/qubits.py
BoxiLi: I think that will never happen, a `SyntaxError` will stop Python from running the code when someone tries to create the list like that. `+` and `-` are build-in functions in Python indeed as you said. Because of that, ``[+, -]`` can never be defined.

src/qutip_qip/qubits.py
purva-thakre: That's pretty interesting ! I knew about recursion of variables but this seems more efficient ? Here, you could skip defining the function variables for recursion, define a loop for them etc. So, is there a name for this method or it is another possible form of recursion ?

It's interesting that you can use `elif` without an `if` as well. I always thought the order was `if --> elif --> else` even in function definition. But since your function will first try to run the code in `else` loop, `elif` portions will be run after `return fibonacci(n-1) + fibonacci(n-2)`.

BoxiLi: > It's interesting that you can use elif without an if as well. 
Ohhhh nonono, that's not true. that was just a bug. You should always start with an `if`.

purva-thakre: > Ohhhh nonono, that's not true. that was just a bug

Ah. Ok. I haven't tried the code yet but my comment was looking through the code itself.

BoxiLi: > I knew about recursion of variables but this seems more efficient? Here, you could skip defining the function variables for recursion, define a loop for them etc. So, is there a name for this method or it is another possible form of recursion ?

I'm not sure if I understand correctly what you mean by *recursion of variables*. But in my use case here, you can indeed replace the recursion with a `for` loop. Because you know that it will take `len(dims)` iterations. However, that is not always true, like e.g. recursive calculation of the [Collatz conjecture](https://en.wikipedia.org/wiki/Collatz_conjecture).

Here, I choose the recursion instead of a `for` loop, because if I wrote a loop, I would have to start from the last item in `dims` (the last qubit). The nice thing about recursion is that you can understand the line
```rest_reduced_indices = _find_reduced_indices(dims[1:])```
as 
>Assuming that I know the result of `_find_reduced_indices(dims[1:])`, I only need to deal with the first indices. Then I can do it in the following way..."

The recursion helps you "realize" that assumption, which makes the logic cleaner.

PR #94, opened 2021-09-19
Support frame transformation and dressed states

BoxiLi:
- Allow a transformation matrix to be given, which will be applied to the Hamiltonian and collapse operators when generating QobjEvo for qutip solvers.
- Provide an option so that the dressed states w.r.t. the drift Hamiltonian will be calculated and applied as a transformation matrix.

### DIFF
--- a/src/qutip_qip/device/processor.py
+++ b/src/qutip_qip/device/processor.py
@@ -36,6 +36,7 @@ from copy import deepcopy
 
 import numpy as np
 from scipy.interpolate import CubicSpline
+from numpy.linalg import eigh
 
 from qutip import (Qobj, QobjEvo, identity, tensor, mesolve, mcsolve)
 from ..operations import expand_operator, globalphase
@@ -97,9 +98,6 @@ class Processor(object):
     num_qubits: int
         The number of component systems.
 
-    pulses : list of :class:`.Pulse`
-        A list of control pulses of this device
-
     t1: float or list
         Characterize the decoherence of amplitude damping of
         each qubit.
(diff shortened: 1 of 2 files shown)


BoxiLi:
The failing test seems to be a Coveralls connection issue. Retry later.

hodgestar (commented):
This looks like a cool new feature. I think we can simplify the complexity of it greatly though by:

- getting rid of `self.use_dressed_states`, `self._transform`, `get_transform` and `set_transform`
- keep `compute_dressed_states`
- add `transform=None` option to `get_noisy_pulse`

So that now one can do:
```python
dressed_states = processor.compute_dressed_states()
pulses = processor.get_noisy_pulses(transform=dressed_states)
```

The `Processor` class now doesn't have to manage any new state, it's simpler and clearer to the user of the class what is happening, and the user has more control (e.g. they could pass in their own transform if they wanted).

Thoughts?

BoxiLi:
That will indeed save us from the `get` and `set` methods and `self._transformation`. But this means that we also need to provide `transformation` as a parameter for other methods that use the dynamics, such as `run_state`, which run the circuit simulation. The transformation is applied when calling `get_qobjevo` (not `get_noisy_pulse` because `Pulse` does not support math operations, and QobjEvo does.). The more widely used public APIs like `run_state` call `get_qobjevo`.

The goal is to provide
- an easy API to use dressed states as logical qubit basis: `use_dressed_states`.
- a more general API that allows one to set custom transformation: `set_transformation`.

In my mind, the use cases of the dressed states are mostly in the resonator-qubit model. One provides a drift Hamiltonian where the qubit and resonator are weakly coupled. If one wants to use the dressed states, one only needs to set `self.use_dressed_states` to `True` (or this could be a default setting for some predefined model).

Seems to me that a simple switch `self.use_dressed_states` is more convenient, especially when dressed states are used as a default basis in some predefined Processor models (in the future)?

PR #95, opened 2021-09-22
Apply the Black code style to the source code

BoxiLi:
There are not many large-scale changes going on right now, except for the qutip5 updates, which I will sort out myself. So it seems a good time to do this? I applied black directly. Didn't check each file separately, but it looks good to me in general. Any caveat on doing this? @hodgestar


BoxiLi:
There is a check on Code Climatethat inconsistent with Black: ``Whitespace before ':'``. It can be turned off if desired.

hodgestar (approved):
Looks great to me -- thank you for doing this.

Let's just turn off the CodeClimate check that conflicts.

hodgestar:
As a separate PR we should run black as part of our GitHub actions so that CI fails if black has not been applied.

BoxiLi:
- Add the lint check to action
- Disable the incompatible check in code climate E203
- Add the max-length 79 configuration to `pyproject.toml`

I ignored code climate issues with "Avoid deeply nested control flow statements. ", which is unrelated to linting issues in this PR.

PR #96, opened 2021-10-02
Add gate class 

BoxiLi:
The first step of refactoring the gate representation. It defines a gate class for each gate function as a subclass of `Gate`, this simplifies several things:
- Using the gate class will allow one to identify the gate type through `isinstance` and `issubclass`, instead of by the name attribute.
- Get rid of several look-up tables in several different places.
- Allow a more natural way of defining custom gates. One only needs to define one's own gate class with a `get_compact_operator` method.
- (To be implemented in a different PR) Make the circuit representation much cleaner, in particular gate decomposition. Each gate decomposition should be implemented inside the gate class, instead of in `circuit.py`

Docstrings are also added to each gate class with examples.

See also https://github.com/qutip/qutip-qip/discussions/15 for more details.

There is a few codeclimate issues regarding Line too long and Similar blocks. I don't plan to fix them because the similar gate class are by design and the long lines are just docstrings with code example output. They are better like this.


tests/test_circuit.py
```diff
@@ -179,55 +177,6 @@ def test_add_gate(self):
         assert qc.gates[6].targets == [5]
         assert qc.gates[5].arg_value == 1.570796
 
-        # Test Exceptions  # Global phase is not included
```
BoxiLi, line 182: These tests are removed because they are too detailed and only checking `ValueError` without the error message also does not correctly check if the exception raised is the correct one. In addition, in the new gate class structure, the input control is moved to each class.

Issue #97, opened 2021-10-05
Show all API-doc on same page and TOC


nathanshammah:
Right now Submodules are separated from Subpackages. Also the TOC in does not show it all together. 
https://qutip-qip.readthedocs.io/en/latest/_apidoc/qutip_qip.html

PR #99, opened 2021-10-10
Use autosummary to generate a summary of API docs

BoxiLi:
Regenerate the API doc with a short summary of all public modules, classes and functions at the top of each page.

The HTML documentation can be downloaded and reviewed in the GitHub action check below.

Fix https://github.com/qutip/qutip-qip/issues/97

image


hodgestar (approved):
This looks great.

Issue #98, opened 2021-10-05
Pulse class: consider changing description or code refactoring


nathanshammah:
It has been brought up that the `Pulse` class can be confusing for the user, as it brings together control pulses, the relative Hamiltonian, and noise added to the control pulses. This issue is opened to discuss this point and see what action to take. 

 There is also some conceptual overlap between `qutip-qip.Pulse` and `qutip.QobjEvo` (and `qutip.Qobj` for the time-constant part).

Right now the `Pulse` class is also discussed in https://qutip-qip.readthedocs.io/en/latest/qip-processor.html#modelling-quantum-hardware-with-processor. 

The actions to be taken can be discussed here, for example:
- A. Code refactoring. 
- B. Remove pulse from the public APIs
- C. Reframing the presentation of `Pulse` in the documentation (and white paper)

BoxiLi:
My preferred solution would be:
- Add a `Model` class that represents the Hamiltonian model in `Processor` (see below).
- Bring the introduction of `Pulse` a bit down as an advanced use, both in the paper and in the tutorial. Phrase it as a data class rather than focusing too much on the concepts. It is different from `QobjEvo` because it separates noise from the ideal dynamics. And it needs to handle concatenation of coefficients, which is not a problem of `QobjEvo`.

The biggest problem I see is that the list of `Pulse` is partially initialized at the initialization of a `Processor` (including only the Hamiltonians, no `coeff`), which is indeed a less-optimal design. So I suggest to add a `Model` class that contains only Hamiltonian information. The class takes the hardware parameters to initialize
```
class Model():
    def __init__(params):
        ...

model = SpinChainModel({"h":0.01, ...})
```
and has a `__call__' that returns the corresponding Hamiltonian.
The Hamiltonian Qobj can be obtained by
```
model("sx0")
```
or whatever identifier that matches the implementation of the class, i.e.
```
model("sx", 0)
```

Advantages are:
- Hamiltonian models are now generated separately, there is not need for a partially initialized `Pulse`.
- There is no need to list all possible Hamiltonians ahead, which may take lots of memory for all-connected systems like ion trap or atoms.

@nathanshammah @quantshah Any opinions?

PR #104, opened 2021-10-22
Separate the physical model into a Model class

BoxiLi:
Previously, the Hamiltonian model in `Processor` is saved as partially initialized pulses. This PR separate the Hamiltonian model from pulse and the `Processor` class. It can now be defined independently and includes only the information of the Hamiltonian. The pulses will only be generated after the circuit is loaded.

The main changes are in `processor.py`.  In particular in the initialization. The others are either docs change or internal updates to fit this new structure. A bunch of getter functions are added to support the old way of building the Hamiltonians using the `Processor`. The data is, however, always saved in `Model`.

In principle `ModelProcessor` can be merged into `Processor`, but it is best to wait for another PR.

There are a few FIXME that are used to remind me to revisit some details later in a different PR. They are unrelated to the Hamiltonian model.

@nathanshammah @quantshah 

Issue #98

**TODO**
- [x] Add more tests
- [ ] Check compatibility with existing notebooks.

**Breaking changes**
- The pulse will not be generated before loading the circuit. So inquiring the Hamiltonian operator through pulse will only return an empty list. One should now use e.g. `Processor.get_control("sx0")` for this.
- `Processor.get_operators_labels()` is renamed as `get_latex_str` for clearity.

### DIFF
--- a/doc/source/qip-processor.rst
+++ b/doc/source/qip-processor.rst
@@ -24,8 +24,8 @@ Like a real quantum device, the processor is determined by a list of Hamiltonian
 
     processor = Processor(2)
     processor.add_control(sigmaz(), cyclic_permutation=True)  # sigmaz for all qubits
-    processor.pulses[0].coeffs = np.array([[1.0, 1.5, 2.0], [1.8, 1.3, 0.8]])
-    processor.pulses[0].tlist = np.array([0.1, 0.2, 0.4, 0.5])
+    processor.set_coeffs = np.array([[1.0, 1.5, 2.0], [1.8, 1.3, 0.8]])
+    processor.set_tlist = np.array([0.1, 0.2, 0.4, 0.5])
 
 It defines a :math:`\sigma_z` operator on both qubits and a pulse that acts on the first qubit.
 An equivalent approach is using the :meth:`.Processor.add_pulse` method.
@@ -159,19 +159,19 @@ To let it find the optimal pulses, we need to give the parameters for :func:`~qu
 
 
       # Same parameter for all the gates
-      qc = QubitCircuit(N=1)
+      qc = QubitCircuit(1)
       qc.add_gate("SNOT", 0)
 
       num_tslots = 10
       evo_time = 10
-      processor = OptPulseProcessor(N=1, drift=sigmaz())
+      processor = OptPulseProcessor(1, drift=sigmaz())
       processor.add_control(sigmax())
       # num_tslots and evo_time are two keyword arguments
       tlist, coeffs = processor.load_circuit(
       qc, num_tslots=num_tslots, evo_time=evo_time)
 
       # Different parameters for different gates
-      qc = QubitCircuit(N=2)
+      qc = QubitCircuit(2)
       qc.add_gate("SNOT", 0)
       qc.add_gate("SWAP", targets=[0, 1])
       qc.add_gate('CNOT', controls=1, targets=[0])
@@ -212,8 +212,7 @@ It is called implicitly when calling the method
     qc.add_gate("X", targets=1)
 
     processor = LinearSpinChain(2)
-    compiler = SpinChainCompiler(
-        2, params=processor.params, pulse_dict=processor.pulse_dict)
+    compiler = SpinChainCompiler(2, params=processor.params)
     resolved_qc = qc.resolve_gates(["RX", "RZ", "ISWAP"])
     tlists, coeffs = compiler.compile(resolved_qc)
     print(tlists)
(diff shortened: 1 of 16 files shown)


src/qutip_qip/device/cavityqed.py
```diff
@@ -255,3 +128,152 @@ def load_circuit(self, qc, schedule_mode="ASAP", compiler=None):
...
+        num_qubits = self.num_qubits
+        return (
+            ["sx" + str(i) for i in range(num_qubits)]
+            + ["sz" + str(i) for i in range(num_qubits)]
+            + ["g" + str(i) for i in range(num_qubits)]
+        )
+
+    def _set_up_controls(self):
+        """
+        Generate the Hamiltonians for the spinchain model and save them in the
```
quantshah, line 204: Should be "generate the Hamiltonians for the cavity-qed model"

doc/source/qip-processor.rst
```diff
@@ -159,19 +159,19 @@ To let it find the optimal pulses, we need to give the parameters for :func:`~qu
 
 
       # Same parameter for all the gates
-      qc = QubitCircuit(N=1)
+      qc = QubitCircuit(1)
```
nathanshammah, line 162:
```suggestion
      qc = QubitCircuit(1)
```
just a question, why did you delete "N=" here and elsewhere?

src/qutip_qip/device/cavityqed.py
```diff
@@ -255,3 +128,152 @@ def load_circuit(self, qc, schedule_mode="ASAP", compiler=None):
         )
         self.global_phase = compiler.global_phase
         return tlist, coeff
+
+
+class CavityQEDModel(Model):
+    """
+    The physical model for a dispersive cavity-QED system.
```
nathanshammah, line 135:
```suggestion
    The physical model for a dispersive cavity-QED system.
```
shall we change "system" with "processor"? Is there any other detail to add?

src/qutip_qip/device/cavityqed.py
```diff
@@ -255,3 +128,152 @@ def load_circuit(self, qc, schedule_mode="ASAP", compiler=None):
...
+
+    Parameters
+    ----------
+    num_qubits : int
+        The number of component systems.
```
nathanshammah, line 140:
```suggestion
        The number of component systems.
```
Sounds a bit strange. "The number of qubits"? or "The number of qubits composing the system"?

src/qutip_qip/device/cavityqed.py
```diff
@@ -255,3 +128,152 @@ def load_circuit(self, qc, schedule_mode="ASAP", compiler=None):
...
+    num_levels : int, optional
+        The truncation level of the resonator.
```
nathanshammah, line 142:
```suggestion
        The truncation level of the Hilbert space for the resonator.
```

src/qutip_qip/device/cavityqed.py
nathanshammah: Let's add brief details about the model Hamiltonian (should be the Tavis-Cummings Hamiltonian).

src/qutip_qip/device/circuitqed.py
```diff
@@ -27,97 +31,145 @@ class SCQubits(ModelProcessor):
...
+        self._default_compiler = SCQubitsCompiler
+        self.pulse_mode = "continuous"
+
+    def topology_map(self, qc):
+        return to_chain_structure(qc)
+
+
+class SCQubitsModel(Model):
+    """
+    The physical model for superconducting qubits with fixed frequency.
```
nathanshammah, line 64:
```suggestion
    The physical model for superconducting-qubit model processor with fixed frequency.
```

nathanshammah: We should add briefly a bit more details about the Hamiltonian. I propose we write it using the math directive in latex.

src/qutip_qip/device/processor.py
```diff
@@ -1020,3 +1198,120 @@ def _pulse_interpolate(pulse, tlist):
...
+        label_list : list of hashable objects
+            A list of hashable objects each corresponds to
+            an available control Hamiltonian.
+        """
+        return list(self._controls.keys())
+
+    def get_noise(self) -> List[Noise]:
+        """
+        Get a list of :obj:`.Noise` objects.
+        Single qubit relaxation (T1, T2) are not included here.
```
nathanshammah, line 1286: Why are the local T1 and T2 not included here?

nathanshammah (changes requested):
I left some comments and suggestions. 

Actually the code climate issues seem reasonable and could be actioned upon: besides blank spaces, there is a chunk of code involving pulses that maybe could be refactored in a private function.

Great to see these additions and the thorough testing.

BoxiLi:
@nathanshammah 
>there is a chunk of code involving pulses that maybe could be refactored in a private function.

Which part do you mean?

doc/source/qip-processor.rst
BoxiLi: I'm trying to discourage the use of capital `N` as the number of qubits, and eventually replace it with `num_qubits`, to be consistent.

But maybe this should not be included here in this PR, I agree. I can revert it if desired.

src/qutip_qip/device/processor.py
BoxiLi: Yeah there is a trade-off. They are not included here because they are defined just as two numbers and will be handled in the processor. I can add them directly here but then whoever implements their own Model will need to do it themselves.

On the other hand, separating them is indeed a bit strange. Now looking at them again, I think it might be better to include also T1 T2 here. Whoever implements a custom model will need to do their own homework, which should not be that difficult if they are already at this stage.

BoxiLi:
@nathanshammah 
>there is a chunk of code involving pulses that maybe could be refactored in a private function.

Which part do you mean?

nathanshammah:
Is this ready for review? The part of code I refer to is the one picked by code climate, 
```
  if isinstance(coeffs, dict):
            pulse_dict = self.get_pulse_dict()
            for pulse_name, value in coeffs.items():
                if pulse_name in pulse_dict:
                    self.pulses[pulse_dict[pulse_name]].coeff = value
```
anyways, this is fixed now.

BoxiLi:
> Is this ready for review?

Yes.

I think it is better to leave a more thorough doc including math equations in a separate PR and get this merged first. We can also update them after the resubmission.

nathanshammah (approved):
LGTM, thanks @BoxiLi.

PR #106, opened 2021-11-09
New Design For General Decomposition Function

purva-thakre:
This PR will complete unfinished work from GSoC 2021. Linking closed PR's just in case - ([gen_method](https://github.com/qutip/qutip-qip/pull/90), [gray_code](https://github.com/qutip/qutip-qip/pull/86), [two_qubit_test](https://github.com/qutip/qutip-qip/pull/82))

Because of the issue of nested dictionary etc, a new function will be added for every list of gates. 

For example, if all control values have to be controlled on 1 then this gate information will be in a separate list, if SWAP gates have to be used to make a gate controlled on last qubit then this gate information will be kept separate and so on. More details will be added as the functions and respective tests are added.

### DIFF
--- a/src/qutip_qip/decompose/decompose_general_qubit_gate.py
+++ b/src/qutip_qip/decompose/decompose_general_qubit_gate.py
@@ -1,0 +1,94 @@
+import numpy as np
+import cmath
+from qutip import Qobj
+from qutip_qip.decompose._utility import (
+    check_gate,
+)
+
+
+def _decompose_to_two_level_arrays(input_gate, num_qubits, expand=True):
+    """Decompose a general qubit gate to two-level arrays.
+
+    Parameters
+    -----------
+    input_gate : :class:`qutip.Qobj`
+        The gate matrix to be decomposed.
+    num_qubits : int
+        Number of qubits being acted upon by the input_gate
+    expand : True
+        Default parameter to return the output as full two-level Qobj. If
+    `expand = False` then the function returns a tuple of index information
+    and a 2 x 2 Qobj for each gate. The Qobj are returned in reversed order.
+    """
+    check_gate(input_gate, num_qubits)
+    input_array = input_gate.full()
+
+    # Calculate the two level numpy arrays
+    array_list = []
+    index_list = []
+    for i in range(2**num_qubits):
+        for j in range(i+1, 2**num_qubits):
+            new_index = [i, j]
+            index_list.append(new_index)
+
+    for i in range(len(index_list)-1):
+        index_1, index_2 = index_list[i]
+
+        # Values of single qubit U forming the two level unitary
+        a = input_array[index_1][index_1]
+        a_star = np.conj(a)
+        b = input_array[index_2][index_1]
+        b_star = np.conj(b)
+        norm_constant = cmath.sqrt(
+            np.absolute(a*a_star)+np.absolute(b*b_star))
+
+        # Create identity array and then replace with above values for
+        # index_1 and index_2
+        U_two_level = np.identity(2**num_qubits, dtype=complex)
+        U_two_level[index_1][index_1] = a_star/norm_constant
+        U_two_level[index_2][index_1] = b/norm_constant
+        U_two_level[index_1][index_2] = b_star/norm_constant
+        U_two_level[index_2][index_2] = -a/norm_constant
+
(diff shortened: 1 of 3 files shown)


nathanshammah:
Thanks @purva-thakre, great to see this coming over!

src/qutip_qip/decompose/decompose_general_qubit_gate.py
```diff
@@ -0,0 +1,168 @@
...
+    num_two_level_gates = len(two_level_output)
+
+    # create a reversed list of keys based on total number of two-level gates
+    # ranging from 1 to n where n is the total number of two-level arrays
+    gate_keys = list(range(1, num_two_level_gates+1))[::-1]
+    gate_info_dict = dict.fromkeys(gate_keys)
+    return(gate_info_dict)
+
+
+def _partial_gray_code(num_qubits, two_level_output):
```
purva-thakre, line 111: Add tests for this -

- [ ] Check length of the output matches to that of the two-level array
- [ ] Check for the gate keys being consistent
- [ ] Check if the output is expected - 2 and 3 qubits

src/qutip_qip/decompose/decompose_general_qubit_gate.py
```diff
@@ -0,0 +1,168 @@
...
+        else:
+            partial_gray_code = [ind1_pos_in_gray_code, ind2_pos_in_gray_code]
+
+        gate_key_dict[len(two_level_indices)-i] = gray_code_string[
+            partial_gray_code[0]:partial_gray_code[1]+1]
+
+    return(gate_key_dict)
+
+
+def _split_partial_gray_code(gate_key_dict):
```
purva-thakre, line 148: Add tests for this -

- [ ]  Check length of the output matches to that of the two-level array
- [ ]  Check if 2 dictionaries are always returned

- [ ] Check for the gate keys being consistent

 

- [ ] Check if the output is expected - 2 and 3 qubits : These will be easier to check with known dictionary values are 'None' or not

purva-thakre, line 148: Add tests for this -

- [ ]  Check length of the output matches to that of the two-level array
- [ ]  Check if 2 dictionaries are always returned

- [ ] Check for the gate keys being consistent

 

- [ ] Check if the output is expected - 2 and 3 qubits : These will be easier to check with known dictionary values are 'None' or not

tests/decomposition_functions/test_general_decomposition.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    input_gate = rand_unitary(2**num_qubits, dims=[[2] * num_qubits] * 2)
+    array_decompose = _decompose_to_two_level_arrays(
+        input_gate, num_qubits, expand=False)
+    empty_dict_output = _create_dict_for_two_level_arrays(array_decompose)
+    assert np.equal(len(empty_dict_output), len(array_decompose))
+
+
+@pytest.mark.parametrize("num_qubits", [2, 3, 4, 5])
+def test_empty_dict_of_two_level_arrays(num_qubits):
```
purva-thakre, line 50: rename this test - forgot to change while copy pasting previous test

src/qutip_qip/decompose/decompose_general_qubit_gate.py
purva-thakre: `_split_partial_gray_code` not giving expected output.

src/qutip_qip/decompose/decompose_general_qubit_gate.py
```diff
@@ -0,0 +1,178 @@
...
+            partial_gray_code = [ind2_pos_in_gray_code, ind1_pos_in_gray_code]
+        else:
+            partial_gray_code = [ind1_pos_in_gray_code, ind2_pos_in_gray_code]
+
+        gate_key_dict[len(two_level_indices) - i] = gray_code_string[
+            partial_gray_code[0] : partial_gray_code[1] + 1
+        ]
+
+    return gate_key_dict
+
```
purva-thakre, line 152: Add a function to make all control values consistent before the partial gray code gets split into n-bit toffoli and the gate needing decomposition. 

If the function is used before ` _split_partial_gray_code` then don't have to worry about searching for controls between the two split dictionaries.

PR #108, opened 2021-11-14
Make circuit latex code accessible in QubitCircuit

BoxiLi:
- `QubitCircuit.latex_code()` now returns the full latex code (before only the code body without header), which can be saved as a tex file and compiled externally.
- The algorithm will print the latex code if pdflatex failed. This simplifies the debugging process.

@purva-thakre This should also simplify your debugging I guess?


hodgestar (approved):
Looks good to me!

PR #115, opened 2021-12-13
Minor correction on the figure label and settings

BoxiLi:
Minor corrections on the labels and plotting settings. The manuscript is correct, I forgot to push this. Raise the PR just to make everything consistent.

### DIFF
--- a/doc/pulse-paper/qft.py
+++ b/doc/pulse-paper/qft.py
@@ -97,12 +97,13 @@ simulation_time = [0.007144343852996826,
  11.539598500728607,
  31.34040207862854]
 
-fig, ax = plt.subplots(figsize=(LINEWIDTH, LINEWIDTH/3*2), dpi=200)
-ax.plot(range(1, 11), compiling_time, "-s", markersize=4, label="solver")
-ax.plot(range(1, 11), simulation_time, "-D", markersize=4, label="compiling")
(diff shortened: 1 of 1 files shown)


BoxiLi:
Black fails on unrelated files? Cannot reproduce off-line.

BoxiLi:
@hodgestar Do you have a clue what happened in the falling test? It points to an unchanged file, but the lint test passed for all the PRs in the last few weeks. Also, the lint test is configured with `--diff` so I don't understand what happened. I cannot reproduce it offline.

hodgestar:
> @hodgestar Do you have a clue what happened in the falling test? It points to an unchanged file, but the lint test passed for all the PRs in the last few weeks. Also, the lint test is configured with `--diff` so I don't understand what happened. I cannot reproduce it offline.

The `--diff` just prints the changes that were made, and the test output seems to have done that?

I'm guessing that maybe a new version of black was released? If so, maybe you can reproduce locally by upgrading black?

I would perhaps also recommend that we just "pip install black" ourselves instead of using the psf/black action. That way we can pin black to a fixed version and upgrade it when we're ready, rather than always having to have the latest.

For now I would ignore the error and fix it in a separate PR.

hodgestar (approved):

PR #121, opened 2022-01-26
Rework on the doc of the pulse-level simulation

BoxiLi:
This is a thorough rework of the documentation of the pulse-level simulation. I adapted some texts from the paper to construct a new doc for `Processor`.

I also add a hyper reference to the qutip doc so that references like `qutip.Qobj` will also work.


src/qutip_qip/device/cavityqed.py
```diff
@@ -134,22 +134,59 @@ class CavityQEDModel(Model):
     """
     The physical model for a dispersive cavity-QED processor
     (:obj:`.DispersiveCavityQED`).
+    It is a qubits-resonator model that describes a system composed of
```
nathanshammah, line 137:
```suggestion
    It is a qubit-resonator model that describes a system composed of
```

src/qutip_qip/device/cavityqed.py
```diff
@@ -134,22 +134,59 @@ class CavityQEDModel(Model):
...
+        \\Delta^{\\rm{max}}_{j}(t) \\sigma^z_{j} +
+        J_{j}(t)
+        (a^\\dagger \\sigma^{-}_{j} + a \\sigma^{+}_{j}).
+
+    The effective qubit-qubit coupling is computed by the
+
+    .. math::
+
+        J = \\frac{g_j g_{j+1}}{2}(\\frac{1}{\\Delta_j} +
+        \\frac{1}{\\Delta_{j+1}}),
```
nathanshammah, line 167: Maybe J_j?

src/qutip_qip/device/circuitqed.py
```diff
@@ -63,6 +62,42 @@ class SCQubitsModel(Model):
     """
     The physical model for superconducting-qubit model processor
     (:obj:`.SCQubits`) with fixed frequency.
+    Each qubit is simulated by a multi-level Duffing model,
```
nathanshammah, line 65: Can we add a reference or link for the exact form to the Duffing model? By the way, always better to have somewhere the full Hamiltonian also with the cross-resonance.

nathanshammah (commented):
Looks great @BoxiLi, thanks. Added some comments.

src/qutip_qip/device/circuitqed.py
BoxiLi: Good point, I'll add a reference. Isn't the full Hamiltonians are [below in the docstring](https://github.com/BoxiLi/qutip-qip/blob/f35525fe390746632f750158d16600ce9ea04840/src/qutip_qip/device/circuitqed.py#L80-L100)? Or are you expecting more details?

Issue #118, opened 2022-01-17
Add Variational Quantum Algorithms Module


EnBr55:
QuTiP currently lacks any functionality directly related to VQAs. It would be useful to have a class-based implementation that allows for problem specification down to the details of parametrized/fixed Hamiltonians and the optimisation method. 

To outline a possible structure for this module, I have implemented something along these lines in [my fork of this repository](https://github.com/EnBr55/qutip-qip/blob/master/src/qutip_qip/vqa.py). This includes:

1. `Optimization_Result` class
  - A convenient interface for results of the VQA optimisation method.
2. `VQA_Block` class
  - A kind of pseudo-circuit-element. Holds one of either a Hamiltonian, a Unitary, or a pre-defined QuTiP gate (e.g. `qutip_qip.operations.x_gate`).
  - Takes a list of qubit targets
  - Can be either an "initial" element or not - this will determine if the block is repeated in the circuit layers.
  - In the case that a Hamiltonian or Unitary is specified, the action of the block is implemented by adding a new entry to the `QubitCircuit`'s `user_gates` dictionary.
3. `VQA` class
  - Initialised with `n_qubits`, `n_layers`
  - Can specify a cost function in terms of:
    - An observable
    - A state
    - A bitstring (single measurement outcome sampled from the circuit) 
  - `add_block(block: VQA_Block)` appends `VQA_Block` to the circuit and, if necessary, updates the circuit's `user_gates` dictionary to include the new gate.
  - `optimise_parameters` method performs optimisation on the free parameters of the circuit and returns an `Optimization_Result` instance.
  - Methods that serve as a wrapper to `QubitCircuit` to simulate the circuit

A usage example can be found at https://github.com/EnBr55/qutip-vqa-examples/blob/main/notebooks/QAOA.ipynb.

I'm currently working on a `Parameterized_Hamiltonian` model as a `VQA_Block` input that will allow for defining more general quantum control problems. 

Any discussion/suggestions as to these proposed models would be great!

nathanshammah:
This is great @EnBr55, looking forward to discussing more with @BoxiLi and other contributors. As a small comment, all names proposed look very "pythonic" :) and the tutorial looks great.

BoxiLi:
[...]

PR #123, opened 2022-02-09
Add Variational Quantum Algorithms Module

EnBr55:
Functionality as outlined in #118

- [X] Core functionality
- [X] Following style requirements
- [X] API documentation
- [x] Tests
  - Mostly done, still in progress
- [x] Manual documentation
  - In progress

This PR introduces a new module, `qutip_qip.vqa`, for defining and simulating Variational Quantum Algorithms. For examples for how this module could be used, please see https://github.com/EnBr55/qutip-vqa-examples/


doc/source/apidoc/qutip_qip.vqa.rst
```diff
@@ -0,0 +1,40 @@
...
+   
+   
+   .. rubric:: Classes
+
+   .. autosummary::
+   
+      Optimization_Result
+      Parameterized_Hamiltonian
+      VQA
+      VQA_Block
```
nathanshammah, line 21:
```suggestion
      VQABlock
```

EnBr55: Good catch, thanks.

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,852 @@
...
+        -------
+        bitstring: str
+            Formatted binary string with length corresponding to
+            dimension of input state
+
+        E.g. the state 1/sqrt(2) * (|0> + |1>)
+        would return 0 and 1 with equal probability.
+        """
+        n_qubits = int(np.log2(state.shape[0]))
+        outcome_indices = list(range(2 ** n_qubits))
```
BoxiLi, line 204: In the new black styling released a few days ago, it is recommended to omit the space around `**`. That's why the lint test is failing. You will get it right with an updated version of black.

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,852 @@
...
+            a bitstring "|010>" would produce a set with the first and last
+            elements of S, and one with the second element of S.
+        top_ten: bool, optional
+            Only plot the ten highest-probability states.
+        """
+        n_qubits = int(np.log2(state.shape[0]))
+        probs = [abs(i.item()) ** 2 for i in state]
+        bitstrings = [
+            "|" + format(i, f"0{n_qubits}b") + ">"
+            for i in range(2 ** n_qubits)
```
BoxiLi, line 821: Same lint issue.

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,852 @@
+"""Variational Quantum Algorithms generation and optimization"""
+import types
+import random
+import numpy as np
+from qutip import basis, tensor, Qobj, qeye, expect
+from qutip.qip.circuit import QubitCircuit
+from scipy.optimize import minimize
+from scipy.linalg import expm_frechet
+from qutip.qip.operations.gates import gate_sequence_product
+import matplotlib.pyplot as plt
```
BoxiLi, line 10: `matplotlib` is only an optional package required, so the test actions do not include that automatically. You can wrap this with a try-catch syntax to avoid failing tests or move this importation into the plotting function.

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,852 @@
...
+
+    def get_top_bitstring(self):
+        """
+        Return the bitstring associated with the highest probability
+        measurement outcome
+
+        Returns
+        -------
+        bitstring: str
+            bitstring in the form |x_0x_1...x_n> where each x_i is '0' or '1'
```
BoxiLi, line 737: Better warp `|x_0x_1...x_n>` with the math mode, ``:math:`|x_0x_1...x_n>` ``. Underscores have special meanings in rst files, so the building doc test will complain.

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,852 @@
...
+
+        Parameters
+        ----------
+        S: list of float, optional
+            The problem instance, in the case of a combinatorial
+            optimization problem.
+        label_sets: bool, optional
+            Replace bitstring labels with sets referring to the inferred
+            output of the combinatorial optimization problem. For example
+            a bitstring "|010>" would produce a set with the first and last
```
BoxiLi, line 762: Same with math ``:math:`|010>` ``

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,852 @@
...
+        Parameters
+        ----------
+        S: list of float, optional
+            Problem instance
+        min_cost: str, optional
+            The minimum cost found by optimization
+        label_sets: bool, optional
+            Replace bitstring labels with sets referring to the inferred
+            output of the combinatorial optimization problem. For example
+            a bitstring "|010>" would produce a set with the first and last
```
BoxiLi, line 812: Math mode as well

BoxiLi:
Bingo! You passed most of the tests. The coverage report is still complaining but I understand more tests are on the way. You can click the button `Details` to see which lines are not covered.

You can also test the plotting function by simply running the function without checking any result. This does not check if the plot generated is correct but at least verifies that the workflow throws no errors. Like in 
https://github.com/qutip/qutip-qip/blob/860856cd2dbb24affe8795dd177ee14b32442ad0/tests/test_processor.py#L136-L153

EnBr55:
Hey @BoxiLi where would be the most appropriate place for the doc page for this module? I was thinking the best match currently might be under https://qutip-qip.readthedocs.io/en/latest/qip-simulator.html but wasn't sure if it should have its own page instead? Thanks :)

BoxiLi:
Yeah you can start a new page for it. Maybe also give a brief intro for VQA.

BoxiLi:
It is a bit annoying that the test needs to be approved every time when a new push is made. Not sure if there is a way to just allow any future tests to run for this PR. Sorry for all the delay, @EnBr55 Ben, please feel free to drop me an email next time you add something and want to run the test. Sometimes I miss the GitHub email and am unaware of any new push.

We can also try to get this PR merged first and add new patches in a different one so that any future PR of yours won't need approval.

The doc test is failing because the online doc generator does not have latex installed. You can render a png locally and add it as a figure, just like the other part of doc.

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,841 @@
...
+            Whether to compute the jacobian or not. If computed, it will be
+            passed to the optimizer chosen by ``method``, regardless of if
+            the method is gradient-based or not.
+            Note that derivatives of unitaries generated by
+            ``ParameterizedHamiltonian`` are calculated with the
+            Frechet derivative of the exponential function,
+            using ``scipy.linalg.expm_frechet``.
+        do_nothing: bool, optional
+            For debugging and benchmarking purposes, perform no operation
+            during optimization, and simply retain the initial free parameters.
```
BoxiLi, line 278: This is a bit confusing, I understand that for debugging you might just want to evaluate the parameters. However, having this in the optimization method seems contradictory to the name of the method.
Is this necessary? Why not just call the evaluation function in debugging?

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,841 @@
...
+                "Initial conditions were neither a list of values"
+                " nor a string specifying initialization."
+            )
+
+        # Return early if do_nothing flag was passed
+        if do_nothing:
+            return self.evaluate_parameters(angles)
+
+        # Function for computing jacobian
+        jac = self.compute_jac if use_jac else None
```
BoxiLi, line 319: This is only used if `layer_by_layer==False`

doc/source/qip-vqa.rst
```diff
@@ -0,0 +1,119 @@
+.. _qip_vqa:
+
+******************************
+Variational Quantum Algorithms
+******************************
+
```
BoxiLi, line 6: You can add your name and/or your email to get some credit if you want. ;)

doc/source/qip-vqa.rst
```diff
@@ -0,0 +1,119 @@
...
+  result.plot()
+
+**Output**:
+
+.. testoutput::
+  :options: +NORMALIZE_WHITESPACE
+
+  Parameter found: 3.14
+
+.. image:: /figures/vqa_circuit_optimization_output.png
```
BoxiLi, line 117: You don't have to import his figure. Using the plot derivative you can directly show the matplotlib figure you generated from the code (https://qutip-qip.readthedocs.io/en/latest/contribution-docs.html#id4). Sorry if my previous comment misleads you. I meant only the circuit figure.

BoxiLi (commented):
@EnBr55 This is just part of the review. The code looks very neat. Mostly just naming plus a few of my questions, to clarify things. You don't have to accept all of them and feel free to discuss :)

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,841 @@
...
+        if not block.name:
+            block.name = "U" + str(len(self.blocks))
+        if block.name in list(map(lambda b: b.name, self.blocks)):
+            raise ValueError("Duplicate Block name in self.blocks")
+        self.blocks.append(block)
+        self.user_gates[block.name] = lambda angles=None: block.get_unitary(
+            angles
+        )
+
+    def get_free_parameters(self):
```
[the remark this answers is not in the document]
EnBr55: Agreed, thanks!

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,841 @@
...
+        """
+        Returns
+        -------
+        initial_state: Qobj
+            Initial circuit state
+        """
+        initial_state = basis(2, 0)
+        for _ in range(self.n_qubits - 1):
+            initial_state = tensor(initial_state, basis(2, 0))
+        return initial_state
```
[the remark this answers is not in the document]
EnBr55: You're right - I think it's sensible to assume the initial state is 0, and the user can prepare a different initial state by adding `VQABlock`s with the "initial" flag set to true. Good call on the `basis([2] * n, [0] * n)`, that's much more convenient! 

I think it might be a bit more friendly to the user if they can call `VQA.get_initial_state()` rather than having to use `basis([2] * VQA.n_qubits, [0] * VQA.n_qubits)`, even if the function is that simple? Happy to hear your thoughts, though.

src/qutip_qip/vqa.py
EnBr55: Oops, I think I meant to remove this! Has been removed now.

doc/source/qip-vqa.rst
EnBr55: Got it, thanks! I've switched to using the plot directive for this section now.

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,841 @@
...
+                raise ValueError("self.cost_func not specified")
+            else:
+                return self.cost_func(final_state)
+        elif self.cost_method == "OBSERVABLE":
+            if self.cost_observable is None:
+                raise ValueError("self.cost_observable not specified")
+            else:
+                return expect(self.cost_observable, final_state)
+        else:
+            raise ValueError(f"Unrecognised cost method: {self.cost_method}")
```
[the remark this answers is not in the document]
EnBr55: Thanks, I've modified the all the messaging that specifies `self` with something more clear.

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,841 @@
...
+        Parameters
+        ----------
+        initial: str or list of floats, optional
+            Initialization method for the free parameters.
+            If a list of floats of dimensions (n,) for n free parameters
+            in the circuit is given, then these are taken to be the initial
+            conditions for the optimizer. Otherwise if a string is given:
+            1. "random" will randomize initial free parameters between 0 and 1,
+            2. "ones" will set each initial free parameter to a value of 1.
+            Defaults to "random"
```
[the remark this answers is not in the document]
EnBr55: I've updated this to use bullet points now. Is it correct to space things with newlines like this within the docstring?

```python
"""
[...]
conditions for the optimizer. Otherwise if a string is given:

* (Default) "random" will randomize initial free parameters
  between 0 and 1.

* "ones" will set each initial free parameter to a value of 1.
"""
```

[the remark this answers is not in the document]
BoxiLi: Seems to work. 
image

You can also download it in our GitHub action 
"Build HTML documentation" from the generated artifacts.(https://github.com/apps/github-actions)@github-actions](https://github.com/qutip/qutip-qip/actions/runs/2201435768)

src/qutip_qip/vqa.py
EnBr55: Good catch, thanks.

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,842 @@
...
+            arg = -1j * self.operator.get_hamiltonian(angles)
+            direction = -1j * self.operator.p_terms[term_index]
+        else:
+            if len(angles) != 1:
+                raise ValueError(
+                    "Expected a single angle for non-"
+                    "ParameterizedHamiltonian instance."
+                )
+            arg = -1j * self.operator * angles[0]
+            direction = -1j * self.operator
```
BoxiLi, line 697: This part has some overlap with the `get_unitary_derivative` method. The method `get_unitary_frechet_derivative` is only used if the operator is `ParameterizedHamiltonian`.

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,842 @@
...
+        bounds: sequence or `scipy.optimize.Bounds`, optional
+            Bounds to be passed to the optimizer. Either
+
+            1. Instance of `scipy.optimize.Bounds`
+            2. Sequence of ``(min, max)`` tuples corresponding to each
+               free parameter.
+        constraints: list of `Constraint`
+            See `scipy.optimize.minimize` documentation.
+        """
+
```
BoxiLi, line 299: Add to the docstring
```python
"""
Returns
-------
result: :obj:`.OptimizationResult`
    The optimized angles and final state.
"""
```

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,842 @@
...
+            Frechet derivative of the exponential function,
+            using ``scipy.linalg.expm_frechet``.
+        layer_by_layer: bool, optional
+            Grow the circuit from a single layer, to ``VQA.n_layers``, at each
+            step holding the parameters found for previous layers fixed.
+        bounds: sequence or `scipy.optimize.Bounds`, optional
+            Bounds to be passed to the optimizer. Either
+
+            1. Instance of `scipy.optimize.Bounds`
+            2. Sequence of ``(min, max)`` tuples corresponding to each
```
BoxiLi, line 294: Using Sphinx enumerating/listing?

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,842 @@
...
+    parameterized_terms: list of Qobj
+        Hamiltonian terms which each require a unique parameter
+    constant_term: Qobj
+        Hamiltonian term which does not require parameters.
+    """
+
+    def __init__(self, parameterized_terms=[], constant_term=None):
+        self.p_terms = parameterized_terms
+        self.c_term = constant_term
+        self.N = len(parameterized_terms)
```
BoxiLi, line 506: Is this the same as `num_parameters`? Better be consistent with the name.

doc/source/qip-vqa.rst
```diff
@@ -0,0 +1,124 @@
...
+
+    from qutip_qip.vqa import VQA
+
+    VQA_circuit = VQA(
+                n_qubits=1,
+                n_layers=1,
+                cost_method="OBSERVABLE",
+            )
+
+After constructing this instance, we are ready begin adding elements to our parameterized circuit. Circuit elements in this module are represented by :class:`.VQABlock` instances. Fundamentally, the role of this class is to generate an operator for the circuit. To do this, it keeps track of its free parameters, and gives information to the :class:`.VQA` instance on how to update them. The operator itself can be generated by a user-defined function call, a parameterized Hamiltonian to exponentiate, a pre-computed unitary operator, or a string referring to a gate that has already been defined (either by the user already, or a gate native to QuTiP).
```
BoxiLi, line 46:
```suggestion
After constructing this instance, we are ready to begin adding elements to our parameterized circuit. Circuit elements in this module are represented by :class:`.VQABlock` instances. Fundamentally, the role of this class is to generate an operator for the circuit. To do this, it keeps track of its free parameters, and gives information to the :class:`.VQA` instance on how to update them. The operator itself can be generated by a user-defined function call, a parameterized Hamiltonian to exponentiate, a pre-computed unitary operator, or a string referring to a gate that has already been defined (either by the user already, or a gate native to QuTiP).
```

doc/source/qip-vqa.rst
```diff
@@ -0,0 +1,124 @@
...
+
+In the absence of specification, a VQA block will assume any :class:`~.Qobj` it is given to be a Hamiltonian, :math:`H`, and will generate a unitary with free parameter, :math:`\theta`, as :math:`U(\theta) = e^{-i \theta H}`. For example,
```
BoxiLi, line 48: If I understand correctly
```suggestion
In the absence of specification, a VQA block takes :class:`~.Qobj` a Hamiltonian :math:`H`, and will generate a unitary with free parameter, :math:`\theta`, as :math:`U(\theta) = e^{-i \theta H}`. For example, 
```

It would be nice if examples for other types of VQA blocks can also be added. Can leave it for another PR.

doc/source/qip-vqa.rst
```diff
@@ -0,0 +1,124 @@
...
+  Adding a Pauli X operator as a block to the circuit, the operation of the entire circuit becomes: :math:`e^{-i t X /2}`.
+
+
+  >>> circ.add_block(VQABlock(sigmax() / 2))
+
+  We can now try to find a minimum in our cost function using the SciPy in-built L-BFGS-B (L-BFGS with box constraints) method. We specify the bounds so that our parameter is :math:`0 \leq t \leq 4`.
+
+  >>> result = circ.optimize_parameters(method="L-BFGS-B", use_jac=True, bounds=[[0, 4]])
+
+  Accessing `result.res.x`, we have the array of parameters found during optimization. In our case, we only had one free parameter, so we examine the first element of this array.
```
BoxiLi, line 113:
```suggestion
  Accessing ``result.res.x``, we have the array of parameters found during optimization. In our case, we only had one free parameter, so we examine the first element of this array.
```

BoxiLi (changes requested):
@EnBr55 Here is the second half of the review. Sorry for the delay.

- There are two png figures with the same content. One is duplication I guess (doc/circ.png)?

- If it is not too late, can we rename the public API
`n_qubits` -> `num_qubits`
`n_layers` -> `num_layers`
`n_paramters` -> `num_parameters`
to be consistent with the rest of the naming convention in qutip-qip? For the intermediate result not accessible to the user it does not matter.

- It is fine to merge as is. If you want to develop further on this module. I would recommend to define different types of VQA blocks as subclasses. In that way you don't have to use lots of if-else branches to check the input. You only need to specify this when defining the VQA blocks, e.g. `UnitaryVQABLock`, `HamiltonianVQABlock` etc..

src/qutip_qip/vqa.py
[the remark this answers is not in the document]
BoxiLi: Sure yeah it is a bit easier.

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,842 @@
...
+            self.c_term if self.c_term else 0
+        )
+        return H_tot
+
+
+class VQABlock:
+    """
+    Component of a ``VQA``. Can return a unitary, and take
+    derivatives of its own unitary. Forms a ``Gate`` in the
+    ``QuantumCircuit`` generated by the ``VQA``.
```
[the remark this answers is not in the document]
BoxiLi: I see that you already know this and used it in the rest doc. It also works for the API docs. `:obj:` here is the same as `:class:`.

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,842 @@
...
+                self.n_parameters = 1
+        elif isinstance(operator, str):
+            self.is_native_gate = True
+            if targets is None:
+                raise ValueError("Targets must be specified for native gates")
+        elif isinstance(operator, ParameterizedHamiltonian):
+            self.n_parameters = operator.N
+        elif isinstance(operator, types.FunctionType):
+            if not isinstance(operator(1), Qobj):
+                raise ValueError("Provided function does not return a Qobj")
```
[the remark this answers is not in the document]
EnBr55: Good call, I agree. I've added a check on the output when the function is called, just to avoid the possibility of weird behaviour if things don't entirely crash on non-`qobj` inputs somewhere.

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,842 @@
...
+        if isinstance(self.operator, types.FunctionType):
+            return self.operator(angles[0])
+        if self.is_unitary:
+            if isinstance(self.operator, types.FunctionType):
+                return self.operator(angles[0])
+            else:
+                return self.operator
+        else:
+            if angles is None:
+                raise ValueError("No parameter given")
```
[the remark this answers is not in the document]
EnBr55: You're right, sounds good to me. Changed and commented to make it a bit more clear.

src/qutip_qip/vqa.py
```diff
@@ -0,0 +1,842 @@
...
+                raise ValueError(
+                    "Expected a single angle for non-"
+                    "ParameterizedHamiltonian instance."
+                )
+            arg = -1j * self.operator * angles[0]
+            direction = -1j * self.operator
+        return Qobj(
+            expm_frechet(arg, direction, compute_expm=False),
+            dims=direction.dims,
+        )
```
BoxiLi, line 711: This seems the only issue that remained. I think `get_unitary_frechet_derivative` can be removed and the single part that is used
```python
            arg = -1j * self.operator.get_hamiltonian(angles)
            direction = -1j * self.operator.p_terms[term_index]
```
can be moved directly to `get_unitary_derivative`?

Once this is cleaned. I believe we are ready to merge!

EnBr55: Was just writing a comment here haha. I'll post below anyway

That's true. I was originally thinking maybe the user would want to define a custom function to generate the Jacobian during optimization, which could depend on this method. Technically the method would have output the same as `get_unitary_derivative` if the operator is a `Qobj` Hamiltonian, but it's probably sensible to require that this method should only be called for `VQABlock`s that use a `ParameterizedHamiltonian`, so we could merge the two methods.

BoxiLi: > I was originally thinking maybe the user would want to define a custom function to generate the Jacobian during optimization, which could depend on this method.

I see. Yes, custom derivatives could be useful. They can even overwrite the full `get_unitary_derivative` I guess? One can surely further explore that framework in future PRs.

BoxiLi (approved):

BoxiLi:
Congrats @EnBr55 ! Feel free to open further PRs for more documentation examples as well as further development!

@nathanshammah Pinging Nathan to share the moment.

EnBr55:
Thanks so much for all the help and advice @BoxiLi @nathanshammah !! :)

PR #125, opened 2022-02-11
Add exporter from QubitCircuit to QIR.

cgranade:
This draft/work-in-progress PR ports the changes from https://github.com/qutip/qutip/pull/1781 to the qutip-qip package as per feedback from @BoxiLi, @nathanshammah, and @hodgestar.

This draft will need to be updated once https://github.com/qir-alliance/pyqir/issues/34 closes, and still needs additional unit testing, but I wanted to open the PR for initial feedback on the design and interface. Thank you!

Example notebook: https://gist.github.com/cgranade/c26fc3fbbe39ac1c455cc0974356c6be


src/qutip_qip/qir.py
```diff
@@ -0,0 +1,180 @@
...
+            op_with_less_controls.classical_controls = op_with_less_controls.classical_controls[1:]
+            branch_body = {value: (lambda: append_operation(module, builder, op_with_less_controls))}
+            builder.if_result(module.results[result], **branch_body)
+            return
+
+        if op.controls:
+            if op.name not in ("CNOT", "CZ") or len(op.controls) != 1:
+                raise NotImplementedError(
+                    "Arbitrary controlled quantum operations are not yet supported."
+                )
```
BoxiLi, line 101: It seems that CZ is missing from the list below?

src/qutip_qip/qir.py
```diff
@@ -0,0 +1,180 @@
...
+        # the reserved operation names in the QIR base profile's quantum
+        # instruction set (QIS).
+        if isinstance(op, Gate):
+            # TODO: Validate indices.
+            append_operation(module, builder, op)
+
+        elif isinstance(op, Measurement):
+            # TODO: Validate indices.
+            if op.name == "Z":
+                builder.m(module.qubits[op.targets[0]], module.results[op.classical_store])
```
BoxiLi, line 149: The measurement in qutip always measure in the Z basis. Regardless of the `name`. So I guess this is not necessary.

src/qutip_qip/qir.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+    # Define as an inner function to make it easier to call from conditional
+    # branches.
+    def append_operation(module: pqg.SimpleModule, builder: pqg.BasicQisBuilder, op: Gate):
+        if op.classical_controls:
+            result = op.classical_controls[0]
+            value = "zero" if op.classical_control_value == 0 else "one"
+            # Pull off the first control and recurse.
+            op_with_less_controls = Gate(**op.__dict__)
+            op_with_less_controls.classical_controls = op_with_less_controls.classical_controls[1:]
```
BoxiLi, line 92: The same should also be applied for `classical_control_value`.

BoxiLi (changes requested):
@cgranade Thanks! I had a discussion with @hodgestar last week. Sorry that it was left here for a while. 
 
I left some comments. It looks good in general and it is great to support the conversion to QIR.
- Some tests are still needed.
- Running `black` should clean the Lint automatically.

Another question: how hard would it be to support the conversion from `QIR` to `qutip-qip`? From our last discussion, I remember there were some functions in QIR that are not supported. But it should be possible to just raise a `NotImplementedError` for those missing functions?

cgranade:
Apologies for missing the notification on your review!

> @cgranade Thanks! I had a discussion with @hodgestar last week. Sorry that it was left here for a while.

No worries, and thanks for your comments!

> 
> I left some comments. It looks good in general and it is great to support the conversion to QIR.
> 
> * Some tests are still needed.

Agreed, was in draft for feedback on the design; wanted to have agreement there before writing tests. That has now been added.

> * Running `black` should clean the Lint automatically.

Done.

> Another question: how hard would it be to support the conversion from `QIR` to `qutip-qip`? From our last discussion, I remember there were some functions in QIR that are not supported. But it should be possible to just raise a `NotImplementedError` for those missing functions?

It's not just missing functions — QIR allows for representing quantum programs in general, and is not limited only to nonadaptive quantum programs (circuits). For example, random walk phase estimation, repeat-until-success loops for circuit synthesis, or magic state distillation (https://www.youtube.com/watch?v=_0QJicz4ZR4&list=PLh7C25oO7PW0FH2JlQYtEPRJNs6t_2N50) are represented in QIR by a graph of different blocks of instructions connected by flow control instructions.

From that perspective, QIR programs in general need not be representable by a circuit at all. I strongly agree it would be good to have some functionality to that end, perhaps similar to @crazy4pi314's [qwop package](https://github.com/crazy4pi314/qwop), but wanted to leave that to a future PR, given that I think there's a bit more design work to do there at the moment. Would love to hop on a call at some point to discuss further!

src/qutip_qip/qir.py
cgranade: Ah, got it; removed that condition, then. Thanks!

src/qutip_qip/qir.py
cgranade: Done, thank you!

src/qutip_qip/qir.py
cgranade: Good catch, thanks!

src/qutip_qip/qir.py
```diff
@@ -0,0 +1,180 @@
...
+    Given a circuit acting on qubits, generates a representation of that
+    circuit using Quantum Intermediate Representation (QIR).
+
+    :param circuit: The circuit to be translated to QIR.
+    :param format: The QIR serialization to be used. If `"text"`, returns a
+        plain-text representation using LLVM IR. If `"bitcode"`, returns a
+        dense binary representation ideal for use with other compilation tools.
+        If `"module"`, returns a PyQIR module object that can be manipulated
+        further before generating QIR.
+    :param module_name: The name of the module to be emitted.
```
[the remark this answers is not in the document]
cgranade: Done, thanks!

BoxiLi (changes requested):
@cgranade Thanks! Looks good!
- Should the other modes (TEXT and BITCODE) also be covered in the test by a simple example?
- Adding the installation to the CI workflow `test.yml` should make the test executed by the GitHub CI and increases the coverage.

For the documentation it would be great to 
- add a short example in https://qutip-qip.readthedocs.io/en/stable/qip-simulator.html#import-and-export-quantum-circuits.
- and a new section in the API doc similar to `qutip_qip.qasm`.

src/qutip_qip/qir.py
```diff
@@ -0,0 +1,231 @@
...
+        f = NamedTemporaryFile(suffix=".bc", delete=False)
+        try:
+            f.write(bitcode)
+        finally:
+            f.close()
+        module = pqp.QirModule(f.name)
+        try:
+            os.unlink(f.name)
+        except:
+            pass
```
BoxiLi, line 226: A small question, why is this `unlink` needed here?

cgranade: Ah, this ensures that the temporary file written in the previous step is deleted when we're done with it. Normally, that's handled when the `TemporaryFile` object gets garbage collected, but since we need to write to it after it's been closed, that obligates us to manually unlink it when we're done.

cgranade:
Thanks for the additional feedback!

> * Adding the installation to the CI workflow `test.yml` should make the test executed by the GitHub CI and increases the coverage.

Sounds good, I'll go on and add that, then.  

> For the documentation it would be great to
> 
> * add a short example in https://qutip-qip.readthedocs.io/en/stable/qip-simulator.html#import-and-export-quantum-circuits.
> * and a new section in the API doc similar to `qutip_qip.qasm`.

Sounds good, I'll get some docs written up shortly, then. Thank you!

cgranade:
@BoxiLi: Apologies for the delay over IEEE Quantum Week! Anyway, I've updated for both above, please let me know if there's anything else you'd like for me to add. Thanks!

doc/source/qip-simulator.rst
```diff
@@ -280,9 +280,9 @@ We are left with a mixed state.
 Import and export quantum circuits
 ==================================
 
-QuTiP supports importation and exportation of quantum circuit in the `OpenQASM 2 W-state <https://github.com/Qiskit/openqasm/tree/OpenQASM2.x>`_ format
-throught the function :func:`.read_qasm` and :func:`.save_qasm`.
-We demonstrate this using the w-state generation circuit.
+QuTiP supports importing and exporting quantum circuits in the `OpenQASM 2.0 <https://github.com/Qiskit/openqasm/tree/OpenQASM2.x>`_ format, as well as exporting circuits to `Quantum Intermediate Representation <https://www.qir-alliance.org/>`.
```
BoxiLi, line 283:
```suggestion
QuTiP supports importing and exporting quantum circuits in the `OpenQASM 2.0 <https://github.com/Qiskit/openqasm/tree/OpenQASM2.x>`_ format, as well as exporting circuits to `Quantum Intermediate Representation <https://www.qir-alliance.org/>`_.
```

doc/source/qip-simulator.rst
```diff
@@ -326,3 +326,80 @@ One can save it in a ``.qasm`` file and import it using the following code:
...
+  circuit.add_measurement("Z", targets=[msg], classical_store=0)
+  circuit.add_measurement("Z", targets=[here], classical_store=1)
+  circuit.add_gate("X", targets=[there], classical_controls=[0])
+  circuit.add_gate("Z", targets=[there], classical_controls=[1])
+
+  print(circuit_to_qir(circuit, "text"))
+
+.. testoutput::
+  ; ModuleID = 'qutip_circuit'
+  source_filename = "qutip_circuit"
```
BoxiLi, line 352:
```suggestion
.. testoutput::

  ; ModuleID = 'qutip_circuit'
  source_filename = "qutip_circuit"
```

doc/source/qip-simulator.rst
```diff
@@ -326,3 +326,80 @@ One can save it in a ``.qasm`` file and import it using the following code:
 
   from qutip_qip.qasm import read_qasm
   qc = read_qasm("source/w-state.qasm")
+
+QuTiP circuits can also be exported to QIR:
+
+.. testcode::
+  from qutip_qip.circuit import QubitCircuit
+  from qutip_qip.qir import circuit_to_qir
```
BoxiLi, line 334:
```suggestion
.. testcode::

  from qutip_qip.circuit import QubitCircuit
  from qutip_qip.qir import circuit_to_qir
```

BoxiLi (commented):
All looks good! The failing doctest is likely just because of the missing empty line. I added the those changes in the comments.

cgranade:
@BoxiLi: Ah, that makes a lot of sense, I'd forgotten that the first line in RST block-role syntax was special in that way! Thanks for catching that, for your review, and for all your feedback! Is there anything else I can do from my end to get this merged in? Thank you again!

doc/source/qip-simulator.rst
```diff
@@ -326,3 +326,82 @@ One can save it in a ``.qasm`` file and import it using the following code:
...
+
+  declare void @__quantum__qis__mz__body(%Qubit*, %Result*)
+
+  declare i1 @__quantum__qis__read_result__body(%Result*)
+
+  declare void @__quantum__qis__x__body(%Qubit*)
+
+  declare void @__quantum__qis__z__body(%Qubit*)
+
+  attributes #0 = { "EntryPoint" "requiredQubits"="3" "requiredResults"="2" }
```
BoxiLi, line 407: I ran the failing doc example and saw that the out put is slightly different from the given ones
E.g.
```
%equal = call i1 @__quantum__qis__read_result__body(%Result* null)
br i1 %equal, label %then, label %else
```
becomes
```    
%0 = call i1 @__quantum__qis__read_result__body(%Result* null)
br i1 %0, label %then, label %else
```
I guess this is just changes in the updated version of QIR?

If that is the case. I would suggest that we don't explicitly check the printed strings. Because QIR may change it in the future version. But, instead, just checking that the code runs fine without erros.
```suggestion
.. doctest::

  >>> from qutip_qip.circuit import QubitCircuit
  >>> from qutip_qip.qir import circuit_to_qir

  >>> circuit = QubitCircuit(3, num_cbits=2)
  >>> msg, here, there = range(3)
  >>> circuit.add_gate("RZ", targets=[msg], arg_value=0.123)
  >>> circuit.add_gate("SNOT", targets=[here])
  >>> circuit.add_gate("CNOT", targets=[there], controls=[here])
  >>> circuit.add_gate("CNOT", targets=[here], controls=[msg])
  >>> circuit.add_gate("SNOT", targets=[msg])
  >>> circuit.add_measurement("Z", targets=[msg], classical_store=0)
  >>> circuit.add_measurement("Z", targets=[here], classical_store=1)
  >>> circuit.add_gate("X", targets=[there], classical_controls=[0])
  >>> circuit.add_gate("Z", targets=[there], classical_controls=[1])

  >>> print(circuit_to_qir(circuit, "text"))  # doctest: +SKIP
```

If you agree with it and it passes the test I think we can merge :)

BoxiLi (approved):

BoxiLi:
Thanks a lot for adding this! @cgranade

cgranade:
Thank you so much as well! Really excited to have this out there! 💕

PariaNaghavi:
To provide an example for generating qir from qutip, I have been referring to the [documentations](https://qutip-qip.readthedocs.io/en/latest/qip-simulator.html?highlight=qir#import-and-export-quantum-circuits); However, I encountered some issues. Specifically, it appears that the current version of qutip-qip (0.3.0) does not provide access to the `.qir` attribute nor the `circuit_to_qir` function. When attempting to import the [qir.py](https://github.com/qutip/qutip-qip/blob/master/src/qutip_qip/qir.py) locally, it is missing dependency on `pyqir.parser`, which is gone in the current version of pyqir (0.7.0). This may be due to version incompatibility between the two libraries with the recent updates. 
Next, I adjusted the code to use the `pyqir-generator` instead of the parser, but I reached the following error about SNOT gate “not implemented in the QIR base profile and may require a custom declaration".

**To replicate:** Please see [here](https://github.com/PariaNaghavi/qir-book/blob/main/qir-book/tutorials/qir-emission/qutip/qir-gen.ipynb)
**Current outcome:** An error due to missing `.qir` attribute and `circuit_to_qir` function in qutip-qip version and the missing dependency `pyqir.parser`
**Expected outcome:** to generate qir, as shown in [docs](https://qutip-qip.readthedocs.io/en/latest/qip-simulator.html?highlight=qir#import-and-export-quantum-circuits)
**Notes:** I would appreciate any suggestions or insights that might help in addressing these issues. Please let me know if there is any additional information or assistance that I can provide to fix this issue.
**Environment:**
QuTiP-QIP version: 0.0.3
pyqir version: 0.7.0
Python version: 3.9

hodgestar:
@PariaNaghavi Would you mind turning your comment above into an issue? This will make it easier to keep track of and more likely that anyone else experiencing a similar issue would find it.

PR #131, opened 2022-03-12
Incorrect spelling of Hadamard

bopardikarsoham:
There is a typo in [qip-basics.rst line 152](https://github.com/qutip/qutip-qip/edit/master/doc/source/qip-basics.rst#L152); it should be `Hadamard` instead of `Hardmard`

### DIFF
--- a/doc/source/Hadamard
+++ b/doc/source/Hadamard
@@ -1,0 +1,1 @@
+M	doc/source/qip-basics.rst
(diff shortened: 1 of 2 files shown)


doc/source/Hadamard
```diff
@@ -0,0 +1 @@
+M	doc/source/qip-basics.rst
```
BoxiLi, line 1: Hi @bopardikarsoham, thanks for the correction!

What is this file for?

bopardikarsoham:
@BoxiLi I have rectified my mistake in this PR #132

PR #136, opened 2022-04-05
Support qutip-v5

BoxiLi:
This is a replacement of #88.

- Add support of qutip=5.0 to `setup.cfg`
- New GitHub actions are added to test qutip-qip against the dev.major version of qutip
- Fix the importation errors for `qutip.Options`, use `qutip.SolverOptions` instead.
- Use `Qobj.full()` to transfer the matrix to NumPy arrays before using `assert_all_close`, since `Qobj.__array__` is removed.
- Tests are adapted for the new QobjEvo in qutip.
  - Replace the `QobjEvo.ops` and `QobjEvo.cte` as they are removed. Create separate test functions to compare `QobjEvo`s.
  - Remove the use of `QobjEvo.tlist`.
  - Slightly modify the pulse and noise module to correctly handle the continuous/discrete pulse in `QobjEvo`.
- Fix/Improve some other unstable/wrong tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,7 +46,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: 3.8
+        python-version: 3.6
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,31 +10,37 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.6, 3.9]
         os: [ubuntu-latest, windows-latest, macOS-latest]
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Python ${{ matrix.python-version }}
+    - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: ${{ matrix.python-version }}
+        python-version: 3.9
     - name: Install qutip-qip
       run: |
         python -m pip install --upgrade pip
         pip install -e .[tests]
-    - name: Test with pytest
+    - name: Test with pytest and generate coverage report
       run: |
-        pytest tests --strict-config --strict-markers --verbosity=1
+        pip install pytest-cov coveralls
+        pytest tests --strict-markers --cov=qutip_qip --cov-report=
+    - name: Upload to Coveralls
+      env:
+        GITHUB_TOKEN: ${{ secrets.github_token }}
+        COVERALLS_FLAG_NAME: ${{ matrix.qutip-version }}
+        COVERALLS_PARALLEL: true
+      run: coveralls --service=github
 
   test-qutip-support:
     # test for qutip master branch
     runs-on: ubuntu-latest
+
     strategy:
-      matrix:
-        # TODO: add dev.major and minimal supported qutip version v4.6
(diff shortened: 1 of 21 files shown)


BoxiLi:
@hodgestar @Ericgig I made a version of qutip-qip that is compatible with the current qutip `dev.major` branch. Would you like to have a quick look?

Not to review if the changes are correct for qutip-qip, I can handle that. But rather, if those changes are indeed what we want other users to make to update from version 4 to 5. This could also provide some examples that we can later write in the documentation when we release qutip5.

src/qutip_qip/compiler/gatecompiler.py
```diff
@@ -247,6 +249,18 @@ def _concatenate_pulses(
                 compiled_tlist[pulse_ind].append(execution_time)
                 compiled_coeffs[pulse_ind].append(coeffs)
 
+        final_time = np.max([tlist[-1] for tlist in compiled_tlist])
```
hodgestar, line 252: It looks like this is unrelated to QuTiP 5. I'm happy for it to stay in this PR potentially, but what does it add / do?

src/qutip_qip/device/cavityqed.py
```diff
@@ -147,7 +147,14 @@ def eliminate_auxillary_modes(self, U):
...
             + [identity(2) for n in range(self.num_qubits)]
         )
-        return psi_proj.dag() * U * psi_proj
+        result = psi_proj.dag() * U * psi_proj
+        # In qutip 5 multiplication of matrices
+        # with dims [[1, 2], [2, 2]] and [[2, 2], [1, 2]]
+        # will give a result of
+        # dims [[1, 2], [1, 2]] instead of [[2], [2]].
+        if result.dims[0][0] == 1:
+            result = result.ptrace(list(range(len(self.dims)))[1:])
```
hodgestar, line 156: This looks like it will bite other users. Is there some better way to support it? Maybe `psi_proj` dims need to be fixed before the multiplication somehow?

hodgestar (approved):

src/qutip_qip/compiler/gatecompiler.py
BoxiLi: This is actually related to the fact that evaluation out of the tlist will give non-zero values in qutip 5. Because of this, I need to determine the final time and add some zeros to the end.

.github/workflows/test.yml
```diff
@@ -10,31 +10,37 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.6, 3.9]
```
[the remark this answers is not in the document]
BoxiLi: Ops, will add this back.

src/qutip_qip/device/cavityqed.py
hodgestar: Since we'd added looking at this to the TODO list, let's leave it for now.

.github/workflows/test.yml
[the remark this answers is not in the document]
hodgestar: Once this is added back, I'm happy for the PR to be merged.

src/qutip_qip/device/cavityqed.py
BoxiLi: @hodgestar, @Ericgig, I was collecting some common error messages that I saw while updating to v5. And I notice this again. This issue still remains. It is still somewhere in a TODO list?

PR #137, opened 2022-04-11
Fix Circuit Plotting for non-reversed states

christian512:
The plotting of circuits had an error for setting `reversed=False`. This was already discussed in a PR for the QuTiP repository. 
See: https://github.com/qutip/qutip/pull/1847

### DIFF
--- a/src/qutip_qip/circuit.py
+++ b/src/qutip_qip/circuit.py
@@ -2011,7 +2011,8 @@ class QubitCircuit:
                     if n in measurement.targets:
                         col.append(r" \meter")
                     elif (n - self.N) == measurement.classical_store:
-                        store_tag = n - measurement.targets[0]
+                        sgn = 1 if self.reverse_states else -1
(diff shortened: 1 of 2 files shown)


BoxiLi:
Thank you very much @christian512 ! There is a small deviation here. In `qutip-qip`, `qc.latex_code()` gives the full executable LaTeX code rather than just the circuit part. The intention is to make the output directly usable. So one gets here
```
\documentclass{standalone}
\usepackage[braket]{qcircuit}
\renewcommand{\qswap}{*=<0em>{\times}}
\begin{document}
\Qcircuit @C=1cm @R=1cm {
&  &  \meter & \qw \\ 
&  &  \qw \cwx[-1]  & \qw \\ 
}
\end{document}
```

christian512:
@BoxiLi , thanks for the comment :)  I will change the two tests for this function.

BoxiLi:
Tests pass! Many thanks @christian512 !

Issue #51, opened 2021-06-01
Add `about()` function


quantshah:
Since we are branching out and making `qutip_qip` a separate library, it might be beneficial to have something like `qutip.about` for qip. The about() function could list out the QuTiP version that the code is run with and help in debugging in the future. 

Here is how the about function in QuTiP is written:

```
"""
Command line output of information on QuTiP and dependencies.
"""
__all__ = ['about']

import sys
import os
import platform
import numpy
import scipy
import inspect
from qutip.utilities import _blas_info
import qutip.settings
from qutip.hardware_info import hardware_info


def about():
    """
    About box for QuTiP. Gives version numbers for QuTiP, NumPy, SciPy, Cython,
    and MatPlotLib.
    """
    print("")
    print("QuTiP: Quantum Toolbox in Python")
    print("================================")
    print("Copyright (c) QuTiP team 2011 and later.")
    print(
        "Current admin team: Alexander Pitchford, "
        "Nathan Shammah, Shahnawaz Ahmed, Neill Lambert, Eric Giguère, "
        "Boxi Li, Jake Lishman and Simon Cross."
    )
    print(
        "Board members: Daniel Burgarth, Robert Johansson, Anton F. Kockum, "
        "Franco Nori and Will Zeng."
    )
    print("Original developers: R. J. Johansson & P. D. Nation.")
    print("Previous lead developers: Chris Granade & A. Grimsmo.")
    print("Currently developed through wide collaboration. "
          "See https://github.com/qutip for details.")
    print("")
    print("QuTiP Version:      %s" % qutip.__version__)
    print("Numpy Version:      %s" % numpy.__version__)
    print("Scipy Version:      %s" % scipy.__version__)
    try:
        import Cython
        cython_ver = Cython.__version__
    except ImportError:
        cython_ver = 'None'
    print("Cython Version:     %s" % cython_ver)
    try:
        import matplotlib
        matplotlib_ver = matplotlib.__version__
    except ImportError:
        matplotlib_ver = 'None'
    print("Matplotlib Version: %s" % matplotlib_ver)
    print("Python Version:     %d.%d.%d" % sys.version_info[0:3])
    print("Number of CPUs:     %s" % hardware_info()['cpus'])
    print("BLAS Info:          %s" % _blas_info())
    print("OPENMP Installed:   %s" % str(qutip.settings.has_openmp))
    print("INTEL MKL Ext:      %s" % str(qutip.settings.has_mkl))
[...]

PR #138, opened 2022-04-17
Added about() function

claretgrace0801:
For solving #51, have added an `about` function to `qutip_qip`. 
This pull request is a continuation of the PR: qutip/qutip#1870 and is dependent on it (this PR won't work without merging the mentioned `qutip` PR. That `qutip` PR does not affect any existing functionality of `qutip`)

**Changelog**
`qutip_qip.about()` calls `qutip.about("qutip_qip")`, which show info about `qutip_qip`

Refer to qutip/qutip#1870 for more details.


BoxiLi (commented):
Looks good. It would be nice if we can rename the function that returns titles and lines as a private function `_about`, and give it to the entry point. Then we can define a `about` that just call `qutip.about()`.

Issue #133, opened 2022-03-25
Allow setting dpi/density when exporting circuits to PNG.


cgranade:
Currently, the density for exporting circuits to PNG is hard-coded to 100dpi (https://github.com/qutip/qutip-qip/blob/master/src/qutip_qip/circuit_latex.py#L122). It'd be really helpful to be able to configure that and export higher-resolution circuit diagrams when appropriate. Thanks! ♥

srinathkp:
Hi @cgranade  / @BoxiLi , I'm interested in qutip project for GSoC'22. As this is marked as a good first issue, I'd like to work on this. I see that we use imagemagick to convert to png and that the density is hard-coded to 100 now. Where can we take the dpi/density argument from ?

BoxiLi:
Hi @srinathkp, great! Currently the plot is only used for plotting in the notebook
https://github.com/qutip/qutip-qip/blob/497fa2f999297e7df6ae5497cc78d6f09bf4cac6/src/qutip_qip/circuit.py#L2055-L2066 
Hence there is no argument taken.

You can consider adding an method like `QubitCircuit.draw(file_type="png", dpi=100)` that outputs a figure file.

srinathkp:
Hi @BoxiLi / @cgranade , the converter functions for _pdf, png_ and _svg_ are made with fixed configurations(hard-coded dpi value of 100) using `circuit_latex._make_converter(configuration)`. We can add an element to `_CONVERTER_CONFIGURATIONS` list with density value from the user but it will mean that a new converter has to be made every time user tries to export an image. A more dynamic approach would be having a new function that directly calls `ImageMagick` tool with arguments from the user ( with appropriate checks and validation ofc ).

Please let me know your thoughts.

BoxiLi:
Hi @srinathkp, yes I agree with your argument. Could you make a PR using that branch? It would be easier to review because the PR shows directly what has been changed.

srinathkp:
Hey @BoxiLi , Thanks for the feedback, Wanted to discuss here before I implement and test it. Will raise a PR when I'm done.

BoxiLi:
In the current way, the dpi is written fixed in the configuration of the converter
https://github.com/qutip/qutip-qip/blob/497fa2f999297e7df6ae5497cc78d6f09bf4cac6/src/qutip_qip/circuit_latex.py#L142-L157

[...]

PR #139, opened 2022-04-25
Feature/export png with dpi

srinathkp:
For #133 - Allow setting dpi/density when exporting circuits to PNG

New method `QubitCircuit.draw()` to export svg and png images. Density in dpi can be an argument from the user for conversion to png ( defaults to 100dpi).


srinathkp:
- [ ] Add test to py_test.py
- [ ] Remove ipynb file

src/qutip_qip/circuit_latex.py
```diff
@@ -240,7 +276,10 @@ def image_from_latex(code, file_type="png"):
                     raise ValueError(
                         "".join(["Unknown output format: '", file_type, "'."])
                     )
-                out = CONVERTERS[file_type](filename)
+                if file_type == "png" and dpi != 100:
+                    out = convert_to_img_with_dpi(filename, file_type, dpi)
+                else:
+                    out = CONVERTERS[file_type](filename)
```
BoxiLi, line 282: Is it possible to completely replace the old converter with this one? It is better to have as fewer duplications as possible. I feel like your `convert_to_img_with_dpi` can already almost replace the only converter. Maybe it can just read from the configurations and change the dpi setting?

srinathkp: I haven't tried `convert_to_img_with_dpi` for SVG image. It's functionally possible to replace `_make_converter` with the new implementation. The advantage of having `_make_converter` function is that we don't have to run `_find_system_command` for every conversion like we do with `convert_to_img_with_dpi` and the converter functions are ready to use with fixed arguments.

BoxiLi: It could be quite often (refreshing the same circuit?) but the generation of the plot through `pdflatex` definitely takes much more time than `_find_system_command`, which checks if the command exists in the current system.

Even if we want to avoid running `_find_system_command` at each run, we can still put your converter inside `_make_converter`, right? So that the generated converter can take dpi as input (and probably raise an error if the dpi is given when the type is SVG).

srinathkp:
Hi @BoxiLi , I've added the density argument to old converter and added a test that looks for exported png files(with different dpi) and SVG.

src/qutip_qip/circuit_latex.py
srinathkp: `_make_converter`_ takes density as argument now. Thanks

src/qutip_qip/circuit.py
```diff
@@ -2066,6 +2072,24 @@ def png(self):
         """
         return DisplayImage(self._raw_png(), embed=True)
 
+    def draw(
+        self,
+        file_type="png",
+        density_dpi=100,
```
BoxiLi, line 2078: Let's name the argument just `dpi`, matplotlib has a similar argument name.

src/qutip_qip/circuit.py
```diff
@@ -2056,6 +2057,11 @@ def latex_code(self):
     def _raw_png(self):
         return _latex.image_from_latex(self.latex_code(), "png")
 
+    def _raw_img(self, file_type="png", density_dpi=100):
+        return _latex.image_from_latex(
+            self.latex_code(), file_type, density_dpi
+        )
+
     if "png" in _latex.CONVERTERS:
         _repr_png_ = _raw_png
```
BoxiLi, line 2067: I guess we can now safely remove `_raw_png` and `_raw_svg` with your `_raw_img`.
```python
    def _raw_img(self, file_type="png", density_dpi=100):
        return _latex.image_from_latex(
            self.latex_code(), file_type, density_dpi
        )

    if "png" in _latex.CONVERTERS:
        _repr_png_ = _raw_img

    if "svg" in _latex.CONVERTERS:
        _repr_svg_ = partial(_raw_img, file_type="svg", density_dpi=None)
```

src/qutip_qip/circuit.py
```diff
@@ -2066,6 +2072,24 @@ def png(self):
         """
         return DisplayImage(self._raw_png(), embed=True)
 
+    def draw(
+        self,
+        file_type="png",
+        density_dpi=100,
+        file_name="exported_pic",
+        file_path="",
+    ):
```
BoxiLi, line 2081: Let's also add some doc string to this function explaining the arguments.

src/qutip_qip/circuit.py
```diff
@@ -2066,6 +2072,24 @@ def png(self):
...
+        if file_type == "svg":
+            mode = "w"
+            image_data = self._raw_svg()
+        else:
+            mode = "wb"
+            image_data = self._raw_img(file_type, density_dpi)
```
BoxiLi, line 2087: We should be able to remove this if else branch and just use `_raw_img`.

src/qutip_qip/circuit.py
```diff
@@ -2056,6 +2057,11 @@ def latex_code(self):
     def _raw_png(self):
         return _latex.image_from_latex(self.latex_code(), "png")
 
+    def _raw_img(self, file_type="png", density_dpi=100):
+        return _latex.image_from_latex(
+            self.latex_code(), file_type, density_dpi
+        )
```
BoxiLi, line 2063: If we also want to also use this for SVG, I think we should set the default dpi to `None` and
```python
    def _raw_img(self, file_type="png", density_dpi=None):
        if file_type == "png"
            density_dpi=100
        return _latex.image_from_latex(
            self.latex_code(), file_type, density_dpi
        )
```

BoxiLi (commented):
Looks good! Thank you!
I left some comments to clear the logic in the code. Don't worry about the coverage, the CI has no pdflatex installed.

src/qutip_qip/circuit.py
srinathkp: As we have the partial method derived for SVG with `dpi=None`, I think `_raw_img `can have the default value of 100 for dpi. However, we can set the default dpi to None in `draw` as users can call the method directly. On a side note, even if we don't set the dpi to None in `_raw_img` or `draw` for SVG, it's ignored during conversion using pdf2svg in `_make_converter` .

src/qutip_qip/circuit.py
srinathkp: We don't need if else branch to get `image_data` now but, we still need if else to assign write modes for SVG & PNG.

BoxiLi: Ah, yes that's right.

src/qutip_qip/circuit.py
BoxiLi: > However, we can set the default dpi to None in draw as users can call the method directly. 

Yes, this makes sense.

> On a side note, even if we don't set the dpi to None in _raw_img or draw for SVG, it's ignored during conversion using pdf2svg in _make_converter.

Agreed. It won't affect the result. It only makes the code a bit more clearer at the first glance.

BoxiLi (approved):
Tested it locally, LGTM!

BoxiLi:
@srinathkp Congrats!

srinathkp:
Thanks Boxi ! Many more to come hopefully 💥🤞

PR #140, opened 2022-05-01
Separate the png and svg tests for circuit plots

BoxiLi:
Separate the tests so that it skips if svg/png converters are not installed.

Use `in_temporary_directory` defined in `conftest.py` to replace the `cleanup`. Because if the test fails on the png part, the files were not cleaned.

### DIFF
--- a/tests/test_circuit.py
+++ b/tests/test_circuit.py
@@ -711,13 +711,13 @@ class TestQubitCircuit:
         qc.add_gate("CSIGN", controls=[0], targets=[1])
 
         if "png" in circuit_latex.CONVERTERS:
-            file_png_200 = "exported_pic_200.png"
-            file_png_400 = "exported_pic_400.png"
-            qc.draw("png", 200, file_png_200.split('.')[0], "")
-            qc.draw("png", 400.5, file_png_400.split('.')[0], "")
-            assert file_png_200 in os.listdir('.')
-            assert file_png_400 in os.listdir('.')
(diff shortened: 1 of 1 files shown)


BoxiLi:
@srinathkp, I made some small changes on the tests so that if one of png and svg is not supported, the tests won't fail. Does this make sense to you?

srinathkp:
Hi @BoxiLi , yeah it makes sense to have checks for PNG and SVG support. Using the fixture looks good too.

tests/test_circuit.py
```diff
@@ -709,27 +709,20 @@ def test_latex_code_non_reversed(self):
...
-        cleanup(file_png_200, file_png_400, file_svg)
+        if "png" in circuit_latex.CONVERTERS:
+            file_png200 = "exported_pic_200.png"
+            file_png400 = "exported_pic_400.png"
+            qc.draw("png", 200, file_png200.split('.')[0], "")
+            qc.draw("png", 400.5, file_png400.split('.')[0], "")
+            assert file_png200 in os.listdir('.')
+            assert file_png400 in os.listdir('.')
+            assert os.stat(file_png200).st_size < os.stat(file_png400).st_size
+        elif "svg" in circuit_latex.CONVERTERS:
```
srinathkp, line 725: Shouldn't we be using `if` here so that we can test both PNG and SVG if we have both converters installed.

BoxiLi: Oops, yes, it should be `if`.

PR #149, opened 2022-06-15
Fixes #148

### LINKED ISSUE #148: Casting complex values to real discards the imaginary part during run state
Hi! I found the following error during run state when using discrete pulse mode and having complex pulse coefficients

**qutip-qip/src/qutip_qip/pulse.py:646**
> ComplexWarning: Casting complex values to real discards the imaginary part
>   new_coeff[new_ind] = old_coeffs[old_ind]

Wikstahl:
This fixes issue #148

### DIFF
--- a/src/qutip_qip/pulse.py
+++ b/src/qutip_qip/pulse.py
@@ -631,7 +631,7 @@ def _fill_coeff(old_coeffs, old_tlist, full_tlist, args=None, tol=1.0e-10):
             old_coeffs = np.concatenate([old_coeffs, [0]])
         new_n = len(full_tlist)
         old_ind = 0  # index for old coeffs and tlist
-        new_coeff = np.zeros(new_n)
+        new_coeff = np.zeros(new_n, dtype=np.complex128)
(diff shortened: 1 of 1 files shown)


BoxiLi:
Hi @Wikstahl, thanks a lot for the PR!

It is probably better to set the `dtype` of the new coefficient array to be the same as the old one, instead of always using a complex array.

Could you also add a new test for the complex coefficients?

Wikstahl:
Sure I can fix this! :)

BoxiLi:
I just pushed a patch with `dtype=old_coeffs.dtype`

This should make the test pass. However, it does not mean that the complex coefficient is fully supported. The `Processor.plot_pulse` will still get a warning because of the complex numbers. But this at least lets the computation run without an error.

In fact, one can always rewrite complex coefficients by real ones with two `sx` and `sy` control terms. So real coefficient also has no loss of generality.

BoxiLi:
Close, we should support complex value properly in plotting if we want.

PR #155, opened 2022-06-27
qutip-qip backends for qiskit

claretgrace0801:
This draft PR tracks changes for the GSoC '22 project: making qutip-qip backends for qiskit.


tests/test_qiskit.py
```diff
@@ -0,0 +1,85 @@
+import pytest
+import sys
+import numpy as np
+from yaml import parse
```
BoxiLi, line 4: `parse` is imported but not used? The test fails because it is not installed. We can include it if necessary but it does not seem like the case?

claretgrace0801: That was included accidently. Removed it now.

src/qutip_qip/qiskit/converter.py
```diff
@@ -0,0 +1,160 @@
...
+    """
+    Map the bit index from qiskit to qutip.
+
+    When we convert a circuit from qiskit to qutip,
+    the 1st bit is mapped to the nth bit and so on.
+    """
+    return total_qubits - 1 - qiskit_index
+
+
+def qiskit_to_qutip(qiskit_circuit: QuantumCircuit) -> QubitCircuit:
```
BoxiLi, line 68: Maybe we can rename it `convert_qiskit_circuit`?

src/qutip_qip/qiskit/converter.py
```diff
@@ -0,0 +1,160 @@
...
+            None
+            if not qiskit_instruction.params
+            else qiskit_instruction.params
+        )
+        if not qiskit_instruction.params:
+            arg_value = None
+        elif len(qiskit_instruction.params) == 1:
+            arg_value = qiskit_instruction.params[0]
+        else:
+            arg_value = qiskit_instruction.params
```
BoxiLi, line 117: Duplication? Line 107-111 seems useless.

src/qutip_qip/qiskit/converter.py
```diff
@@ -0,0 +1,160 @@
...
+    for qiskit_index, qubit in enumerate(qiskit_circuit.qubits):
+        qubit_map[qubit] = _get_qutip_index(
+            qiskit_index, qiskit_circuit.num_qubits
+        )
+
+    clbit_map = {}
+    for qiskit_index, clbit in enumerate(qiskit_circuit.clbits):
+        clbit_map[clbit] = _get_qutip_index(
+            qiskit_index, qiskit_circuit.num_clbits
+        )
```
BoxiLi, line 92: We can probably update `_get_qutip_index` such that if the input is a list of integers, it iterates over it and calls itself.

It also has a few use cases later. So that you don't have to use `[qubit_map[qreg] for qreg in qiskit_qregs]`.

src/qutip_qip/qiskit/converter.py
```diff
@@ -0,0 +1,160 @@
...
+            qutip_circuit.add_gate(
+                _map_gates[qiskit_instruction.name],
+                targets=[qubit_map[qreg] for qreg in qiskit_qregs],
+                arg_value=arg_value,
+            )
+
+        elif qiskit_instruction.name in _map_controlled_gates.keys():
+            qutip_circuit.add_gate(
+                _map_controlled_gates[qiskit_instruction.name],
+                controls=[qubit_map[qiskit_qregs[0]]],
```
BoxiLi, line 130: Maybe add an in-line comment that we assume the first element in `qregs` is the control qubit. Maybe this is the convention used for the gate we support at the moment but I don't know if the same follows e.g. for the Toffoli gate.

src/qutip_qip/qiskit/backend.py
```diff
@@ -0,0 +1,172 @@
...
+            for i, count in enumerate(statistics.cbits):
+                counts[convert_to_hex(count)] = round(
+                    statistics.probabilities[i], 3
+                )
+        else:
+            counts = None
+
+        statevector = random.choices(
+            statistics.final_states, weights=statistics.probabilities
+        )[0]
```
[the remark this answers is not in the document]
claretgrace0801: Yep, that is correct.

src/qutip_qip/qiskit/backend.py
```diff
@@ -0,0 +1,172 @@
...
+        counts = {}
+
+        def convert_to_hex(count):
+            return hex(int("".join(str(i) for i in count), 2))
+
+        if statistics.cbits[0] is not None:
+            for i, count in enumerate(statistics.cbits):
+                counts[convert_to_hex(count)] = round(
+                    statistics.probabilities[i], 3
+                )
```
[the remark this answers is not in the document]
claretgrace0801: Changing this. Rounding is not required. Qiskit handles the rounding.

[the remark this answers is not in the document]
BoxiLi: It is better to keep the default as not rounded. Although the rounded one looks cleaner, it unnecessarily makes the result inaccurate. If we do want a better display, we should only round it when displaying, not the original result. But I guess `.data()` is meant to be raw data and one can define some `.probability` for nice looking and give it some parameters e.g. `truncation`.

BTW, why does `counts` returns a probability... I know that we don't repeat. So we probably don't have a well-defined `counts`? How does the qiskit circuit simulator handle this?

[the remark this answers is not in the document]
claretgrace0801: Qiskit backends have an option to set the number of `shots`. I haven't implemented any configurable backend options yet and will be doing that soon. After doing that, the probabilities will be replaced by the actual `counts`. We can use random sampling from the obtained probabilities for this purpose.

src/qutip_qip/qiskit/backend.py
```diff
@@ -0,0 +1,199 @@
...
+        qutip_circ = convert_qiskit_circuit(qiskit_circuit)
+        job_id = str(uuid.uuid4())
+        job = Job(
+            backend=self,
+            job_id=job_id,
+            result=self._run_job(job_id, qutip_circ),
+            warnings=qutip_circ._warnings
+            if hasattr(qutip_circ, "_warnings")
+            else [],
+        )
```
BoxiLi, line 57: This part is a bit strange, the warning message should not be part of the circuit. I understand that you want to pass the message from the converter up to this level. But hiding it as an attribute looks very confusing.

There are standard ways to catch a warning message. Just like `try` `except` for the error message, you can also get write a `with` condition that catches warnings:
```python
with warnings.catch_warnings(record=True) as w:
    pass
```
But I think this is also not the best way to handle this here.

Maybe a better way is to the following:
- define a qutip-backend-unique configuration key e.g. `allow_custom_gate` in the configuration with the default `True`.
- Add a key-word parameter `allow_custom_gate=True` to `convert_qiskit_circuit`.
- When converting the circuit here, we read from the `BackendConfiguration` (There should be a way to access it here.) and pass it to `convert_qiskit_circuit`.
- If the custom gate is allowed. Use it without warning. Otherwise, we raise an *error*.

I think this makes sense especially because, for the pulse-level processor, we cannot use any custom gates.

src/qutip_qip/qiskit/converter.py
```diff
@@ -0,0 +1,175 @@
...
+    "crx": "CRX",
+    "cry": "CRY",
+    "crz": "CRZ",
+    "cp": "CPHASE",
+}
+
+_ignore_gates = ["id", "barrier"]
+
+
+def _make_user_gate(unitary: np.ndarray, instruction: Instruction):
```
BoxiLi, line 36: Let's make it clear that it is the qiskit Instracution, `qutip` also has its own `Instruction`.

BoxiLi (changes requested):
Looks good. The main comment is about the warning. Otherwise, I think it is ready!

src/qutip_qip/qiskit/converter.py
```diff
@@ -158,10 +168,12 @@ def convert_qiskit_circuit(qiskit_circuit: QuantumCircuit) -> QubitCircuit:
             pass
 
         else:
-            qutip_circuit._warnings.append(
-                f"Conversion warning: {qiskit_instruction.name} is not a \
-gate in qutip_qip. This gate will be simulated using it's corresponding unitary matrix."
-            )
+            if not allow_custom_gate:
+                raise Exception(
```
BoxiLi, line 172: `Exception` -> `RuntimeError`

src/qutip_qip/qiskit/backend.py
```diff
@@ -100,6 +118,31 @@ def __init__(self, configuration=None, provider=None, **fields):
...
+            Returns the Counts object sampled according to
+            the given probabilities and configured shots.
+        """
+        shots = self.options.shots
+        samples = random.choices(
+            list(count_probs.keys()), list(count_probs.values()), k=shots
+        )
+        counts = dict.fromkeys(count_probs.keys(), 0)
+        for sample in samples:
+            counts[sample] += 1
```
BoxiLi, line 143: There might be faster way to do this.

BoxiLi (approved):
Looks good

BoxiLi:
Seems that codeclimate is undergoing some connection error. I'll just merge.

PR #160, opened 2022-08-18
Fix access on arg_label

christian512:
The `arg_label` used here was undefined. `arg_label` is a member of `gate`, hence this should fix the problem.


BoxiLi:
Thanks! @christian512

PR #162, opened 2022-08-22
Final time calculation in `_concatenate_pulses`.

christian512:
The calculation of `final_time` was done by taking the last element of the `tlist` in `compiled_tlist`. However, the elements in `compiled_tlist` are again lists, where the first element is just a list and the second element are the actual times.
E.g. `tlist = [[0.0], np.array([...])]`

 So we have to take the second ( i.e. last) element from that and then the last element to get the largest time.
The previous version sometimes produced errors when using `processor.load_circuit(...)`.

### DIFF
--- a/src/qutip_qip/compiler/gatecompiler.py
+++ b/src/qutip_qip/compiler/gatecompiler.py
@@ -249,7 +249,7 @@ class GateCompiler(object):
                 compiled_tlist[pulse_ind].append(execution_time)
                 compiled_coeffs[pulse_ind].append(coeffs)
 
-        final_time = np.max([tlist[-1] for tlist in compiled_tlist])
+        final_time = np.max([tlist[-1][-1] for tlist in compiled_tlist])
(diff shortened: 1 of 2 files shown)


christian512:
The CI fails due to the error that is fixed in this PR: https://github.com/qutip/qutip-qip/pull/163

hodgestar:
@christian512 Do you have a test we can add for this?

BoxiLi:
This is very strange... I think they should be either all lists or all arrays. Let me have a closer look at it.

BoxiLi:
Yes, you are right. It wasn't showing up because in most of the examples all the sub `tlist`s have the same size and NumPy treat them as a 2D array. 
Let me try to construct a test for this.

@christian512 do you mind if I push to this branch?

hodgestar:
> @christian512 do you mind if I push to this branch?

@BoxiLi Christian is on holiday this week, so I think he is fine with you pushing to this branch at the moment.

PR #163, opened 2022-08-22
Fix usage of `rand_unitary` and `rand_ket`

christian512:
The signature of `rand_unitary` has changed recently in QuTiP 5. The argument `dims` no longer exists. This PR changes the two usages of `rand_unitary` by checking the version of QuTiP and then using the corresponding definition. 
The same holds for `rand_ket`.

The tests were recently failing due to this change in QuTiP 5.


BoxiLi (approved):
Thank you! @christian512

PR #167, opened 2022-09-07
Fix bug for circuit with selected measurement result

### LINKED ISSUE #164: Circuit.run ignores supplied measure_results.
Constructing a simple circuit like so:
```python
qc = qip_circuit.QubitCircuit(N=1, num_cbits=1)
qc.add_gate("SNOT", targets=0)
qc.add_measurement("M0", targets=0, classical_store=0)
```
and then performing
```python
classical_store = [None]
final_state = qc.run(q0, cbits=classical_store, measure_results=[0])
```
Does not always return the `0` state.

The issue is that `.run()` passes `measure_results` to the `CircuitSimulator` constructor, but not to `CircuitSimulator.run` which then overwrites `measure_results` with `None`.

Also, the docstring for `.run_statistics` documents `measure_results` but doesn't take it as a parameter. This should just be removed from the docstring.

BoxiLi:
The parameter measure_results was not correctly processed in QubitCircuit.run because the Simulator is initialized twice. This commit removes the parameters state, cbits and measure_results from the `__init__` of CircuitSimulator and only initializes the simulator before it is run.

The CircuitSimulator class should only include information about the circuit while the information about the state is only provided when running the simulation.

Fix #164

### DIFF
--- a/doc/source/qip-simulator.rst
+++ b/doc/source/qip-simulator.rst
@@ -156,8 +156,8 @@ and computation proceeds. To demonstrate, we continue with our previous circuit:
 .. testcode::
 
   from qutip_qip.circuit import CircuitSimulator
-
-  sim = CircuitSimulator(qc, state=zero_state)
+  sim = CircuitSimulator(qc)
+  sim.initialize(zero_state)
 
 This initializes the simulator object and carries out any pre-computation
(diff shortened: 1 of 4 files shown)


hodgestar (approved):
Thank you for fixing this Boxi. Looks perfect to me.

PR #168, opened 2022-09-08
Fix bugs in to_chain_structure

### LINKED ISSUE #165: Chain transpiler calls QubitCircuit with incorrect arguments
The method `to_chain_structure` calls:
```python
qc_t = QubitCircuit(qc.N, qc.reverse_states)
```
but the second argument to `QubitCircuit` is `input_states`. It should call
```python
qc_t = QubitCircuit(qc.N, reverse_states=qc.reverse_states)
```
It should probably copy more arguments from the circuit being modified too (e.g. input states, output states, num_cbits).

BoxiLi:
- Copy the circuit correctly in to_chain_structure
- Correctly handle measurement in the circuit.

Fix #165

### DIFF
--- a/src/qutip_qip/transpiler/chain.py
+++ b/src/qutip_qip/transpiler/chain.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 from ..circuit import QubitCircuit
 
 
@@ -27,7 +28,8 @@ def to_chain_structure(qc, setup="linear"):
     # FIXME This huge block has been here for a long time.
     # It could be moved to the new compiler section and carefully
     # splitted into smaller peaces.
-    qc_t = QubitCircuit(qc.N, qc.reverse_states)
+    qc_t = deepcopy(qc)
+    qc_t.gates = []
     swap_gates = [
         "SWAP",
         "ISWAP",
@@ -265,12 +267,8 @@ def to_chain_structure(qc, setup="linear"):
                     j = j + 1
 
         else:
-            qc_t.add_gate(
-                gate.name,
-                gate.targets,
-                gate.controls,
-                gate.arg_value,
-                gate.arg_label,
-            )
+            # This gate can be general quantum operations
+            # such as measurement or global phase.
+            qc_t.add_gate(gate)
 
     return qc_t
--- a/src/qutip_qip/transpiler/chain.py
+++ b/src/qutip_qip/transpiler/chain.py
@@ -28,8 +28,7 @@ def to_chain_structure(qc, setup="linear"):
     # FIXME This huge block has been here for a long time.
     # It could be moved to the new compiler section and carefully
     # splitted into smaller peaces.
-    qc_t = deepcopy(qc)
-    qc_t.gates = []
+    qc_t = qc._create_empty_copy()
     swap_gates = [
         "SWAP",
         "ISWAP",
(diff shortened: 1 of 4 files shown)


src/qutip_qip/transpiler/chain.py
```diff
@@ -27,7 +28,8 @@ def to_chain_structure(qc, setup="linear"):
     # FIXME This huge block has been here for a long time.
     # It could be moved to the new compiler section and carefully
     # splitted into smaller peaces.
-    qc_t = QubitCircuit(qc.N, qc.reverse_states)
+    qc_t = deepcopy(qc)
```
hodgestar, line 31: Does QubitCircuit implement its own copy support? E.g. an implementation of `__copy__` or `__deepcopy__`. If not, I recommend adding it now and not relying on the default deepcopy which just copies all attributes and is not guaranteed to be sensible for this case.

hodgestar (changes requested):
Properly copying the circuit looks like the right approach, but let's implement it properly on the circuit class itself.

src/qutip_qip/transpiler/chain.py
BoxiLi: Thanks for the quick feedback! What would be the advantages of implementing our own __copy__? I feel like the implementation will just be copying a few attributes. And if a new attribute is added, the method also has to be updated every time.

hodgestar: It's a bit tough to explain succinctly, because in the very simple cases the default copy will work. The issue is that it is not *guaranteed* to work and there are many many possible edge cases where it won't work. This is why `__copy__` and `__deepcopy__` exist.

If you don't explicitly implement `__copy__` then someone modifying `QubitCircuit` might do something that causes `__copy__` to do the wrong thing, and then code that calls `copy` would be safe.

If we implement `__copy__` then we are committing to supporting `copy(qc)` working. I.e. we're making it part of our API, and that seems like a good thing. Without `__copy__` we are making no such guarantee.

QubitCircuit maintains quite a lot of internal state and just copying that state might do the wrong thing. I wouldn't be able to tell for sure that the default `copy` does the right thing without checking thoroughly, and that is part of why we should write our own `__copy__` -- because it would be guaranteed to do the right thing for QubitCircuit while the default would not (boh now and in the future).

I'm about 90% sure I could dig into QubitCircuit and find a problem, but whether there actually is an issue is not the point. The point is to provide certainty that `copy` will work and to make it clear to ourselves that we are committing to make it work.

It's entirely possible that the implementation of `__copy__` might be somewhat trivial, and that's completely okay.

BoxiLi:
Not really very confident about my implementation. Let's discuss next week @hodgestar  :)

hodgestar:
@BoxiLi Let's review and rework a bit if needed on Monday.

BoxiLi:
Hi Simon, would it be ok if we just fix the error by using `deepcopy` here and move a full functional implementation of `Circuit.copy()` to another PR? Of course a test is till missing here.

hodgestar (approved):

hodgestar:
I'm also okay with a simpler fix if you prefer.

BoxiLi:
Fix the bug in `to_chain_structure` only. Leave the circuit copying feature for a different PR.

The failing test is because of q-ctrl. We should use the master branch instead of the released ones for the check of qutip.master branch.

hodgestar:
I see there is also a warning for `DeprecationWarning: invalid escape sequence '\s'` in the tests that failed, but that can be left for another PR.

BoxiLi:
> I see there is also a warning for `DeprecationWarning: invalid escape sequence '\s'` in the tests that failed, but that can be left for another PR.

Yes... That is likely the docsting using LaTeX without adding an `r`.

PR #169, opened 2022-09-09
Fix bugs in the analytical pulse shapes

BoxiLi:
- Fix bugs in the analytical pulse shapes
- Compress the test. The test no longer depends on the processor class. Test both the maximum and the total area.


hodgestar (approved):
Codeclimate is unhappy, but this fix looks good to me.

PR #170, opened 2022-09-11
Add a minus sign for plotting the swap gate

### LINKED ISSUE #166: Swap gates are incorrectly rendered in LaTeX
The function `latex_code()` in `circuit.py` does:
```python
col.append(r" \qswap \qwx[%d] \qw" % distance)
```
but the `qcircuit` package expects this to be `\qwx[-%d]` -- note the minus sign. This causes the pdflatex compilation to fail if the swap goes below the circuit (at least recent enough versions of qcircuit throw an error).

See https://mirrors.evoluso.com/CTAN/graphics/qcircuit/qcircuit.pdf for the qcircuit documentation.

BoxiLi:
Fix #166

### DIFF
--- a/src/qutip_qip/circuit/circuit.py
+++ b/src/qutip_qip/circuit/circuit.py
@@ -1171,7 +1171,7 @@ class QubitCircuit:
 
 
 _latex_template = r"""
-\documentclass{standalone}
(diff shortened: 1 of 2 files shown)


hodgestar (approved):
Code Climate has a complaint, but this look good to me.

PR #173, opened 2022-09-25
Align the signature of gate functions and expand_operator with the ones in qutip-v5

### LINKED ISSUE #171: Align the signatures of gate functions with qutip-v5
In qutip-v5 the gate functions no longer take parameters `N` , `targets` and `control`, which permute the qubits if needed and generate gate operators in a larger Hilbert space. Although many of the gates are moved to `qutip.core.gates`, these additional parameters are removed in qutip-v5. Also, the signature of `expand_operator` is slightly changed to be more general.

To avoid confusion and code duplication, we should adapt the functions in `qutip-qip` accordingly.

Proposals:
- Import the gate functions from qutip when v5 is used. Otherwise still use the one defined in `qutip-qip`.
- Align the signatures of gate functions and `expand_operator` with qutip-v5, and issue deprecation warnings when needed.

Optional:
- In principle, we can try to keep the compatibility and define a wrapper that expands the obtained `Qobj`. This will need some complications but should be feasible.
- Do we also want to add some deprecation warning for this to the gates in qutip-v4 in the module `qutip.qip`. Some people are still using it. It would make the migration from 4 to 5 easier.

@hodgestar Do you think we should implement the optional ones?

BoxiLi:
- Remove deprecated old modules and functions
- Update the signatures of gate functions and `expand_operator`, we keep them consistent with the ones in qutip-v5 to avoid confusion. In particular, to expand the `Qobj` generated by a gate function or permute the qubits, one needs to apply `expand_operator` separately.

close #171


src/qutip_qip/operations/gates.py
```diff
@@ -73,7 +79,8 @@ def x_gate(N=None, target=0):
 
     """
     if N is not None:
-        return gate_expand_1toN(x_gate(), N, target)
+        _deprecation_warnings_gate_expansion()
```
hodgestar, line 82: I'm a bit confused about why we issue the deprecation warning here since we still call `expand_operator` so presumably what this function returns is still the expanded operator? What am  I not understanding?

BoxiLi: Hi, Simon, thanks for reviewing!

The change here is two-folded. It consists of
- Removing the use of `gate_expand_2toN`, which was deprecated a long time ago.
- Deprecating the expansion of gate unitaries in the gate function. To be consistent with qutip-5.

hodgestar: Ah, I see.

Can the expansion be done in a consistent way for more complex gates? E.g. two, three or four qubit gates?

BoxiLi: Yes, `expand_operator` works for arbitrary qudits gate. It is also included `qutip-v5`.

BoxiLi: Ah, you mean actually a two-qubit gate with the `dims=[[2,2],[2,2]]`, not `dims=[[4],[4]]`. The current version should also work. There is an example in the function doc, expanding a CNOT gate.

hodgestar (approved):

PR #177, opened 2022-11-09
SWAP gate for (non-) reversed circuit

BoxiLi:
SWAP needs to be treated differently for circuits with reversed and non-reversed states order.

The fix in https://github.com/qutip/qutip-qip/pull/170 was not complete. The reason for using `\qwx[%d]` without the minus sign is actually because in `qutip-qip` the order of states is reversed.

### DIFF
--- a/src/qutip_qip/circuit/circuit.py
+++ b/src/qutip_qip/circuit/circuit.py
@@ -967,7 +967,9 @@ class QubitCircuit:
                                 distance = abs(
                                     gate.targets[1] - gate.targets[0]
                                 )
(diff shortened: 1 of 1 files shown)


hodgestar (approved):
Code change looks good. Is it possible to add some sort of test for this? If it's very hard, I am okay with skipping it. It would just be nice because we keep getting this wrong. :)

BoxiLi:
Indeed, it would be nice to have some tests indeed. In the test actions of `qutip-qip` we don't yet install texlive and other things. So some more work is needed to properly test circuit plotting. Let's maybe do it in a separate PR. Opened an issue for it #178.

PR #180, opened 2022-12-03
Improve the docs for QubitCircuit.add_gate

### LINKED ISSUE #179: Regarding the types of the parameters of QubitCircuit.add_gate
Currently the parameters for `QubitCircuit.add_gate` [require](https://qutip.org/docs/latest/apidoc/classes.html#qutip.qip.circuit.QubitCircuit.add_gate):

1. targets and controls to be a list
2. arg_value to be a float

1\. is fine, however just integers work as in the method it's straight up put in a `Gate` object [which does accept both integers and lists](https://qutip.org/docs/latest/apidoc/classes.html#qutip.qip.Gate).
In one of your examples in `Scheduler.schedule` [the targets and controls are also inputted as plain ints](https://qutip.org/docs/latest/apidoc/classes.html#qutip.qip.compiler.Scheduler.schedule), which some IDEs will mark as errors or for example mypy just straight up gives it an error in CI/CD if you run a type checker.
So my suggestion is to update the type hint in `add_gate` to also include integers inline with the `Gate` object.

My main issue is with 2. `arg_value` is an object that is put raw in a (potentially) user defined method for the gate. Two issues with that:
1. It doesn't have to be a float. Might be some situations were you don't input an angle?
2. It can be more than one argument

For example if you have a gate that implements a general single qubit you need two angles to define that: `def r_gate(theta: float, phi: float) -> Qobj`. [However at most one parameter is allowed](https://github.com/qutip/qutip-qip/blob/17d37afc24ee7258ee1192a42bad91a84af64ee8/src/qutip_qip/circuit/circuit.py#L911-L918)

So for now I have done something like `qc.add_gate("r_gate", targets=[q], arg_values=(np.pi, 0))`, (which gives a type error because `(np.pi, 0)` is a tuple, not a float). And then in the method I need to add `theta, phi = args`.  
Whereas ideally I can just do `qc.add_gate("r_gate", targets=q, theta=np.pi, phi=0)` or `qc.add_gate("r_gate", targets=q, args={"theta": np.pi, "phi": 0}`.
So I request to change the `arg_value` with a proper `**args` that are directly inputted into the gate function as arguments. 
Or at least change the type of `arg_value` to from float Any

BoxiLi:
Fix #179

### DIFF
--- a/src/qutip_qip/circuit/circuit.py
+++ b/src/qutip_qip/circuit/circuit.py
@@ -203,22 +203,24 @@ class QubitCircuit:
         gate: string or :class:`~.operations.Gate`
             Gate name. If gate is an instance of :class:`~.operations.Gate`,
             parameters are unpacked and added.
-        targets: list
-            Gate targets.
-        controls: list
-            Gate controls.
-        arg_value: float
-            Argument value(phi).
-        arg_label: string
+        targets: int or list, optional
+            Index for the target qubits.
+        controls: int or list, optional
+            Indices for the (quantum) control qubits.
+        arg_value: Any, optional
+            Arguments for the gate. It will be used when generating the
+            unitary matrix. For predefined gates, they are used when
+            calling the ``get_compact_qobj`` methods of a gate.
+        arg_label: string, optional
             Label for gate representation.
-        index : list
+        index : list, optional
             Positions to add the gate. Each index in the supplied list refers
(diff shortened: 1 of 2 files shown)


hodgestar (approved):

hodgestar:
I see the test against dev.major failed, but that seems unrelated to this PR.

PR #181, opened 2022-12-08
Fix parsing of package versions in packaging 22.

BoxiLi:
Copied from https://github.com/qutip/qutip/pull/2037


hodgestar (approved):

hodgestar:
The tests won't pass until we also apply the fix to qutip dev.major.

PR #182, opened 2022-12-08
Filter the measured probability

BoxiLi:
Use the `atol` option in qutip to determine the tolerance for measurement. And filter out the states with a probability smaller than this.

See also: https://github.com/qutip/qutip/issues/2029


hodgestar (approved):
It would be good to have a test for this, but the change looks right.

BoxiLi:
The changes are moved to the basic measurement class in `qutip-qip`. It is a lower-level function so we don't need to apply the filtering twice in the code. A test is added.

hodgestar (approved):

PR #183, opened 2022-12-09
Add a default max_step to the processors

BoxiLi:
Similar to https://github.com/qutip/qutip/pull/2040, with a separate handling of `options` for qutip-v5.

### DIFF
--- a/doc/pulse-paper/qft.py
+++ b/doc/pulse-paper/qft.py
@@ -11,7 +11,7 @@ plt.rcParams.update({"text.usetex": False, "font.size": 10})
 
 
 import numpy as np
-from qutip import basis, fidelity
+from qutip import basis, fidelity, Options
 from qutip_qip.device import LinearSpinChain
 from qutip_qip.algorithms import qft_gate_sequence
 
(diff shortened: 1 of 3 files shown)


BoxiLi:
Although this does solve the skipping pulse issue. On my computer, it increases the total time for the tests from 40 seconds to 3 min...

BoxiLi:
The main slow down is actually at the computation of the full tlist, not the time evolution.

hodgestar:
That is quite a big slowdown. Can we make `get_full_tlist` more efficient? Or avoid calculating the full tlist and just get the smallest difference?

BoxiLi:
Having a closer look again, the main problem is actually with the `CircuitQED`, where a much smaller sampling step is needed for `coeff` to represent a continuous pulse shape. What we get from `get_full_tlist` is only the smallest sampling step size in tlist, rather than the gate time. The gate time information is not kept, unfortunately. What you suggest in the other repo in `qutip`, saving the `max_step` in QobjEvo would be helpful.

This problem does not exist in `qutip.qip` inside qutip--v4, because there is no `CircuitQED` model.

BoxiLi:
An alternative is to use something like one-tenth of the total circuit pulse time. This does not solve the problem completely. But it will avoid those mistakes that keep popping up when people try and compile one gate and send it to the solver. When there are more gates and a non-trivial circuit, the probability of this happening will be very low.

hodgestar:
I'm okay with a hackier solution like `0.1 * total_time` for now, and a comment explaining the problem and perhaps referencing a GitHub issue that explains the difficulties.

hodgestar (approved):

PR #186, opened 2022-12-11
Update the github actions

BoxiLi:
- Add back the test against qutip@master (which was removed because @master was labelled as version 5.0.0 and could not be distinguished from @dev.major)
- Add a test against python 3.11
- Redistribute the test cases
- Update the versions of the GitHub built-in actions

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,47 +6,48 @@ on:
   pull_request:
 
 jobs:
-  test:
+  cases:
+    name: ${{ matrix.os }}, python${{ matrix.python-version }}, ${{ matrix.case-name }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         include:
-          - case-name: Ubuntu qutip@dev.major
+          - case-name: qutip@dev.major
             os: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -6,50 +6,58 @@ on:
...
             qutip-version: '@dev.major'
             qiskit-version: ''
             pyqir-version: ''
             python-version: '3.10'
-          - os: windows-latest
-            qutip-version: '==4.6.*'
-            qiskit-version: ''
+          - case-name: qutip@master
+            os: ubuntu-latest
+            qutip-version: '@master'
```
hodgestar, line 23: It's likely that we'll soon freeze qutip master and move development of 4.7.X to  `qutip-4.7.X` so that we can eventually use master for v5. You could switch to 4.7.X now, or just do it later when master switches over.

hodgestar (approved):

.github/workflows/test.yml
BoxiLi: Thank! Yes, that makes sense. All the tests should still pass in that case. I'll switch later.

PR #189, opened 2023-01-19
Remove tests again the `dev.major` branch of Qutip.

Ericgig:
We merged the branch `dev.major` branch into `master` last Monday. Thus tests against both of these branches are redundant.
I remove the test using `qutip@dev.major` since all further work for v5 will be done in `master`. It was the only test using python 3.10, so updated another test to use it.

There are no longer any active development branch for v4, all test using qutip v4 are ran using released version.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,6 +12,11 @@ jobs:
     strategy:
       matrix:
         include:
+          - case-name: qutip@qutip-4.7.X
+            os: ubuntu-latest
+            qutip-version: '@qutip-4.7.X'
+            pyqir-version: ''
+            python-version: '3.10'
           - case-name: qutip@master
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -12,12 +12,6 @@ jobs:
     strategy:
       matrix:
         include:
-          - case-name: qutip@dev.major
```
hodgestar, line 15: I think this case should stay but become `qutip@master` and then the case below should become `qutip@qutip-4-7-X`.

hodgestar: We will still make fixes to 4.7.X and qutip-qip should still test against it.

BoxiLi (approved):

hodgestar (approved):

Issue #184, opened 2022-12-10
Keep the information of gate duration after the compilation


BoxiLi:
Currently, the compiler only returns the compiled pulse `cofficients` and `tlist`. It is sufficient for the solvers, but a few options in the solver may need more information. For instance, `max_step` is needed to prevent the solver from skipping pulses after a long idling period or initial states that are invariant under the given Hamiltonian (https://github.com/qutip/qutip/issues/2003, https://github.com/qutip/qutip/pull/2040).

This can be realized in 2 different ways:
- The `Compiler` returns more information. On top of the compiled pulses, we could let the compiler return also additional information such as the minimum gate time.
- Include the information in e.g. in the `Coefficents` class of qutip-v5 and return the `Coefficient` Object. This may also provide the opportunely to support different max step sizes for different gates. But it needs considerably more work and may need changes in the core qutip.

PR #190, opened 2023-01-19
Fix bug resulting from changing v5's default solver

Ericgig:
We removed the old solvers that were existing in parallel to the new ones in qutip v5. This broke a few tests.

- Solvers options are now simple dict. The default options are per solver basis and unknown to `qutip.Options`.
- `qutip.Options` and `qutip.SolverOptions` are equivalent and deprecated.
- `mcsolve` can accept density matrix input, it use `state=` instead of `rho0=`.
- `array(qobj)` should have been not working for a while now, `full()` is the method preferred when converting to a numpy array.
- I had to reduce the `max_step` for one test... Unless there was a pulse thinner than the previous value and the solver happen to be lucky and find it previously, I don't understand why I had to reduce it that much. I hope that the idea in #184, when implemented, will be able to bring some light on the issue.

### DIFF
--- a/src/qutip_qip/device/processor.py
+++ b/src/qutip_qip/device/processor.py
@@ -1221,7 +1221,7 @@ class Processor(object):
             )
         elif solver == "mcsolve":
             evo_result = mcsolve(
-                H=noisy_qobjevo, state=init_state, tlist=tlist, **kwargs
+                noisy_qobjevo, init_state, tlist=tlist, **kwargs
             )
 
         return evo_result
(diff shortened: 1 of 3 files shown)


hodgestar (approved):
Looks good to me, but I think @BoxiLi should approve the max_step change because he knows the most about how this works and I'd like to make sure he sees it.

Ericgig:
Yes `Options` is only there for compatibility and will raise a warning on use. 

The test needing smaller max_step was `test_device.test_numerical_evolution[DispersiveCavityQED, SNOT]`.

BoxiLi (approved):
It is likely just a coincidence. The pulse consists of 2 long pulses and 2 short pulses. They are quite unevenly distributed. I guess the new solver has a different way of estimating the step size.

It is not very bad to reduce it from 1/10 to 1/25 of the total circuit time, since it is only relevant when testing circuits with very few gates.

PR #192, opened 2023-02-18
Documentation compatibility with `qutip-v5`

BoxiLi:
- Remove some unnecessary branching that was used for the compatibility of `qutip-v5`. They are no longer needed in the new alpha version.
- Update the pulse paper examples to be compatible with `qutip-v5`.
- Update the documentation to be compatible with `qutip-v5`.

Note that `doctest` still doesn't fully pass with `qutip-v5` because of the spaces in printed `Qobj._repr_`. But this should be fine if we only build the doc against `qutip-v4.7` for now.


BoxiLi:
@hodgestar 
DO you think if worth have something like this in `qutip-v5`?

https://github.com/qutip/qutip-qip/blob/8c3af6644a8b16aedfa3af3b3dc85211cc58b8ef/src/qutip_qip/utilities.py#L8-L16

hodgestar (approved):

hodgestar:
We should definitely have a nice solution for cases like `q.dag() * q` which return a scalar in v5 and a Qobj in v4, but I don't think that solution can be in `qutip`. We could add a function to v5, but no such function is in v4 and we're not planning a 4.8. So the function probably has to be in libraries that use `qutip`, like `qutip_qip`.

Rather than the function you have now, I recommend creating a `qutip_qip.compat` module (so one can easily remove these later) and inside having a function like:

```python

def to_scalar(qobj_or_scalar):
    if isinstance(qobj_or_scalar, Qobj):
        if qobj_or_scalar.dims == [[1], [1]]:   # <-- maybe this check should also handle more dims?
            return qobj_or_scalar[0, 0]
        else:
            raise ValueError(...)
    return qobj_or_scalar
```

This avoids explicit version checks and implements a more specific set of functionality.

BoxiLi:
Yes, the example you provide is better, avoiding the version check. There are indeed other cases instead of just `dims=[[1],[1]]` but this covers >99% of situations.

And now thinking about it again. You are also right that unless we can have this for both qutip-v4 and v5, it is not gonna be sufficient. Maybe we could just provide the example above as part of the upgrade guide.

PR #193, opened 2023-02-18
Add QROT and MS gate

BoxiLi:
Add the QROT gate class.


BoxiLi:
Hi @ajrazander Would you like to have a review?

Does this definition of single-qubit rotation make sense? Or is there a better name than `QROT`?

ajrazander:
The gate looks good! The `QROT` name could be improved. In the literature it's usually referred to as just `R`. I kind of like that as it fits with the naming scheme that already exists for the `Rx`, `Ry`, and `Rz` gates. Using `R` would also put it in the list of gates near the `Rx`, `Ry`, `Rz` gates which would make searching for it easier and more recognizable for new users.

PR #194, opened 2023-03-01
Added Rxx and Ryy to gateset

ajrazander:
Added Rxx and Ryy two-qubit gates to the qutip-qip's gateset. Both of these are native gates for trapped ions (just a matter of laser phase).

Still to do/discuss:
[] The Rxx gate is the same as the current `molmer_sorensen` gate. Do we keep both or only one?
[] What's the process for adding tests for these two gates?


BoxiLi:
Thank you!

> The Rxx gate is the same as the current molmer_sorensen gate. Do we keep both or only one?

This is a good question. We should reduce the redundancy as much as possible. Keeping both of them will be very confusing for others.

Now thinking about the similarity between MS gate and RXX/RYY. Do you think it makes sense to put two of them under the name `molmer_sorensen`? We could add an additional argument to it
```molmer_sorensen(theta, phase="x", N=None, targets=[0, 1]) ```
and return the corresponding matrix.

Maybe we could also make the phase more general as a complex number? But I don't know if that has a physical meaning.

> What's the process for adding tests for these two gates?

What you have added so far are just functions that return the numpy array. To make them accessible through gates, you also need to add the gate class and make this gate available in a few other places. Like in https://github.com/qutip/qutip-qip/pull/193/files

BoxiLi:
BTW, the lint errors are fixed in the main branch. If you pull from the master it should be fixed.

ajrazander:
I removed the rxx and ryy gates and added the phase term you suggested to the already existing ms gate! :)

BoxiLi (changes requested):
Thank you for the updates! It looks good. Added a few suggestions.

You could use `pytest qutip_qip/tests` to run the test locally and check if they succeed. And `black qutip_qip/src` for the lint

ajrazander:
Thanks Boxi! I really appreciate your guidance through this!

Running `black` (23.1.0) on `qutip-qip/src` affected some files I'm not working on:
`circuit/circuit.py`, `circuit/circuitsimulator.py`, `device/processor.py`, `qasm.py`, `qir.py`, `qiskit/backend.py`, `transpiler/chain.py`

Show I not include these file changes in the commit?

BoxiLi:
Happy to help!

No, are you using the newest version of balck? It probably has something to do with the recent updates of black.  Or you can also just fix those shown in the failed check below. Click on them and you will see the details.

BoxiLi (approved):

BoxiLi:
Merged! Thank you! @ajrazander Congrats!

I just notice that you were working on your master branch. It is almost always better to start a different development branch and keep your master synced with the upstream one. This could be helpful e.g. when you have multiple ongoing independent works.

PR #198, opened 2023-04-16
Install qutip-qtrl master version for the tests

BoxiLi:
- Test against the qutip alpha release.
- Install qutip-qtrl from master

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,6 +17,11 @@ jobs:
             qutip-version: '==5.0.0a1'
             pyqir-version: ''
             python-version: '3.10'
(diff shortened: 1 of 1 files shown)


hodgestar (changes requested):
It would perhaps be nice to keep the `qutip-4.7.X` build to be sure we don't break `qutip-qip` on the `qutip-4.7.X` branch before a release.

I think installing `qutip-qtrl` from master when testing qutip master makes sense.

BoxiLi:
Thanks Simon! Yes, I can add the 4.7.X check back. I was thinking that not many changes happen now on 4.7.X.

hodgestar (approved):

PR #200, opened 2023-04-17
Cleaning a few warnings in the test

BoxiLi:
- pytest warning to error
- Use raw string literals for docstrings with LaTeX
- Skip plotting tests correctly when matplotlib not installed

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -195,10 +195,7 @@ def test_numerical_circuit(circuit, device_class, kwargs, schedule_mode):
     "processor_class",
     [DispersiveCavityQED, LinearSpinChain, CircularSpinChain, SCQubits])
 def test_pulse_plotting(processor_class):
-    try:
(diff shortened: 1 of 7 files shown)


tests/test_device.py
```diff
@@ -5,6 +5,7 @@
 import warnings
 import numpy as np
 import pytest
+plt = pytest.importorskip("matplotlib.pyplot")
```
hodgestar, line 8: Will this skip all the tests in this file?

BoxiLi: You are right this skips the whole test module. Importation has been moved inside the test function.

hodgestar (approved):

hodgestar (approved):

BoxiLi:
Thank you @hodgestar , just narrowed down the range of this warning filter a bit more.

hodgestar (approved):
Still looks good to me. Thank you for cleaning this up!

PR #202, opened 2023-04-17
Update processor.py

caoyingkang:
a typo in the docstring of the class `Processor`


BoxiLi:
Thank you for the correction! @caoyingkang

PR #218, opened 2023-09-27
Remove comments after meaningful statements in QASM reader

NoriyukiK-1qbit:
QASMBench has lines that have comments after instructions, and those lines cause errors. This fix remove such comments.

### DIFF
--- a/tests/test_qasm.py
+++ b/tests/test_qasm.py
@@ -145,5 +145,5 @@ def test_read_qasm():
         qc = read_qasm(filepath)
         qc2 = read_qasm(filepath2)
         assert True
-    except SyntaxError: # inline comments caused SyntaxError
+    except SyntaxError:  # inline comments caused SyntaxError
         assert False
--- a/tests/test_qasm.py
+++ b/tests/test_qasm.py
@@ -134,3 +134,15 @@ def test_export_import():
 
     for u0, u1 in zip(props, read_props):
         assert (u0 - u1).norm() < 1e-12
+
+def test_read_qasm():
+    filename = "w-state.qasm"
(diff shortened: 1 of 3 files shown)


BoxiLi:
Thank you! Could you add a test to `test_qasm.py` and verify that it works after adding these lines?

NoriyukiK-1qbit:
Hi BoxiLi,
Thanks for your comment. I added the test, and it worked as expected. Please find added files.
NK

tests/test_qasm.py
```diff
@@ -134,3 +134,16 @@ def test_export_import():
...
+    filename = "w-state.qasm"
+    filepath = Path(__file__).parent / 'qasm_files' / filename
+    filename2 = "w-state_with_comments.qasm"
+    filepath2 = Path(__file__).parent / 'qasm_files' / filename2
+    try:
+        qc = read_qasm(filepath)
+        qc2 = read_qasm(filepath2)
+        assert True
+    except SyntaxError:  # inline comments caused SyntaxError
+        assert False
```
BoxiLi, line 149: There is no need to use try except here. If an error is raised during the execution, pytest will catch it and report the error with the corresponding error message. If the error is caught manually and replaced by assert (which throw an error if False), it actually hides the true error message.

BoxiLi (changes requested):
Thank you! Looks good! I left a comment with a minor change in the test file.

NoriyukiK-1qbit:
Hi!
try/catch is removed as requested.

BoxiLi (approved):

BoxiLi:
@NoriyukiK-1qbit Thank you!

PR #225, opened 2023-12-10
Simplify the structure of `CircuitSimulator`

BoxiLi:
Refactor the gate level `CircuitSimulator`
- Avoid expanding all propagators and then applying them one by one. Only expand when applying the unitary. This saves memory and also make it easier for further improvement, i.e., exploring other ways to perform the matrix multiplication (e.g. `einsum`).
- Switch some public members to private. Those members that track the execution of the circuit evaluation should not be public.
- Small changes to improve the clarity.

### DIFF
--- a/src/qutip_qip/circuit/circuitsimulator.py
+++ b/src/qutip_qip/circuit/circuitsimulator.py
@@ -1,14 +1,11 @@
 from itertools import product, chain
-import os
 
 import numpy as np
 
-from . import circuit_latex as _latex
 from ..operations import (
     Gate,
     Measurement,
     expand_operator,
-    gate_sequence_product,
 )
 from qutip import basis, ket2dm, Qobj, tensor
 import warnings
@@ -225,7 +222,9 @@ def _gate_sequence_product(U_list, ind_list):
 
 def _gate_sequence_product_with_expansion(U_list, left_to_right=True):
     """
-    Calculate the overall unitary matrix for a given list of unitary operations.
+    Calculate the overall unitary matrix for a given list of unitary
+    operations. Assuming that all unitary has the same dimension.
+    This is only for backward compatibility.
 
     Parameters
     ----------
@@ -354,13 +353,13 @@ class CircuitSimulator:
         """
         prev_index = 0
         U_list_index = 0
-        inds_list = []
+        ind_list = []
 
         for gate in circuit.gates:
             if isinstance(gate, Measurement):
                 continue
             else:
-                inds_list.append(gate.get_all_qubits())
+                ind_list.append(gate.get_all_qubits())
 
         U_list = circuit.propagators(expand=False, ignore_measurement=True)
 
@@ -371,7 +370,7 @@ class CircuitSimulator:
                     ops.append(
                         self._compute_unitary(
                             U_list[prev_index:U_list_index],
-                            inds_list[prev_index:U_list_index],
+                            ind_list[prev_index:U_list_index],
                         )
                     )
                     prev_index = U_list_index
@@ -383,7 +382,7 @@ class CircuitSimulator:
                         ops.append(
                             self._compute_unitary(
                                 U_list[prev_index:U_list_index],
-                                inds_list[prev_index:U_list_index],
+                                ind_list[prev_index:U_list_index],
                             )
                         )
                         prev_index = U_list_index
@@ -397,7 +396,7 @@ class CircuitSimulator:
             ops.append(
                 self._compute_unitary(
                     U_list[prev_index:U_list_index],
-                    inds_list[prev_index:U_list_index],
+                    ind_list[prev_index:U_list_index],
                 )
             )
             prev_index = U_list_index + 1
@@ -452,7 +451,7 @@ class CircuitSimulator:
         self.measure_results = measure_results
         self.measure_ind = 0
 
-    def _compute_unitary(self, U_list, inds_list):
+    def _compute_unitary(self, U_list, ind_list):
         """
         Compute unitary corresponding to a product of unitaries in U_list
         and expand it to size of circuit.
@@ -462,7 +461,7 @@ class CircuitSimulator:
         U_list: list of Qobj
             list of predefined unitaries.
 
-        inds_list: list of list of int
+        ind_list: list of list of int
             list of qubit indices corresponding to each unitary in U_list
 
         Returns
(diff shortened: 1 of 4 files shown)


BoxiLi:
[Remove the support of precompute_unitary](https://github.com/qutip/qutip-qip/pull/225/commits/e419b912b9c0170dcab006790ab6ff66207a7cc9)
This option computes the full unitary of all the gates and caches it. It is beneficial if the circuit has a small number of qubits but a very long depth. This feature makes the migration to the new efficient circuit evaluation very hard because the input gates are `Qobj`s and not gates. If the unitary for the circuit is desired, one can just call QubitCircuit.compute_unitary to get the circuit unitary and then apply it to different states.


[Use np.einsum for gate-state multiplication](https://github.com/qutip/qutip-qip/pull/225/commits/9e4806405138e70df0ef564fd27a4054cdd7d888)
Use np.einsum to apply the gate matrix directly on the state vector, instead of expanding it first to the full dimension.
- The private member self._state will remain a numpy array during the sequential application of gates.
- If a measurement occurs, the state has to be transformed to a Qobj.
- Returning the state after each step is removed as it inevitably creates Qobj and creates a copy in v4.


Some benchmarking for the circuit simulator after the improvement:
Note that the circuit simulator in the old version is very inefficient with v5 because `qc.propagators()` return a list of *dense* matrix! This needs to be deprecated (or fixed) in a different PR.
image

src/qutip_qip/circuit/circuitsimulator.py
```diff
@@ -225,7 +223,9 @@ def _gate_sequence_product(U_list, ind_list):
 
 def _gate_sequence_product_with_expansion(U_list, left_to_right=True):
     """
-    Calculate the overall unitary matrix for a given list of unitary operations.
+    Calculate the overall unitary matrix for a given list of unitary
+    operations. Assuming that all unitary has the same dimension.
+    This is only for backward compatibility.
```
hodgestar, line 228:
```suggestion
    Calculate the overall unitary matrix for a given list of unitary
    operations, assuming that all operations have the same dimension.
    This is only for backward compatibility.
```

src/qutip_qip/circuit/circuitsimulator.py
```diff
@@ -419,64 +308,68 @@ def initialize(self, state=None, cbits=None, measure_results=None):
...
-
-        inds_list: list of list of int
-            list of qubit indices corresponding to each unitary in U_list
-
-        Returns
-        -------
-        U: Qobj
-            resultant unitary
+        Returns:
+            `qutip.Qobj`: _description_
```
hodgestar, line 366:
```suggestion
            `qutip.Qobj`: The current state of the simulator.
```

src/qutip_qip/circuit/circuitsimulator.py
```diff
@@ -559,38 +453,43 @@ def _decimal_to_binary(decimal, length):
...
-            self._evolve_state(op)
+            for i in range(len(operation.classical_controls)):
+                cbit_index = operation.classical_controls[i]
+                control_value = cbits_conditions[i]
+                matched[i] = cbits[cbit_index] == control_value
+            return all(matched)
 
-        self.op_index += 1
-        return self.state
+        op = self.qc.gates[self._op_index]
```
hodgestar, line 469:
```suggestion
        op = self._qc.gates[self._op_index]
```
Not sure if this is definitely better, but maybe nice to avoid the property lookup.

src/qutip_qip/circuit/circuitsimulator.py
```diff
@@ -559,38 +453,43 @@ def _decimal_to_binary(decimal, length):
...
+                    operation, self.cbits
+                )
+            else:
+                apply_gate = True
+            if not apply_gate:
+                return current_state
+            if self.mode == "state_vector_simulator":
+                state = self._evolve_state_einsum(op, current_state)
+            else:
+                state = self._evolve_state(operation, current_state)
```
hodgestar, line 488:
```suggestion
            if op.classical_controls is not None:
                apply_gate = _check_classical_control_value(
                    op, self.cbits
                )
            else:
                apply_gate = True
            if not apply_gate:
                return current_state
            if self.mode == "state_vector_simulator":
                state = self._evolve_state_einsum(op, current_state)
            else:
                state = self._evolve_state(op, current_state)
```
I didn't find `operation` to be much clearer and mixing both seemed a bit odd.

src/qutip_qip/circuit/circuitsimulator.py
```diff
@@ -599,41 +498,61 @@ def _evolve_state(self, U):
...
+            # tensor form.
+            state = state.reshape(self._tensor_dims)
+        # Prepare the gate tensor.
+        gate = self.qc._get_gate_unitary(gate)
+        gate_array = gate.full().reshape(gate.dims[0] + gate.dims[1])
+        # Compute the tensor indices and call einsum.
+        num_site = len(state.shape)
+        ancillary_indices = range(num_site, num_site + len(targets_indices))
+        index_list = range(num_site)
+        new_index_list = list(index_list).copy()
```
hodgestar, line 543:
```suggestion
        new_index_list = list(index_list)
```
Since `index_list` is a range, `list(index_list)` is already a new list.

src/qutip_qip/circuit/circuitsimulator.py
```diff
@@ -599,41 +498,61 @@ def _evolve_state(self, U):
...
+        for j, k in enumerate(targets_indices):
+            new_index_list[k] = j + num_site
+        state = np.einsum(
+            gate_array,
```
hodgestar, line 547: Question for the future -- could one in theory provide multiple gates at once here? Would that potentially be better / faster? Possibly with or without use of `np.einsum_path` to select an optimal contraction path?

hodgestar (commented):
Could we also add tests (or maybe just parameterize some existing ones) for both the einsum and non-einsum modes?

src/qutip_qip/circuit/circuitsimulator.py
BoxiLi: Yes, that could be faster if those gates are mostly local. Then it is not that easy to decide unless one first builds the directed graph of the circuit, because gates can be commuting. So instead of this, we should rather try to get tensor network to work and let it does the heavy lifting.

BoxiLi:
Could we also add tests (or maybe just parameterize some existing ones) for both the einsum and non-einsum modes?

@Simon there is no non-einsum mode anymore. There is no need to keep that option. Measurement is not yet compatible with einsum, so in that case it just fails back to normal Qobj.

PR #226, opened 2023-12-18
Remove the unnecessary use of circuit.propagatos

BoxiLi:
`QubitCircuit.compute_unitary()` is faster after the refactoring and saves memory.


hodgestar (approved):
I'm also wondering if we should add a couple of circuit benchmarks to the qutip-benchmarks repository?

BoxiLi:
Yes, that would be nice, We haven't been focused that much on circuit simulation efficiency so far. But if there are people eager to create it, it would surely be awesome.

PR #227, opened 2023-12-18
Use CSR as default for expand_operator

BoxiLi:
Transform the qobj to CSR only if it is Dense, in case more data types are introduced in the future.

### DIFF
--- a/src/qutip_qip/operations/gates.py
+++ b/src/qutip_qip/operations/gates.py
@@ -1226,7 +1226,7 @@ def expand_operator(
         The indices of subspace that are acted on.
         Permutation can also be realized by changing the orders of the indices.
     N : int
-         Deprecated. Number of qubits. Please use `dims`.
+        Deprecated. Number of qubits. Please use `dims`.
     cyclic_permutation : boolean, optional
         Deprecated.
         Expand for all cyclic permutation of the targets.
@@ -1281,11 +1281,9 @@ def expand_operator(
      [0. 0. 0. 1. 0. 0. 0. 0.]]
     """
     if parse_version(qutip.__version__) >= parse_version("5.dev"):
-        if dtype is None:
-            if isinstance(oper.data, qutip.data.Dense):
-                oper = oper.to("csr")
-        else:
-            oper = oper.to(dtype)
+        # If no data type specified, use CSR
+        dtype = dtype or qutip.settings.core["default_dtype"] or qutip.data.CSR
+        oper = oper.to(dtype)
     if N is not None:
         warnings.warn(
             "The function expand_operator has been generalized to "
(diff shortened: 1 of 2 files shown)


hodgestar (approved):
Could we add a small test for this? E.g. check that a Dense operator really is expanded to a CSR? Otherwise looks good.

src/qutip_qip/operations/gates.py
```diff
@@ -1232,6 +1233,9 @@ def expand_operator(
         E.g. if ``N=3`` and `oper` is a 2-qubit operator,
         the result will be a list of three operators,
         each acting on qubits 0 and 1, 1 and 2, 2 and 0.
+    dtype : str, optional
+        Data type of the output `Qobj`. Only for qutip version larger than 5.
```
hodgestar, line 1237: In https://github.com/qutip/qutip/pull/2321 the QuTiP creation operators are currently being updated to ensure that if a `dtype` is specified, the returned Qobj has the requested type, even if the dispatch logic picks something else, because, e.g., a data layer is missing its own implementation of some operation.

It would be nice to be consistent here.

The mentioned PR starts such functions with `dtype = dtype or settings.core["default_dtype"] or _data.CSR` and maybe that is good for this case too? It would mean that the dtype is always CSR unless specified, but perhaps that is better than the current more complex behaviour? Thoughts?

hodgestar (commented):
Thanks for adding the test.

Why were the qiskit tests removed?

src/qutip_qip/operations/gates.py
BoxiLi: Oh thanks, this looks good, especially using the `settings.core["default_dtype"]`.

The point here is to use sparse as much as possible when the input is `Dense`. But it does make sense that if a default data type is specified, one should use that.

BoxiLi:
Deleting `qiskit_test` was a mistake

BoxiLi:
And maybe I should make the same function in qutip-v5 should also be updated accordingly.
https://github.com/qutip/qutip/blob/fcb4117c25e22fe2f553a3bd6d0eda6458f6ff12/qutip/core/tensor.py#L416

hodgestar:
> And maybe I should make the same change to the function in qutip-v5 . https://github.com/qutip/qutip/blob/fcb4117c25e22fe2f553a3bd6d0eda6458f6ff12/qutip/core/tensor.py#L416

I would say yes. Ideally the two functions should remain the same.

BoxiLi:
@Ericgig Does the change look good to you? I'll make the same PR to the main qutip

Ericgig:
I am fine with setting csr as the default type to `expand_operator`.

PR #231, opened 2024-04-02
Drop support of Python=3.7 and 3.8

BoxiLi:
The minimum supported Python version is now 3.9


BoxiLi:
Will merge as is, the filling tests are due to qiskit (and its dependencies), and will be fixed in a different PR.

PR #237, opened 2024-05-14
Add phi parameter to expand_operator call

nwlambert:
Just fixing a very small bug.  The optional phi parameter was not sent to the iterative call to the molmer_sorensen gate when expand_operator is used.


hodgestar (approved):
Should probably have a test but the expansion is deprecated in any case.

hodgestar:
The CI error from MacOS latest, Python 3.9 is unrelated:
```
ERROR: Could not find a version that satisfies the requirement pyqir-generator==0.6.2 (from versions: 0.7.0)
ERROR: No matching distribution found for pyqir-generator==0.6.2
```

PR #238, opened 2024-06-14
Remove QIR

BoxiLi:
QIR hasn't been updated for a while which leads to test failing, besides the supported QIR version os out of date.


BoxiLi:
The failing tests are irrelevant.

PR #239, opened 2024-06-17
matplotlib circuit rendering module

gadhvirushiraj:
#### Major Task Outline

- [x] Single Qubit Gate Rendering
- [x] Multi-Qubit Gate Rendering
- [x] Custom Gate Rendering
- [x] Dynamic Figure Dimension
- [x] Gate Customization Options
- [x] Circuit-Plot Customization Options
- [x]  Color Themes
- [ ] Jupyter Notebook Tutorials

####  Bugs to solve:

- [x] fix text-tabeling placement issue for wires
- [ ] fix line-widths for larger figures

#### Features to add:

- [x] add arg-val option as gate label text
- [x] add multiline gate labeling support
- [x] add more color themes

### DIFF
--- a/src/qutip_qip/circuit/circuit.py
+++ b/src/qutip_qip/circuit/circuit.py
@@ -975,6 +975,7 @@ class QubitCircuit:
         dpi=None,
         file_name="exported_pic",
         file_path="",
+        **kwargs,
     ):
         """
         Export circuit object as an image file in a supported format.
@@ -1014,7 +1015,8 @@ class QubitCircuit:
                 f.write(image_data)
         elif renderer == "matplotlib":
             from .mat_renderer import MatRenderer
-            mat = MatRenderer(self)
+
+            mat = MatRenderer(self, **kwargs)
             mat.canvas_plot()
         else:
             raise ValueError(
--- a/src/qutip_qip/circuit/circuit.py
+++ b/src/qutip_qip/circuit/circuit.py
@@ -200,6 +200,7 @@ class QubitCircuit:
         classical_controls=None,
         control_value=None,
         classical_control_value=None,
+        style=None,
     ):
         """
         Adds a gate with specified parameters to the circuit.
@@ -245,6 +246,7 @@ class QubitCircuit:
                 classical_controls=classical_controls,
                 control_value=control_value,
                 classical_control_value=classical_control_value,
+                style=style,
             )
 
         if index is None:
@@ -863,7 +865,7 @@ class QubitCircuit:
         ignore_measurement: bool, optional
             Whether :class:`.Measurement` operators should be ignored.
             If set False, it will raise an error
-            when the circuit has measurement.
+            when the circuit hasf measurement.
 
         Returns
         -------
@@ -885,16 +887,9 @@ class QubitCircuit:
             )
         for gate in gates:
             if gate.name == "GLOBALPHASE":
-                qobj = gate.get_qobj(self.N)
-            else:
-                qobj = self._get_gate_unitary(gate)
-                if expand:
-                    all_targets = gate.get_all_qubits()
-                    qobj = expand_operator(
-                        qobj, dims=self.dims, targets=all_targets
-                    )
-            U_list.append(qobj)
-        return U_list
+                qobj = gate.get_ (self.N)
+                U_list.append(qobj)
+                continue
 
     def _get_gate_unitary(self, gate):
         if gate.name in self.user_gates:
--- a/src/qutip_qip/circuit/circuit.py
+++ b/src/qutip_qip/circuit/circuit.py
@@ -970,7 +970,7 @@ class QubitCircuit:
 
     def draw(
         self,
-        renderer="latex",
+        renderer="matplotlib",
         file_type="png",
         dpi=None,
         file_name="exported_pic",
@@ -982,7 +982,7 @@ class QubitCircuit:
         Parameters
         ----------
         renderer : choose the renderer for the circuit.
-                   Default: 'latex'
+                   Default: 'matplotlib'
 
         file_type : Provide a supported image file_type eg: "svg"/"png".
                    Default : "png".
@@ -1012,6 +1012,10 @@ class QubitCircuit:
                 os.path.join(file_path, file_name + "." + file_type), mode
             ) as f:
                 f.write(image_data)
+        elif renderer == "matplotlib":
+            from .mat_renderer import MatRenderer
+            mat = MatRenderer(self)
+            mat.canvas_plot()
         else:
             raise ValueError(
                 f"Unknown renderer '{renderer}' not supported. Please choose from 'latex', 'matplotlib', 'text'."
--- a/src/qutip_qip/circuit/circuit.py
+++ b/src/qutip_qip/circuit/circuit.py
@@ -865,7 +865,7 @@ class QubitCircuit:
         ignore_measurement: bool, optional
             Whether :class:`.Measurement` operators should be ignored.
             If set False, it will raise an error
-            when the circuit hasf measurement.
+            when the circuit has measurement.
 
         Returns
         -------
@@ -887,9 +887,16 @@ class QubitCircuit:
             )
         for gate in gates:
             if gate.name == "GLOBALPHASE":
-                qobj = gate.get_ (self.N)
-                U_list.append(qobj)
-                continue
+                qobj = gate.get_qobj(self.N)
+            else:
+                qobj = self._get_gate_unitary(gate)
+                if expand:
+                    all_targets = gate.get_all_qubits()
+                    qobj = expand_operator(
+                        qobj, dims=self.dims, targets=all_targets
+                    )
+            U_list.append(qobj)
+        return U_list
 
     def _get_gate_unitary(self, gate):
         if gate.name in self.user_gates:
--- a/src/qutip_qip/circuit/mat_renderer.py
+++ b/src/qutip_qip/circuit/mat_renderer.py
@@ -2,7 +2,7 @@
 Module for rendering a quantum circuit using matplotlib library.
 """
 
-from typing import List
+from typing import List, Union
 
 import numpy as np
 import matplotlib.pyplot as plt
@@ -31,29 +31,14 @@ class MatRenderer:
     qc : QuantumCircuit Object
         The quantum circuit to be rendered.
 
-    padding : float, optional
-        The padding around the circuit. The default is 0.3.
-
-    dpi : int, optional
-        The resolution of the image. The default is 150.
-
-    bgcolor : str, optional
-        The background color of the plotted circuit. The default is "#FFFFFF" (white).
-
-    wire_labels : list, optional
-        The labels for the wires.
+    style : dict, optional
+        The style dictionary for the circuit. The default is None.
     """
 
     def __init__(
         self,
         qc: QubitCircuit,
-        padding=0.3,
-        dpi: int = 150,
-        bgcolor: str = "#FFFFFF",
-        wire_label: List[str] = None,
-        condense=0.15,
-        bulge=True,
-        end_wire_ext=2,
+        style: dict = None,
     ) -> None:
 
         self.wire_sep = 0.5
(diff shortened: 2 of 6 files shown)


src/qutip_qip/circuit/circuit.py
```diff
@@ -857,7 +859,7 @@ def propagators(self, expand=True, ignore_measurement=False):
         ignore_measurement: bool, optional
             Whether :class:`.Measurement` operators should be ignored.
             If set False, it will raise an error
-            when the circuit has measurement.
+            when the circuit hasf measurement.
```
BoxiLi, line 862: ?
```suggestion
            when the circuit has measurement.
```

src/qutip_qip/circuit/circuit.py
```diff
@@ -879,16 +881,9 @@ def propagators(self, expand=True, ignore_measurement=False):
...
-                if expand:
-                    all_targets = gate.get_all_qubits()
-                    qobj = expand_operator(
-                        qobj, dims=self.dims, targets=all_targets
-                    )
-            U_list.append(qobj)
-        return U_list
+                qobj = gate.get_ (self.N)
+                U_list.append(qobj)
+                continue
```
BoxiLi, line 886: I guess this is only temporary for debug?

src/qutip_qip/operations/gateclass.py
```diff
@@ -153,17 +158,45 @@ def __init__(
...
+                (
+                    default_theme[self.name]
+                    if self.name in default_theme
+                    else "#000000"
+                ),
+            )
+            self.fontsize = style.get("fontsize", 10)
+            self.fontcolor = style.get("fontcolor", "#FFFFFF")
+            self.fontweight = style.get("fontweight", "normal")
+            self.fontstyle = style.get("fontstyle", "normal")
```
BoxiLi, line 185: It is best to move this part also to the plotting files, not in the gate files. The default type could be simply `None`, and all the default settings will be handled when plotting the circuit.

This avoids some complicated cross-importation. `default_theme` only needs to be imported in the plotting code.

src/qutip_qip/circuit/mat_renderer.py
```diff
@@ -0,0 +1,863 @@
...
+
+                self.merged_qubits = gate.targets.copy()
+                if isinstance(gate, Gate) and gate.controls is not None:
+                    self.merged_qubits += gate.controls.copy()
+                self.merged_qubits.sort()
+
+                # check if the gate is a measurement gate
+                if isinstance(gate, Measurement):
+
+                    find_layer = [
```
BoxiLi, line 817: Why is this list named `find_layer`? Is this a list of the wires that this gate includes?

src/qutip_qip/operations/gateclass.py
```diff
@@ -153,17 +158,45 @@ def __init__(
...
+            self.color = style.get(
+                "color",
+                (
+                    default_theme[self.name]
+                    if self.name in default_theme
+                    else "#000000"
+                ),
+            )
+            self.fontsize = style.get("fontsize", 10)
+            self.fontcolor = style.get("fontcolor", "#FFFFFF")
```
BoxiLi, line 183: It would be nice to move this part to the plotting code, not in the gate files. The default input of style could just be `None`, and all the default settings will be added when plotting. 

This avoids importing the `default_theme` and also avoids having too many attributes in the gate class.

BoxiLi (commented):
Thanks for the draft! Left a few comments. 

One small point, we can also start to replace the default latex plotting with this. And  during this process we will also figure out what the default API should be.

src/qutip_qip/circuit/circuit.py
gadhvirushiraj: i think some rebase issue, fixed

src/qutip_qip/operations/gateclass.py
gadhvirushiraj: Agreed, we can move it to `circuit.py` once the plotting API is finalized. For now, I've temporarily moved it to `mat_renderer.py`.

src/qutip_qip/circuit/mat_renderer.py
gadhvirushiraj: I should rename that. It is a list of layers up to which the plotting is done in the wire; for all the wires that the new gate is acting on.

src/qutip_qip/operations/gateclass.py
BoxiLi: Actually `mat_rendere.py` is good. We should avoid adding things to `circuit.py`. In fact, I hope that we can move the full plotting functions out of `circuit.py`.

BoxiLi:
BTW, just noticed this, Notebook Tutorials should not be included in this repo. It should be in the `qutip-tutorial` repo.

src/qutip_qip/circuit/mat_renderer.py
```diff
@@ -0,0 +1,936 @@
...
+        self.gate_width = 0.2
+        self.gate_pad = 0.05
+        self.label_pad = 0.2
+        self.font_size = 10
+        self.default_layers = 2
+        self.arrow_lenght = 0.06
+        self.connector_r = 0.01
+        self.target_node_r = 0.12
+        self.display_layer_len = 0
+        self.start_pad = 0.1
```
BoxiLi, line 71: Would it be cleaner to separate the properties about plotting configuration into a separate dict or data class?
Those are the properties that we allow the use to play with.
This will distinguish them from the other properties like `self.qc`, `self.qwires`, which I believe is should not be changed by the user.

src/qutip_qip/circuit/mat_renderer.py
```diff
@@ -0,0 +1,942 @@
...
+
+        # user defined style
+        style = {} if style is None else style
+        if style.get("bulge", True):
+            self.bulge = "round4"
+        else:
+            self.bulge = "square"
+        self.dpi = style.get("dpi", 150)
```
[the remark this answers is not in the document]
gadhvirushiraj: `self.gate_height` represents the minimum size of the gate, I should rename it. Some of the properties above are just const. and other properties like the radius of the control node `control_node_radius` can be made customizable,  but I think providing this level of customization would be excessive.

src/qutip_qip/circuit/mat_renderer.py
```diff
@@ -0,0 +1,942 @@
...
+        self.layer_sep = style.get("layer_sep", 0.5)
+        self.gate_pad = style.get("gate_pad", 0.05)
+        self.label_pad = style.get("label_pad", 0.1)
+        self.bgcolor = style.get("bgcolor", "#FFFFFF")
+        self.wire_label = style.get("wire_label", None)
+        self.end_wire_ext = style.get("end_wire_ext", 2)
+        self.align_layer = style.get("align_layer", False)
+        self.title = style.get("title", None)
+        self.fig_height = style.get("fig_height", None)
+        self.fig_width = style.get("fig_width", None)
```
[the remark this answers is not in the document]
gadhvirushiraj: Agreed, that would make it much cleaner.  😄

src/qutip_qip/circuit/mat_renderer.py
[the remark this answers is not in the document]
BoxiLi: What about rename those attributes as private `self._min_gate_height`

src/qutip_qip/circuit/mat_renderer.py
```diff
@@ -0,0 +1,1025 @@
...
+        self._ax.add_artist(text_obj)
+
+        bbox = text_obj.get_window_extent(
+            renderer=self._ax.figure.canvas.get_renderer()
+        )
+        inv = self._ax.transData.inverted()
+        bbox_data = bbox.transformed(inv)
+        text_obj.remove()
+
+        return bbox_data.width * 2.54 * 3
```
BoxiLi, line 224: What is this 2.54 * 3, what does this magical number mean?

src/qutip_qip/circuit/mat_renderer.py
```diff
@@ -0,0 +1,1025 @@
...
+                    < gate_width + self.style.gate_margin * 2
+                ):
+                    self._layer_list[wire][layer] = (
+                        gate_width + self.style.gate_margin * 2
+                    )
+            else:
+                temp = xskip - sum(self._layer_list[wire]) if xskip != 0 else 0
+                self._layer_list[wire].append(
+                    temp + gate_width + self.style.gate_margin * 2
+                )
```
BoxiLi, line 264: Maybe leave a few words just briefly commenting on what those if-else branch corresponds to, i.e., the current layer is full / if not full what will happen.

src/qutip_qip/circuit/mat_renderer.py
```diff
@@ -0,0 +1,1025 @@
...
+                xskip + self.style.gate_margin + self.style.gate_pad,
+                pos * self.style.wire_sep,
+            ),
+            self._control_node_r,
+            color=color,
+            zorder=self._zorder["node"],
+        )
+        self._ax.add_artist(control_node)
+
+    def _draw_target_node(self, pos: int, xskip: float, color: str) -> None:
```
BoxiLi, line 375: I understand that these target_node and control_node are for CNOT gate only?

src/qutip_qip/circuit/mat_renderer.py
```diff
@@ -0,0 +1,1025 @@
...
+            wire_down = plt.Line2D(
+                [0, max_len],
+                [
+                    (i + self._qwires) * self.style.wire_sep,
+                    (i + self._qwires) * self.style.wire_sep,
+                ],
+                lw=1,
+                color=self.style.wire_color,
+                zorder=self._zorder["wire"],
+            )
```
BoxiLi, line 305: What is the difference between this part and the `_draw_qbridge` and `_draw_cbridge`?

src/qutip_qip/circuit/mat_renderer.py
```diff
@@ -0,0 +1,1025 @@
...
+        layer : int
+            The layer the gate is acting on.
+        """
+
+        wire_list = list(
+            range(self.merged_qubits[0], self.merged_qubits[-1] + 1)
+        )
+        com_xskip = self._get_xskip(wire_list, layer)
+
+        if gate.name == "CNOT":
```
BoxiLi, line 719: Add also the name "CX" here, they are the same gate

src/qutip_qip/circuit/mat_renderer.py
```diff
@@ -0,0 +1,1025 @@
...
+            self._draw_control_node(gate.controls[1], com_xskip, self.color)
+            self._draw_target_node(gate.targets[0], com_xskip, self.color)
+            self._draw_qbridge(
+                gate.targets[0], gate.controls[0], com_xskip, self.color
+            )
+            self._draw_qbridge(
+                gate.targets[0], gate.controls[1], com_xskip, self.color
+            )
+            self._manage_layers(
+                2 * self.style.gate_pad + self._target_node_r / 3,
```
BoxiLi, line 756: It seems that all the calling of _manage_layers takes `2 * self.style.gate_pad + self._target_node_r / 3`, Can we somehow define this as a separate function to reduce the duplication and make it easier to add new gates?
Something like a `_gate_pad_to_gate_width_map`?

BoxiLi (changes requested):

src/qutip_qip/circuit/mat_renderer.py
```diff
@@ -0,0 +1,1025 @@
...
+            else self.style.fig_height
+        )
+        if self._ax is None:
+            self.fig, self._ax = plt.subplots(
+                figsize=(self.style.fig_width, self.style.fig_height),
+                dpi=self.style.dpi,
+                facecolor=self.style.bgcolor,
+            )
+
+        self.canvas_plot()
```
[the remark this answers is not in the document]
gadhvirushiraj: This was a temporary, `canvas_plot` was intended to be called directly by the `draw` api. Now that the api PR has been merged, I'll make the change.

src/qutip_qip/circuit/mat_renderer.py
gadhvirushiraj: This part refers to rendering the classical wire horizontally, where as `_draw_cbridge` and `_draw_qbridge` are to renderer classical and qubit across the two wires. Example, `_draw_cbridge` is used to render the connection from a mesurment gate to classical line and  `_draw_qbridge` are used for connecting control node to gates.

src/qutip_qip/circuit/mat_renderer.py
gadhvirushiraj: `_draw_control_node` is also utilized by some multi-qubit gates such as FREDKIN. Additionally,  if a user defines custom gates with a control node, `_draw_control_node` will be used. `_draw_target_node` is also used by the TOFFOLI gate.

src/qutip_qip/circuit/mat_renderer.py
gadhvirushiraj: `bbox_data.width` returns width in inches, 2.54 is for conversion to centimeters and 3 is multiplied to bring it in same range as of the defined const 😄

src/qutip_qip/circuit/mat_renderer.py
gadhvirushiraj: As of now only two calling take `2 * self.style.gate_pad + self._target_node_r / 3` as argument (CNOT and TOFFOLI). Calling inside `_draw_singleq_gate`, `SWAP` and  general `_draw_multiq_gate` are different.

BoxiLi:
Need any more feedback from me? I believe the PR is ready if the suggestions are addressed.

BoxiLi:
@gadhvirushiraj Any updates? Would be nice to get it merged before the next meeting?

BoxiLi:
The code climate one is no big issue.

BoxiLi:
Just let me know by asking for a review when you are done.

src/qutip_qip/circuit/mat_renderer.py
```diff
@@ -0,0 +1,1025 @@
...
+        """
+
+        xskip = []
+        if self.style.align_layer:
+            wire_list = list(range(self._qwires))
+
+        for wire in wire_list:
+            xskip.append(sum(self._layer_list[wire][:layer]))
+
+        return max(xskip)
```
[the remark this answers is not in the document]
gadhvirushiraj: I can't come up with a good solution for this other than adding if statements when adding gates, which I think would be a redundant approach. So, resolving conversation for now.

BoxiLi:
The coverage is failing because we haven't added tests. We need to add some later. Copy some from the notebook for example.

BoxiLi (approved):

PR #241, opened 2024-07-09
Shifting LaTeX Code and Draw API Refactor

gadhvirushiraj:
## Shifting LaTeX Code and Circuit API Refactor

Transferring all LaTeX circuit rendering functionality from `circuit.py` to a dedicated file named `texrenderer.py`. Additional renderers such as `matplotlib` and `text` will be incorporated into the `draw()` method as they become available.


gadhvirushiraj:
@BoxiLi  should we move `circuit_latex.py` as well, or maybe rename it to `latex_utils.py` ?

BoxiLi:
I think it would be nice to merge them into `texrenderer.py`. One file suffice, it is not so big.

BoxiLi:
I activated the tests but some of them are failing I believe. You can verify by `pytest tests` locally on your machine.

gadhvirushiraj:
@BoxiLi All tests are passing, and `circuit_latex.py` has been moved. Do we need to include any particular deprecation message, considering `qc.draw()` continues to function as it did before?

src/qutip_qip/circuit/circuit.py
```diff
@@ -1119,17 +997,19 @@ def draw(
...
+            else:
+                mode = "wb"
+                if file_type == "png" and not dpi:
+                    dpi = 100
+            latex = TeXRenderer(self)
+            image_data = latex.raw_img(file_type, dpi)
+            with open(
+                os.path.join(file_path, file_name + "." + file_type), mode
+            ) as f:
+                f.write(image_data)
```
BoxiLi, line 1012: Please add a else statement here and an raise an error saying something like `f"Unknown {renderer}, please choose from "latex_render", ... and ...` etc. Right now there is only latex rendering of course.

BoxiLi (changes requested):
Left one comment, will merge after this is addressed.

src/qutip_qip/circuit/circuit.py
gadhvirushiraj: Added the error message and kept the options simple: latex, matplotlib, and text for easy recall.

BoxiLi (approved):

PR #244, opened 2024-08-08
Text based Circuit Rendering Module 

gadhvirushiraj:
#### Major Task Outline

- [x] Single Qubit Gate Rendering
- [x] Multi-Qubit Gate Rendering
- [x] Custom Gate Rendering
- [x] Measurement Rendering 
- [ ] Customization Options
- [ ] Jupyter Notebook Tutorials


tests/test_textrenderer.py
```diff
@@ -0,0 +1,83 @@
...
+        if isinstance(gate, Gate):
+            if len(gate.targets) == 1 and gate.controls is None:
+                parts, _ = tr._draw_singleq_gate(gate.name)
+            else:
+                parts, _ = tr._draw_multiq_gate(gate)
+        elif isinstance(gate, Measurement):
+            parts, _ = tr._draw_measurement_gate(gate)
+
+        len_parts = [len(part) for part in parts]
+        assert (
```
BoxiLi, line 66: Add a comment that this is to test all the `len_parts` are the same.

src/qutip_qip/circuit/text_renderer.py
```diff
@@ -0,0 +1,506 @@
...
+
+class TextRenderer:
+    """
+    A class to render a quantum circuit in text format.
+    """
+
+    def __init__(self, qc: QubitCircuit):
+
+        self._qc = qc
+        self._qwires = qc.N
```
BoxiLi, line 23: A design question, technically speaking, the Renderer class should be independent of the quantum circuit provided.

Therefore, `qc` should better be an input argument for some `canvas_plot` for `draw` methods. Not as part of the initialization of the renderer class.

BoxiLi:
There is one thing that I forgot last time, the public API should be added to the apidoc https://github.com/qutip/qutip-qip/tree/master/doc/source/apidoc

src/qutip_qip/circuit/mat_renderer.py
```diff
@@ -15,130 +15,16 @@
...
-        Controls color of acsent elements (eg. cross sign in the target node)
-        and set as deafult color of gate-label. Can be overwritten by gate specific color.
-        The default is None.
-
-    wire_label : Optional[List], optional
-        The labels of the wires. The default is None.
-
-    wire_color : Optional[str], optional
-        The color of the wires. The default is None.
-    """
```
BoxiLi, line 93: How do I know which are available for the Textrender and which are for the Matrenderer? Is there a way to document it or make it clearer?

src/qutip_qip/circuit/text_renderer.py
```diff
@@ -0,0 +1,497 @@
...
+            elif wire == wire_list[0]:
+                self._render_strs["top_frame"][wire] += bar_conn
+                self._render_strs["mid_frame"][wire] += cross_conn
+                self._render_strs["bot_frame"][wire] += " " * len(bar_conn)
+            else:
+                self._render_strs["top_frame"][wire] += bar_conn
+                self._render_strs["mid_frame"][wire] += mid_bar_conn
+                self._render_strs["bot_frame"][wire] += bar_conn
+
+    def layout(self):
```
BoxiLi, line 382: What is the difference between `layout` and `print_circuit`?

BoxiLi (commented):
This is just a first quick review. A more detailed one will follow.

You can also add the test for matplotlib here, should be fine. There should also be one test for the circuit.draw() interface, making sure that the API is working properly.

The tex renderer should also be renamed to match the convention.

src/qutip_qip/circuit/mat_renderer.py
BoxiLi: These options should probably be summarised or somehow demonstrated to the user either in the user guide or in the doc string referred to in the user guide. Especially how they can be defined via the `circuit.draw` interface, as dict? Or as class?

src/qutip_qip/circuit/text_renderer.py
```diff
@@ -0,0 +1,497 @@
...
+            self._render_strs["mid_frame"][i] = (
+                f" {label} :" + self._render_strs["mid_frame"][i]
+            )
+            self._render_strs["top_frame"][i] = " " * (len(f" {label} :") + 2)
+            self._render_strs["bot_frame"][i] = " " * (len(f" {label} :") + 2)
+
+        self._layer_list = {
+            i: [len(self._render_strs["mid_frame"][i])]
+            for i in range(self._qwires + self._cwires)
+        }
```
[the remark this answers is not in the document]
BoxiLi: What does `_layer_list` actually save? I assume `_layer_list[i]` is a list of length for each element for a wire? Maybe document it somewhere.i

src/qutip_qip/circuit/text_renderer.py
```diff
@@ -0,0 +1,497 @@
...
+        wire_len = [sum(wire) for wire in self._layer_list.values()]
+        max_len = max(wire_len) + self.style.end_wire_ext
+
+        for i, length in enumerate(wire_len):
+            if length < max_len:
+                diff = max_len - length
+                if i < self._qwires:
+                    self._render_strs["mid_frame"][i] += "─" * diff
+                else:
+                    self._render_strs["mid_frame"][i] += "═" * diff
```
BoxiLi, line 96: What is the difference between `_extend_line` and `_adjust_layer`. Why does `_extend_line` not add space to the top and bottom? I feel like `_extend_line` can be replaced by `adjust_layer` by a proper definition of `xskip`

src/qutip_qip/circuit/text_renderer.py
```diff
@@ -0,0 +1,497 @@
...
+        Returns
+        -------
+        tuple
+            The parts of the gate to be drawn. The parts are the top_frame, mid_frame, and bot_frame.
+
+        int
+            The width of the gate.
+        """
+
+        lid_seg = "─" * (ceil(self.style.gate_pad) * 2 + len(gate_name))
```
BoxiLi, line 116: From the default Style, `gate_pad=0.05`, however here I believe we need an integer here? `ceil` is a bit weird. It will cause weird behaviour. I think we should just have a different default value. It seems that it would be nice to summarize the default value somewhere. This is related to the other comment of mine yesterday. Maybe have a separate default style for each renderer?

src/qutip_qip/circuit/text_renderer.py
```diff
@@ -0,0 +1,497 @@
...
+        mid_connect = f"─┤{pad}{' ' * len(gate_text)}{pad}├─"
+        mid_connect_label = f"─┤{pad}{gate_text}{pad}├─"
+
+        # Adjust top_frame or bottom if there is a control wire
+        if gate.controls:
+            mid_index = len(bot_frame) // 2
+            if gate.controls[0] < gate.targets[0]:
+                bot_frame = (
+                    bot_frame[:mid_index] + "┬" + bot_frame[mid_index + 1 :]
+                )
```
BoxiLi, line 159: I think you  can just use `bot_frame[:mid_index]="┬"`

src/qutip_qip/circuit/base_renderer.py
```diff
@@ -0,0 +1,212 @@
...
+        """
+
+        if self.style.align_layer:
+            wire_list = list(range(self._qwires))
+
+        xskip = []
+        for wire in wire_list:
+            xskip.append(sum(self._layer_list[wire][:layer]))
+
+        return max(xskip)
```
BoxiLi, line 168: Does `aligh_layer` also work for the text renderer?

src/qutip_qip/circuit/text_renderer.py
```diff
@@ -0,0 +1,497 @@
...
+
+        width : int
+            The width of the gate.
+        """
+
+        bar_conn = " " * (width // 2) + "║" + " " * (width // 2)
+        mid_bar_conn = "─" * (width // 2) + "║" + "─" * (width // 2)
+        mid_bar_classical_conn = "═" * (width // 2) + "║" + "═" * (width // 2)
+        classical_conn = "═" * (width // 2) + "╩" + "═" * (width // 2)
+        self._adjust_layer(wire_list, xskip)
```
BoxiLi, line 216: There is an `_adjust_layer` for all `_update_*` methods except for `_update_singleq`, why?

src/qutip_qip/circuit/text_renderer.py
```diff
@@ -0,0 +1,497 @@
...
+
+            if isinstance(gate, Measurement):
+                self._adjust_layer(wire_list, xskip)
+                self._update_singleq(gate.targets, parts)
+                self._update_cbridge(gate, wire_list, xskip, width)
+            elif len(gate.targets) == 1 and gate.controls is None:
+                self._update_singleq(wire_list, parts)
+            else:
+                sorted_targets = sorted(gate.targets)
+                self._update_target_multiq(
```
BoxiLi, line 428: Is it possible to put one `self._adjust_layer(wire_list, xskip)` in line 419, i.e., before all the `_update_*` methods. And remove all of those inside the `_update_*` methods?

src/qutip_qip/circuit/text_renderer.py
```diff
@@ -0,0 +1,497 @@
...
+
+            # update the render strings for the gate
+            layer = max(len(self._layer_list[i]) for i in wire_list)
+            xskip = self._get_xskip(wire_list, layer)
+            self._manage_layers(width, wire_list, layer, xskip)
+
+            if isinstance(gate, Measurement):
+                self._adjust_layer(wire_list, xskip)
+                self._update_singleq(gate.targets, parts)
+                self._update_cbridge(gate, wire_list, xskip, width)
```
BoxiLi, line 423: For clarity
```python
            elif gate.name == "SWAP":
                self._update_swap_gate(gate)
                continue
```
should be moved to later part, together with other `_update_*` methods.

src/qutip_qip/circuit/text_renderer.py
```diff
@@ -0,0 +1,497 @@
...
+
+    def _update_swap_gate(self, gate: Gate):
+        """
+        Update the render strings for the SWAP gate
+        """
+
+        wire_list = list(range(min(gate.targets), max(gate.targets) + 1))
+        layer = max(len(self._layer_list[i]) for i in wire_list)
+        xskip = self._get_xskip(wire_list, layer)
+        self._adjust_layer(wire_list, xskip)
```
BoxiLi, line 360: This part could be removed once ` _update_swap_gate` is moved to later part in `self.layout`

src/qutip_qip/circuit/text_renderer.py
```diff
@@ -0,0 +1,497 @@
...
+        """
+
+        lid_seg = "─" * (ceil(self.style.gate_pad) * 2 + len(gate_text))
+        pad = " " * ceil(self.style.gate_pad)
+
+        top_frame = f" ┌{lid_seg}┐ "
+        bot_frame = f" └{lid_seg}┘ "
+        mid_frame = f" │{pad}{' ' * len(gate_text)}{pad}│ "
+        mid_connect = f"─┤{pad}{' ' * len(gate_text)}{pad}├─"
+        mid_connect_label = f"─┤{pad}{gate_text}{pad}├─"
```
BoxiLi, line 151: Low priority:
I feel like ` _draw_singleq_gate` can be merged into this `_draw_multiq_gate`.

src/qutip_qip/circuit/base_renderer.py
```diff
@@ -0,0 +1,212 @@
...
+                    # update with new layer width
+                    self._layer_list[wire][layer] = (
+                        gate_width + self.style.gate_margin * 2
+                    )
+            else:
+                # add layer width: new layer width + missing layer widths if exits
+                temp = xskip - sum(self._layer_list[wire]) if xskip != 0 else 0
+                self._layer_list[wire].append(
+                    temp + gate_width + self.style.gate_margin * 2
+                )
```
BoxiLi, line 212: Personally, I think it would be nice to explicitly return the self._layer_list just to make it clear that this has been changed. But it is up to you.

src/qutip_qip/circuit/text_renderer.py
```diff
@@ -0,0 +1,497 @@
...
+        """
+
+        lid_seg = "─" * (ceil(self.style.gate_pad) * 2 + len(gate_name))
+        pad = " " * ceil(self.style.gate_pad)
+
+        top_frame = f" ┌{lid_seg}┐ "
+        mid_frame = f"─┤{pad}{gate_name}{pad}├─"
+        bot_frame = f" └{lid_seg}┘ "
+
+        return (top_frame, mid_frame, bot_frame), len(top_frame)
```
BoxiLi, line 123: Should we add an assert checking that the length of all three should be the same? What do you think? This could even replace the last two tests.

src/qutip_qip/circuit/base_renderer.py
BoxiLi: I guess text rendering is by default aligned.

src/qutip_qip/circuit/text_renderer.py
gadhvirushiraj: The `layout` function iterates through the gates and updates the `_render_strs` list of strings, while `print_circuit` handles displaying the `_render_strs` in the correct order, either in the terminal or within an IPython environment.

src/qutip_qip/circuit/text_renderer.py
```diff
@@ -0,0 +1,497 @@
...
+                xskip - len(self._render_strs["bot_frame"][wire])
+            )
+            if wire < self._qwires:
+                self._render_strs["mid_frame"][wire] += "─" * (
+                    xskip - len(self._render_strs["mid_frame"][wire])
+                )
+            else:
+                self._render_strs["mid_frame"][wire] += "═" * (
+                    xskip - len(self._render_strs["mid_frame"][wire])
+                )
```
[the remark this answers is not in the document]
gadhvirushiraj: Since we're dealing with strings and the gates vary in width, `_adjust_layer` adds padding to maintain consistent layer alignment across the different wires. The function `_add_wire_all` appears to be of different menaing, but to clarify its purpose, renaming it to `_adjust_layer_pad` would make its role clearer.

src/qutip_qip/circuit/text_renderer.py
```diff
@@ -0,0 +1,497 @@
...
+        # user defined style
+        style = {} if style is None else style
+        style["gate_margin"] = 0
+        self.style = StyleConfig(**style)
+
+        super().__init__(self.style)
+        self._qc = qc
+        self._qwires = qc.N
+        self._cwires = qc.num_cbits
+        self._layer_list = {i: [] for i in range(self._qwires + self._cwires)}
```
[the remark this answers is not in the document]
gadhvirushiraj: This was originally used to address the classical wire issue with measurement joint, though it was handled differently. I'll change it to list.

src/qutip_qip/circuit/base_renderer.py
gadhvirushiraj: Yes, it does perfectly 😃

gadhvirushiraj: > I guess text rendering is by default aligned.

It’s not enabled by default—we still need to set the `align_layer=True` argument. This is essential because we’re continuing to use the layer concept to store gates at different horizontal positions, aiming for maximum compactness. The only change is that the minimum shift in the case of text is now 1 unit, rather than a floating-point value.

src/qutip_qip/circuit/text_renderer.py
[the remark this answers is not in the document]
gadhvirushiraj: Dict was used to address the classical wire issue with measurement joint, though it was handled differently. So, its a list.

Currently, it shifts the entire wire by a space for `q10` because of extra digit.  Will make change to add padding to make all label len same.

src/qutip_qip/circuit/text_renderer.py
gadhvirushiraj: - Since we're dealing with strings and the gates vary in width, `_adjust_layer` adds padding to maintain consistent layer alignment across the different wires. 

- `extend_wire` is only used at the end of circuit, after all gates are added to extend wire to certain extra length

gadhvirushiraj: It's a nice idea to reuse the `_adjust_layer` instead of `_extend_wire`, will give it a try.

gadhvirushiraj: Update: Removed `_extend_line`

src/qutip_qip/circuit/text_renderer.py
gadhvirushiraj: Padding via `_adjust_layer` is primarily necessary when a gate interacts with multiple wires, as it ensures consistent padding across all of them.

However, while testing with align_layer=True, I discovered an edge case where additional padding is needed. I'll likely implement this adjustment after each gate is added.

src/qutip_qip/circuit/text_renderer.py
gadhvirushiraj: Sure, the `_draw_singleq_gate` function follows a similar structure to MatRenderer, with separate drawing functions for single and multi-qubit gates.

src/qutip_qip/circuit/text_renderer.py
```diff
@@ -0,0 +1,497 @@
...
+                        wire_list[0] if not is_top_closer else wire_list[-1]
+                    )
+                    self._update_qbridge(
+                        gate,
+                        list(
+                            range(
+                                min(sorted_controls[0], closest_pos),
+                                max(sorted_controls[-1], closest_pos) + 1,
+                            )
+                        ),
```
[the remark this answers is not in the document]
gadhvirushiraj: > "What actually will happen if a gate is controlled by q0 and q2 and target on q1?"

This case is not handled as of now, will add logic to tackle it.

src/qutip_qip/circuit/text_renderer.py
gadhvirushiraj: It can replace one test, `test_parts_len`, and the other test checks if all `_render_str` or wire lengths are consistent. I'm okay with either option; the choice might come down to which provides better coverage, while both options ensure testing. Let me know what you think.

BoxiLi: It won't affect the coverage because the general test will also cover these lines of code I think.

The point is test_parts_len is to find some specific error and make debugging easier. If it is equivalent to adding assert here. I think it is better to have assert here because it helps detect problems early, where the cause is clear, rather than later when some other operation fails.

src/qutip_qip/circuit/text_renderer.py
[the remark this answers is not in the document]
BoxiLi: For some reason it already works ;)

image

doc/source/qip-basics.rst
```diff
@@ -272,31 +272,136 @@ gate function returning a :class:`qutip.Qobj` and save it in the attribute ``use
...
 
-.. image:: /figures//quantum_circuit_example.png
+- **Matplotlib (Default)**:
+
+    .. code-block:: python
+          
+        # plot the quantum circuit using  MatRenderer
+        qc.draw("matplotlib")
+
+    .. image:: /figures//mat_circuit_example.png
```
BoxiLi, line 299: It is better to use the plot derivative than upload the `png` here. We used it for latex because there is really no other way.

https://github.com/qutip/qutip-qip/blob/master/doc/source/qip-processor.rst#spinchain

It needs to be rebased to remove png from the history.

BoxiLi: You can download and view the rendered `html` here in the artifact.

https://github.com/qutip/qutip-qip/actions/runs/10496547795?pr=244

doc/source/qip-basics.rst
```diff
@@ -272,31 +272,136 @@ gate function returning a :class:`qutip.Qobj` and save it in the attribute ``use
...
+    
+               ┌──────┐  ┌──────┐  ┌───────┐  ┌───┐   
+        q1 :───┤ CNOT ├──┤ SNOT ├──┤       ├──┤ M ├───
+               └───┬──┘  └──────┘  │       │  └─╥─┘   
+                   │               │       │    ║     
+        q0 :───────▇───────────────┤ ISWAP ├────║─────
+                                   └───────┘    ║     
+                                                ║     
+        c0 :════════════════════════════════════╩═════
+
```
BoxiLi, line 371: Somehow this line 367 is shifted？

tests/test_matrenderer.py
```diff
@@ -0,0 +1,14 @@
...
+
+@pytest.mark.parametrize("qc_fixture", ["qc1", "qc2", "qc3"])
+def test_matrenderer(request, qc_fixture):
+    qc = request.getfixturevalue(qc_fixture)
+
+    with patch("matplotlib.pyplot.show"):
+        try:
+            qc.draw("matplotlib")
+        except Exception as e:
+            assert False, f"Error: {e}"
```
BoxiLi, line 14: Let's merge the test files since it is just one test you are already import the circuits from the other one.

And there is no need to try and catch because pytest will automatically catch the exceptions

BoxiLi (commented):
Looks good!

Do you think the MatRenderer and TextRenderer and the style class should be exposed to the users? If so, the public API should be added to the apidoc https://github.com/qutip/qutip-qip/blob/master/doc/source/apidoc/qutip_qip.circuit.rst.

BoxiLi:
And this comments of min:


> A design question, technically speaking, the Renderer class should be independent of the quantum circuit provided.
> 
> Therefore, qc should better be an input argument for some canvas_plot for draw methods. Not as part of the initialization of the renderer class.
>

BoxiLi:
We can (and should) replace other latex circuits in the doc with the new matplotlib plotting. But let's not add that to this PR just for clarity.

BoxiLi:
Add the default value of the style configuration to the doc.

gadhvirushiraj:
@BoxiLi 
I've updated the `circuit.py` file. The `QubitCircuit` object now uses matplotlib as the default representation (if matplotlib is available); otherwise, it falls back to text-based rendering. I've also removed the unnecessary LaTeX hooks for Jupyter Notebooks. Currently, I've added a `bool option` to the draw function for saving in the respective format, didn't find  a better approach for integrating the save functionality.

src/qutip_qip/circuit/circuit.py
```diff
@@ -940,89 +947,77 @@ def compute_unitary(self):
...
-            Applicable for PNG, NA for SVG.
-            Default : None, though it's set to 100 internally for PNG
+        dpi : int, optional
+            The image density in dots per inch (dpi). Applicable for PNG, not for SVG.
+            Default is None (set to 100 internally for PNG).
 
-        file_name : Filename of the exported image.
-                    Default : "exported_pic"
+        file_path : str, optional
+            The path to save the image file. Default is the current working directory.
```
BoxiLi: It seems that the default is `circuit` not current working directory

BoxiLi (commented):
Yeah the interface of `circuit.draw` looks good.

src/qutip_qip/circuit/circuit.py
```diff
@@ -940,84 +947,77 @@ def compute_unitary(self):
...
-                    Default : "exported_pic"
+        save : bool, optional
+            If True, the image will be saved to the specified path.
+            Default is False.
 
-        file_path : Path to which the file has to be exported.
-                    Default : ""
-                    Note : User should have write access to the location.
+        kwargs : dict
+            Additional keyword arguments passed to the renderer.
```
BoxiLi, line 984: Need to point to the doc where all the kwargs  are listed.

gadhvirushiraj: All the kwarg arguments are mentioned in `base_renderer.py`'s `StyleConfig Dataclass`, I have added a note.

doc/source/qip-basics.rst
```diff
@@ -272,31 +272,168 @@ gate function returning a :class:`qutip.Qobj` and save it in the attribute ``use
...
+        and sets as default color of gate-label. Can be overwritten 
+        by gate-specific color.
+    * - ``wire_label : Optional[List] = None``
+      - Labels of the wires.
+    * - ``wire_color : Optional[str] = None``
+      - Color of the wires.
+
+- **Text**:
+
+  .. testcode::
```
BoxiLi, line 368: I found the problem @gadhvirushiraj . The indent matters here! `testcode`, `testoutput` and `**Customization Parameters**` should not have indent.

BoxiLi (approved):

BoxiLi:
Sorry for the delay and thank you very much for the fix @gadhvirushiraj ! It looks all good. Merge.

Lets see how the tutorials run!

PR #245, opened 2024-08-14
Add RZX gate to the SCQubits model

BoxiLi:
- Add the RZX gate for the SCQubits. It was the building block for CNOT but is now separated to generate Hamiltonian simulation more conveniently.
- Fix a bug in the gate decomposition, if the gate is added as a native gate, the resolve function lets it pass without an error.
- Fix a bug in the compilation of the ZX Hamiltonian. The wrong hardware parameters were used for the compilation (mismatch in the qubit label.


hodgestar:
It looks like quite a few small details changed here. Could you add tests?

BoxiLi:
Yes! The test is on the way. This is for the example in the qutip paper ;)

src/qutip_qip/circuit/circuit.py
```diff
@@ -588,9 +589,6 @@ def resolve_gates(self, basis=["CNOT", "RX", "RY", "RZ"]):
                     basis_1q.append(gate)
                 else:
                     pass
-                    raise NotImplementedError(
```
hodgestar, line 591: Unrelated to this PR, but there is a lot of strange magic here in the construction of the list of basis operations. I don't know why. Do you perhaps?

src/qutip_qip/compiler/circuitqedcompiler.py
```diff
@@ -226,13 +273,10 @@ def cnot_compiler(self, gate, args):
...
         result += self.gate_compiler[gate1.name](gate1, args)
 
-        zx_coeff = self.params["zx_coeff"][q1]
-        area = 1 / 2
-        coeff, tlist = self.generate_pulse_shape(
-            args["shape"], args["num_samples"], maximum=zx_coeff, area=area
-        )
-        pulse_info = [("zx" + str(q1) + str(q2), coeff)]
-        result += [Instruction(gate, tlist, pulse_info)]
+        from ..operations.gateclass import RZX
```
hodgestar, line 276: Is RZX used somewhere here?

hodgestar (commented):
Thanks for adding the tests. Looks good. I left some questions.

src/qutip_qip/circuit/circuit.py
BoxiLi: Yeah, you are right that there is some complicated logic here.

The goal is to decompose those gates to the native basis gates (like SWAP to CNOT), unless it is already provided in the basis as a native gates, such as `RZX` (e.g. for Hamiltonian simulation).

`basis_2q_valid` and `basis_1q_valid` contain the basis gate list of which decomposition is implemented in qutip. Previously, if `RZX` is provided as a native gate the algorithm raises a not-implemented error, because it does not know how to decompose other gates to `RZX`. But now we want it to just pass the RZX gate on without trying to decompose any other gates (like CNOT) to it.

The overall logic of the decomposition can be drastically simplified if it is refactored, since now we have the gate class. The decomposition scheme can be defined under each gate class, not independently. But this will take some work.

src/qutip_qip/compiler/circuitqedcompiler.py
BoxiLi: Oh yeah you are right I end up just using the Gate class

src/qutip_qip/circuit/circuit.py
hodgestar: Thanks for the great explanation. Would be cool to sort this out, but definitely not for here. :)

hodgestar (approved):
I see there is still a doctest error, but otherwise the changes look good to me.

Issue #253, opened 2024-10-04
Using `add_1q_gate` throws `TypeError`


gadhvirushiraj:
### **Problem Description**
Using `add_1q_gate` for circuits results in a `TypeError`. This function used to work previously.

### **Code to Reproduce the Bug**
```
qc = QubitCircuit(3)
qc.add_gate("H", targets=[1])
qc.add_gate("RZ", targets=[2], arg_value=0.5)
qc.add_1q_gate("RX", 0, arg_value=0.5)
qc.add_gate("SWAP", targets=[0, 2])
```

### **Error Thrown**
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[2], [line 4](vscode-notebook-cell:?execution_count=2&line=4)
      [2](vscode-notebook-cell:?execution_count=2&line=2) qc.add_gate("H", targets=[1])
      [3](vscode-notebook-cell:?execution_count=2&line=3) qc.add_gate("RZ", targets=[2], arg_value=0.5)
----> [4](vscode-notebook-cell:?execution_count=2&line=4) qc.add_1q_gate("RX", 0, arg_value=0.5)
      [5](vscode-notebook-cell:?execution_count=2&line=5) qc.add_gate("SWAP", targets=[0, 2])

File ~/Desktop/GSOC/qutip-qip/src/qutip_qip/circuit/circuit.py:312, in QubitCircuit.add_1q_gate(self, name, start, end, qubits, **kwargs)
    [292](https://file+.vscode-resource.vscode-cdn.net/Users/rushirajgadhvi/Desktop/GSOC/learning/notebooks/~/Desktop/GSOC/qutip-qip/src/qutip_qip/circuit/circuit.py:292) """
    [293](https://file+.vscode-resource.vscode-cdn.net/Users/rushirajgadhvi/Desktop/GSOC/learning/notebooks/~/Desktop/GSOC/qutip-qip/src/qutip_qip/circuit/circuit.py:293) Adds a single qubit gate with specified parameters on a variable
    [294](https://file+.vscode-resource.vscode-cdn.net/Users/rushirajgadhvi/Desktop/GSOC/learning/notebooks/~/Desktop/GSOC/qutip-qip/src/qutip_qip/circuit/circuit.py:294) number of qubits in the circuit. By default, it applies the given gate
   (...)
[...]

PR #254, opened 2024-10-05
Fix a bug in add_1q_gate

BoxiLi:
@gadhvirushiraj 

In your example in #253 you defined `start` but not `end`, that is a bit weird way of using it. I think it is better to use `qubits=` in that case. I write now an explicit error message.

Would you like to review the changes I made?


gadhvirushiraj (approved):
Looks good to me

PR #255, opened 2024-10-19
Add type hints to the circuit module

BoxiLi:
- Add type hints to the circuit module
- Rename `texrenderer.py` to `tex_renderer.py`
- Remove `circuit.adjacent_gate as it is replaced by to_chain_structure` long ago.


hodgestar:
What Python versions do we need to support? If 3.9+ should we do `from __future__ import annotations` and then not have to import so many things from typing? E.g. `Optional[Union[int, List[int]]` becomes `int | list[int] | None`.

gadhvirushiraj:
> What Python versions do we need to support? If 3.9+ should we do `from __future__ import annotations` and then not have to import so many things from typing? E.g. `Optional[Union[int, List[int]]` becomes `int | list[int] | None`.

@boxi, I wanted to check on the python versions currently supported. The new circuit module type hint PR would require adjustments based on this suggestion.

BoxiLi:
We support what qutip supports, and qutip is aligning with NumPy. So it is definitely 3.9+. Do you need even newer version?

gadhvirushiraj:
> We support what qutip supports, and qutip is aligning with NumPy. So it is definitely 3.9+. Do you need even newer version?

thanks for clarifying @BoxiLi. this works good.

PR #256, opened 2024-10-19
Add type hints to the operations module

BoxiLi:
- Add type hints
- Remove the support that expands a gate Qobj to a larger Hilbert space, it has been deprecated in the last minor version.


src/qutip_qip/operations/gateclass.py
```diff
@@ -96,81 +98,21 @@
...
-"""
-.. testsetup::
-
-   import numpy as np
-   np.set_printoptions(5)
-"""
-
 
 class Gate:
-    r"""
```
hodgestar, line 108: Is removal of the docstrings intentional?

PR #257, opened 2024-10-26
Fix Plotting Dimensions and Aspect Ratio Consistency

gadhvirushiraj:
### BUG / ISSUE:
 - Single-qubit circuit plotting faced dimensional inconsistencies, particularly with figure aspect ratios when title is added. Also, particular case of Single Qubit Circuit the visible plot size in IPython is not visible. 
 - Unable to plot globalphase gate

### FIX:
- Separate plotting dimensions specifically for IPython environments
- Adjusted figure sizing to be dynamic, removing direct user control over figure size to maintain aspect ratio.

### DIFF
--- a/src/qutip_qip/circuit/mat_renderer.py
+++ b/src/qutip_qip/circuit/mat_renderer.py
@@ -644,7 +644,14 @@ class MatRenderer(BaseRenderer):
 
         else:
 
-            adj_targets = [i + self._cwires for i in sorted(gate.targets)]
+            adj_targets = [
+                i + self._cwires
+                for i in sorted(
+                    gate.targets
+                    if gate.targets is not None
+                    else list(range(self._qwires)) # adaptation for globalphase
+                )
+            ]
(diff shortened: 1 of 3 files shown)


src/qutip_qip/circuit/mat_renderer.py
```diff
@@ -881,25 +876,37 @@ def _fig_config(self) -> None:
...
+
+        try:
+            get_ipython()
+            max_dim = max(xlim[1] - xlim[0], ylim[1] - ylim[0])
+            self.fig.set_size_inches(max_dim, max_dim, forward=True)
+        except NameError:
+            self.fig.set_size_inches(
+                xlim[1] - xlim[0], ylim[1] - ylim[0], forward=True
             )
-        self._ax.set_aspect("equal")
```
BoxiLi, line 902: Looks good.

Why do we need to use this "square" size for ipython? Maybe leave some inline comments explaining the reason?

PR #260, opened 2025-01-10
Added file save checks for renderers

### LINKED ISSUE #249: saving a circuit with latex method as a PDF fails
If a user wants to call 

`circuit.draw(renderer="latex", file_type="pdf", file_name="circuit")`

it fails with

`TypeError: TeXRenderer._convert_pdf() takes 1 positional argument but 2 were given`

This is because https://github.com/qutip/qutip-qip/blob/7633f6e4dbae7e6931eda800df21d42016df8431/src/qutip_qip/circuit/texrenderer.py#L267C4-L267C33

is called with a DPI argument.

gadhvirushiraj:
Fixes #249

Added a test to verify that the Matplotlib, Text, and LaTeX renderers  saves circuit files in `PNG, TXT, and PDF` formats. The test checks file creation using the `tmpdir` fixture for multiple QC fixtures.

### DIFF
--- a/tests/test_renderer.py
+++ b/tests/test_renderer.py
@@ -172,3 +172,37 @@ def test_matrenderer(request, qc_fixture):
 
     with patch("matplotlib.pyplot.show"):
         qc.draw("matplotlib")
+
+
+@pytest.mark.parametrize("qc_fixture", ["qc1", "qc2", "qc3"])
+def test_circuit_saving(request, qc_fixture, tmpdir):
+    """
+    Test if the different renderers can save the circuit in different formats.
+    """
+
+    qc = request.getfixturevalue(qc_fixture)
+
+    # test MatRenderer
+    with patch("matplotlib.pyplot.show"):
+        qc.draw("matplotlib", save=True, file_path=str(tmpdir.join("test")))
+    assert tmpdir.join(
+        "test.png"
+    ).check(), "MatRenderer saved PNG file not found."
+
+    # test TextRenderer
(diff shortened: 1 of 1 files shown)


BoxiLi:
Thanks @gadhvirushiraj , the error seems independent of the changes?

tests/test_renderer.py
```diff
@@ -172,3 +172,25 @@ def test_matrenderer(request, qc_fixture):
...
+@pytest.mark.parametrize("qc_fixture", ["qc1", "qc2", "qc3"])
+def test_circuit_saving(request, qc_fixture, tmpdir):
+    """
+    Test if the different renderers can save the circuit in different formats.
+    """
+
+    qc = request.getfixturevalue(qc_fixture)
+
+    # test MatRenderer
+    with patch("matplotlib.pyplot.show"):
```
BoxiLi, line 186: What is this line for?

gadhvirushiraj: when you call `draw()`, it invokes `plt.show()` at the end of the rendering process to display the circuit. This triggers a pop-up window for the plot, which requires manual closure to allow the test case/script to proceed. Using with `patch("matplotlib.pyplot.show"):` suppresses this behavior/

BoxiLi: Ah nice, could you add this as a comment to the code? Then I'll merge.

gadhvirushiraj:
> Thanks @gadhvirushiraj , the error seems independent of the changes?

yeah, something else

PR #261, opened 2025-01-10
Add family package entry point with about information

hodgestar:
A draft implementation of the `about` entrypoint for https://github.com/qutip/qutip/pull/2604.

### DIFF
--- a/src/qutip_qip/family.py
+++ b/src/qutip_qip/family.py
@@ -1,9 +1,7 @@
 """QuTiP family package entry point."""
 
+from . import __version__
 
-def about():
+def version():
     """Return information to include in qutip.about()."""
-    return "qutip-qip", [
-        "Moo: 1.5",
-        "Bar: 2.0",
-    ]
+    return "qutip-qip", __version__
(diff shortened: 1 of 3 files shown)


src/qutip_qip/family.py
```diff
@@ -0,0 +1,9 @@
+"""QuTiP family package entry point."""
+
+
+def about():
+    """Return information to include in qutip.about()."""
+    return "qutip-qip", [
+        "Moo: 1.5",
+        "Bar: 2.0",
```
hodgestar, line 8: @BoxiLi What information should I include here?

BoxiLi: Thanks @hodgestar !
Those are the dependent packages I assume? If we exclude those already listed in qutip, we only have optional packages such as `qutip-qtrl` and `qiskit` here.

hodgestar: I change the implementation to provide just the version of the family package. We can add more complex functionality later if it turns out we need it.

hodgestar:
The test failures here should be fixed by the release of QuTiP 5.1.1.

Ericgig (approved):

hodgestar:
@BoxiLi We should merge this before the next qutip-qip release but we can wait for the tests to pass if you like.

Issue #269, opened 2025-02-27
ZZ Crosstalk typo


ferrispnugraha:
From Appendix C in [E. Magesan et al, PRA 2020](https://doi.org/10.1103/PhysRevA.101.052308) and ignoring higher order terms, I think ZZ/2 coefficient as in [`zz_coeff`](https://github.com/qutip/qutip-qip/blob/c003888d2e5b938ad70bf301d8ed1244a189cd68/src/qutip_qip/noise.py#L594) should have used `wq` instead of `wr`. This can be explained as $\frac{J^2 (\delta_1 + \delta_2)}{(\delta_1 + \Delta)(\Delta - \delta_2)}=J^2 (-\frac{1}{\delta_1 + \Delta} + \frac{1}{\Delta - \delta_2})$, where $\Delta = \bar{\omega}_1-\bar{\omega}_2$.

BoxiLi:
Hi @ferrispnugraha ! Great to see you here. Yes I think you are right, must have misread the definition on Delta here.

BoxiLi:
Oh yeah, for sure, please!

Issue #268, opened 2025-02-27
SCQubit Frequency and Coupling Details


ferrispnugraha:
Referring to [A. Blais et al, REv. Mod. Phys. 2021](https://doi.org/10.1103/RevModPhys.93.025005) and [E. Magesan et al, PRA 2020](https://doi.org/10.1103/PhysRevA.101.052308), I want to confirm if the dressed resonator frequency in [SCQubitsModel](https://github.com/qutip/qutip-qip/blob/master/src/qutip_qip/device/circuitqed.py#L287) should be 

```python
# Dressed resonator frequency
        wr_dr = []
        for i in range(num_qubits - 1):
            tmp = wr[i]
            tmp -= g[2 * i] ** 2 / (wq[i] - wr[i] + alpha[i])
            tmp -= g[2 * i + 1] ** 2 / (wq[i + 1] - wr[i] + alpha[i+1]) # use Q_{i+1}'s anharmonicity
            wr_dr.append(tmp)
        self.params["wr_dressed"] = wr_dr
```
as $\omega_r' = \omega_r - \frac{g^2}{\Delta + \alpha}$.

In addition, the [effective qubit coupling](https://github.com/qutip/qutip-qip/blob/c003888d2e5b938ad70bf301d8ed1244a189cd68/src/qutip_qip/device/circuitqed.py#L296) can rely on the dressed qubit frequencies instead, since $J = \frac{g_1 g_2 (\bar{\omega_1} + \bar{\omega_2} - 2\omega_r)}{2(\bar{\omega_1}-\omega_r)(\bar{\omega_2}-\omega_r)}$. We can change `wq` with `wq_dr`.

BoxiLi:
Yes, you are right. `alpha[i+1]` should be used. This `wr_dressed` is not really used anywhere later, I think.

> In addition, the [effective qubit coupling](https://github.com/qutip/qutip-qip/blob/c003888d2e5b938ad70bf301d8ed1244a189cd68/src/qutip_qip/device/circuitqed.py#L296) can rely on the dressed qubit frequencies instead, since $J = \frac{g_1 g_2 (\bar{\omega_1} + \bar{\omega_2} - 2\omega_r)}{2(\bar{\omega_1}-\omega_r)(\bar{\omega_2}-\omega_r)}$. We can change `wq` with `wq_dr`.

[...]

PR #270, opened 2025-03-05
Use wq instead of wr for zz_coeff

ferrispnugraha:
Refer the fix to issue #269

### DIFF
--- a/src/qutip_qip/device/circuitqed.py
+++ b/src/qutip_qip/device/circuitqed.py
@@ -283,7 +283,7 @@ class SCQubitsModel(Model):
             if i != (num_qubits - 1):
                 tmp += g[2 * i] ** 2 / (wq[i] - wr[i])
             wq_dr.append(tmp)
-        self.params["wq_dressed_ori"] = wq_dr
+        self.params["wq_dressed_cavity"] = wq_dr
         # Dressed resonator frequency
         wr_dr = []
         for i in range(num_qubits - 1):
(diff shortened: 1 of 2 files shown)


BoxiLi:
Thanks! Could you also include the fix in #268? The value is not used, but let's still fix it.

ferrispnugraha:
> BTW, the failing test is some qutip core package updates, unrelated to your change.

Regarding #268, I am not very sure on where to put the Z shift for the qubit-qubit coupling term. Should it lie in the sz pulse?

BoxiLi:
I think it can be added after the definition of  Effective qubit coupling strength `J` and before the zx terms

Here, one can further modify the elements `self.params["wq_dressed"]` to include this shift. Similar to those lines
https://github.com/qutip/qutip-qip/blob/c003888d2e5b938ad70bf301d8ed1244a189cd68/src/qutip_qip/device/circuitqed.py#L289-L294

compute this shift and modify with `self.params["wr_dressed"] += new_shift`. In the final reduced Hamiltonian, we only have the Transmons, and the capacity is traced out.

I think it is better to add it here instead of the `sz` pulse because it is always-on.

ferrispnugraha:
All done, thanks for the guidance, @BoxiLi ! Please help to verify

BoxiLi:
Sorry for the delay! Looks good, I have just one comment. Could we maybe rename `wq_dressed_ori` to something like `wq_dressed_cavity`? I understand that this represents the dressed qubit frequency caused by the cavity without the J coupling.

ferrispnugraha:
No worries, the changes are in the latest commit now

BoxiLi:
Thanks a lot!

Issue #262, opened 2025-01-12
Generate typehint for all the functions


BoxiLi:
Currently, not all functions and classes in the codebase have type hints. In #255 and #256, I used an LLM to generate type hints, but the process was inefficient and cumbersome, as it required feeding each function individually (at least at that time). This approach makes it challenging to regenerate or update type hints when changes occur.

To improve this:
- It would be nice to have them automatically generated. We should use the modern syntax `int | list[int] | None`. @gadhvirushiraj found MonkeyType, which will generate type hints based on the testing code.
- Those type hints added always need some post-processing. And we may find some small bugs while handling them. Therefore it is best to do this module by module, allowing for manageable PRs and manual adjustments.

Mayank447:
Couple of points for anyone who wants to resolve this in the future:

1. As of now current minimum Python version we support is Python 3.10. So use `|` in place of `Optional`, `list` in place of `List` etc. Python 3.11 and 3.12 does include several new features for typehinting (so if the minimum python is bumped to those version by then), please utilise them.

2. This task should be done module by module in separate PRs instead of a single one. For the several that will become clear below.

3. Currently in the codebase there are several undefined behaviour/bugs that are not captured in typehints/docstring e.g. at several places in the codebase `Iterable`, `Sequence` is being used interchangebly. Similarly in the codebase at several places we should be using `Sequence` (a more general input category which our function supports) in typehints instead of `list` etc. The process of adding typehints indirectly helps in finding such bugs. Additionally there is quite a bit of technical debt (things which must be simplified in design), adding typehints for these is complicated. This is also the reason I am personally not in favour using AI for this task because of the latter.

4. We should use `pydoclint` to ensure the doc string matches the function signature and implementation.

5. Wherever necessary custom types can be added to `typing.py`.

PR #271, opened 2025-03-12
Adding type hints to circuit module

gadhvirushiraj:
### **PR Summary**  
This PR introduces type hints to the codebase using `MonkeyType`. Initial applied to a single circuit module, and if feasible, can be extended to other modules.  (Issue: #262)

Encountered issues with `pytest decorators` while running through the test suite. This was resolved using [pytest-monkeytype](https://github.com/mariusvniekerk/pytest-monkeytype) lib, which allows `MonkeyType` to work correctly with pytest. Still requires a bit post processing, but it is minimal.

`QuTiP-QIP supports Python 3.9+`, so we are unable to use Python 3.10+ syntax (e.g., `|` for `Union`). Also, `MonkeyType` is unable to generate in this format directly.

#### **Files Type Hint Status**  
- [ ] `_decompose.py`  
- [x] `base_renderer.py` (already present)
- [x] `circuit.py`  
- [x] `circuitsimulator.py`  
- [x] `color_theme.py`  (not needed)
- [x] `mat_renderer.py`  already present)
- [ ] `texrenderer.py`  
- [x] `text_renderer.py` (already present)

### DIFF
--- a/src/qutip_qip/circuit/circuit.py
+++ b/src/qutip_qip/circuit/circuit.py
@@ -2,18 +2,15 @@
 Quantum circuit representation and simulation.
 """
 
-import os
+from __future__ import annotations
 import inspect
-from io import BytesIO
-from itertools import product
-from functools import partialmethod
-from typing import Optional, Union, Tuple, List, Dict, Any
+from typing import Optional, Union, Tuple, List, TYPE_CHECKING
 from collections.abc import Iterable
 
 import numpy as np
 from copy import deepcopy
 
-from .texrenderer import TeXRenderer, CONVERTERS
+from .texrenderer import TeXRenderer
 from ._decompose import _resolve_to_universal, _resolve_2q_basis
 from ..operations import (
     Gate,
@@ -27,6 +24,9 @@ from .circuitsimulator import (
 )
 from qutip import Qobj, qeye
 
+if TYPE_CHECKING:
+    from IPython.core.display import Image
+    from ..qasm import QasmOutput
 
 try:
     from IPython.display import Image as DisplayImage, SVG as DisplaySVG
@@ -79,13 +79,13 @@ class QubitCircuit:
 
     def __init__(
         self,
-        N,
-        input_states=None,
-        output_states=None,
-        reverse_states=True,
-        user_gates=None,
-        dims=None,
-        num_cbits=0,
+        N: int,
+        input_states: Optional[Union[List[str], List[Optional[str]]]] = None,
+        output_states: Optional[List[Optional[str]]] = None,
+        reverse_states: bool = True,
+        user_gates: None = None,
+        dims: Optional[List[int]] = None,
+        num_cbits: int = 0,
     ):
         # number of qubits in the register
         self.N = N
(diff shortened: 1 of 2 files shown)


src/qutip_qip/circuit/circuit.py
```diff
@@ -200,16 +209,16 @@ def add_measurement(
...
-        classical_control_value=None,
-        style=None,
+        gate: Any,
+        targets: Optional[Union[List[int], int]] = None,
+        controls: Optional[Union[List[int], int]] = None,
+        arg_value: Optional[Any] = None,
+        arg_label: Optional[str] = None,
+        index: Optional[List[int]] = None,
+        classical_controls: Optional[List[int]] = None,
+        control_value: None = None,
```
BoxiLi, line 219: This should be 
```
control_value: Optional[int] = None
```
I think

src/qutip_qip/circuit/circuit.py
```diff
@@ -454,12 +474,12 @@ def reverse_circuit(self):
...
         self,
-        state,
-        cbits=None,
-        U_list=None,
-        measure_results=None,
-        precompute_unitary=False,
-    ):
+        state: Qobj,
+        cbits: Optional[List[int]] = None,
+        U_list: None = None,
```
BoxiLi, line 479: This is a bug, `U_list` is no longer used. We can remove it.

src/qutip_qip/circuit/circuitsimulator.py
```diff
@@ -293,10 +310,17 @@ def __init__(
...
         return self._qc
 
-    def initialize(self, state=None, cbits=None, measure_results=None):
+    def initialize(
+        self,
+        state: Optional[Qobj] = None,
+        cbits: Optional[List[int]] = None,
+        measure_results: Optional[
+            Union[List[int], Tuple[str, str], Tuple[str, str, str], Tuple[()]]
+        ] = None,
```
BoxiLi, line 322: This is wrong again. I guess probably we should not return tuples with undetermined length in the code and should use a list instead? What do you think?

This should be list of int or str.

src/qutip_qip/circuit/circuitsimulator.py
```diff
@@ -371,7 +395,14 @@ def state(self):
...
             return self._state
 
-    def run(self, state, cbits=None, measure_results=None):
+    def run(
+        self,
+        state: Qobj,
+        cbits: Optional[List[int]] = None,
+        measure_results: Optional[
+            Union[List[int], Tuple[str, str], Tuple[str, str, str], Tuple[()]]
+        ] = None,
```
BoxiLi, line 404: Same

src/qutip_qip/circuit/circuitsimulator.py
```diff
@@ -588,7 +629,18 @@ class CircuitResult:
...
+        self,
+        final_states: Optional[
+            Union[
+                List[Optional[Qobj]],
+                List[Qobj],
+                Qobj,
+            ]
+        ],
+        probabilities: Union[List[int], float64, int, List[float64]],
+        cbits: Optional[Union[List[List[int]], List[int], List[int64]]] = None,
```
BoxiLi, line 642: This float64 and int64 are about weird. I don't see the reason.

BoxiLi (commented):
Overall looks good. I found some edge cases in the comments.

BoxiLi:
Close since the codebase has been updated significantly since the PR.

PR #275, opened 2025-04-08
Created qpe.py

AbdullahKazi500:
implementation provides a quantum phase estimation algorithm that:

Complements the existing QFT implementation (actually using it as part of the process)
Uses fewer than 10 qubits making it suitable for pulse-level simulation
Returns a QubitCircuit object following the existing pattern
For the issue

### DIFF
--- a/src/qutip_qip/algorithms/qpe.py
+++ b/src/qutip_qip/algorithms/qpe.py
@@ -1,0 +1,123 @@
+"""
+This module provides the circuit implementation for Quantum Phase Estimation.
+"""
+
+import numpy as np
+from ..operations import Gate, snot, cphase, expand_operator
+from ..circuit import QubitCircuit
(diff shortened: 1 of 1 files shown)


BoxiLi:
Hey, is this closed on purpose or by accident?

hodgestar:
> Hi @BoxiLi this was supposed to be my contribution PR for Gsoc Qutip

Where did this strange idea come from?

hodgestar:
Note: @AbdullahKazi500 edited his comment to remove the line I was referring to & then I accidentally deleted one of the comments, and there is no undo button, so the context is missing in this comment thread.

For the record, the strange idea I was referring to is the idea (mentioned by @AbdullahKazi500) that QuTiP only accepts GSoC submissions from people enrolled in Computer Science programs. This is definitely not true.

Issue #264, opened 2025-01-12
Add more interesting algorithms


BoxiLi:
`qutip-qip.algorithm` remains mostly empty, it has only a qft algorithm, it would be very interesting to include more common algorithms.

We should focus on algorithms that use fewer than 10 qubits, so that there might be a chance to simulate them with the pulse-level simulator, such as the repetition error correction code. The algorithms should have a unified format, all return a `QubitCircuit` object.

rochisha0:
hi @BoxiLi , can i take up this issue?

hodgestar:
@rochisha0 Instead of asking whether you can take on an issue, it would be better to think a bit about the issue and either ask a question about what needs to be done, or propose some specific work you'd like to do and ask for feedback.

An example of the first might be: "What algorithm would you suggest adding next?"

An example of the second might be: "I'd like to add a VQE algorithm, what do you think?"

In open source projects issues are not typically earmarked for any one person. People contribute as capacity allows and interest dictates. You don't need permission to start work on an issue, but do get feedback on the approach early.

BoxiLi:
Thanks, @hodgestar for the suggestions!

Indeed those issues are all open. Ideas or proposals/implementations in more detail are surely all welcome @rochisha0 ;)

rochisha0:
Thanks @hodgestar  I understand that some open-source repositories handle things differently by assigning a person to each issue to avoid overlapping efforts.

@BoxiLi  I'll take some time to brainstorm and come up with ideas.

nwlambert:
just to add, if you are familiar with some small examples of some basic error correction codes, those would be great!  but just a suggestion.

rochisha0:
@nwlambert thanks for the suggestion.

I read the `qft` code, it has `qft` implemented in 3 ways ["qft", "qft_steps", "qft_gate_sequence"].

Do we require all 3 for future algorithms or just return `QubitCircuit` object?

@BoxiLi

rochisha0:
@BoxiLi any info on the same?

BoxiLi:
Hey, sorry, I remember I replied to this somehow. Must have missed it.

I would say it is enough to have one function that returns the full Circuit object.

quantockhills:
[...]

PR #276, opened 2025-04-12
Add quantum phase estimation in the list of algorithms

rochisha0:
Adding quantum phase estimation in the list of algorithms. Issue #264

### DIFF
--- a/src/qutip_qip/algorithms/qft.py
+++ b/src/qutip_qip/algorithms/qft.py
@@ -9,7 +9,7 @@ from qutip import Qobj
 from ..decompose import decompose_one_qubit_gate
 
 
-__all__ = ["qft", "qft_steps", "qft_gate_sequence"]
+__all__ = ["qft", "qft_steps", "qft_gate_sequence", "inverse_qft_gate_sequence"]
 
 
 def qft(N=1):
@@ -133,6 +133,56 @@ def qft_gate_sequence(N=1, swapping=True, to_cnot=False):
                 qc.add_gate("SWAP", targets=[N - i - 1, i])
     return qc
 
+def inverse_qft_gate_sequence(N=1, swapping=True, to_cnot=False):
+    """
+    Create the inverse quantum Fourier transform gate sequence.
+    
+    Parameters
+    ----------
+    N : int, optional
+        Number of qubits (default: 1)
+    swapping : bool, optional
+        Flag to include SWAP gates at the beginning (default: True)
+    to_cnot : bool, optional
+        Flag to decompose controlled phase gates to CNOT gates (default: False)
+        
+    Returns
+    -------
+    QubitCircuit
+        Gate sequence implementing inverse QFT
+    """
+    if N < 1:
+        raise ValueError("Minimum value of N can be 1")
+    
+    qc = QubitCircuit(N)
+    
+    if N == 1:
+        qc.add_gate("SNOT", targets=[0])
+    else:
+        if swapping:
+            for i in range(N // 2):
+                qc.add_gate("SWAP", targets=[N - i - 1, i])
+        
+        # Inverse QFT main algorithm
+        for i in range(N - 1, -1, -1):
+            qc.add_gate("SNOT", targets=[i])
+            
+            for j in range(i - 1, -1, -1):
+                if not to_cnot:
+                    qc.add_gate(
+                        "CPHASE",
+                        targets=[j],
+                        controls=[i],
+                        arg_label=r"{-\pi/2^{%d}}" % (i - j),
+                        arg_value=-np.pi / (2 ** (i - j)),
+                    )
+                else:
+                    decomposed_gates = _cphase_to_cnot(
+                        [j], [i], -np.pi / (2 ** (i - j))
+                    )
+                    qc.gates.extend(decomposed_gates)
+    
+    return qc
 
 def _cphase_to_cnot(targets, controls, arg_value):
     rotation = Qobj([[1.0, 0.0], [0.0, np.exp(1.0j * arg_value)]])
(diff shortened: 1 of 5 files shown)


rochisha0:
@BoxiLi is there a controlled U operator in QuTiP?

BoxiLi:
You can define a controlled gate object and add the object
https://github.com/qutip/qutip-qip/blob/d58a82dc54958428bd1d8695c486fda72db981a9/src/qutip_qip/operations/gateclass.py#L970-L993

rochisha0:
@BoxiLi this is ready for review

src/qutip_qip/algorithms/qpe.py
```diff
@@ -0,0 +1,212 @@
+import numpy as np
+from qutip import Qobj
+from qutip_qip.operations import Gate, ControlledGate
+from qutip_qip.circuit import QubitCircuit
+from qutip_qip.decompose import decompose_one_qubit_gate
+
+__all__ = ["qpe", "CustomGate"]
+
+
+class CustomGate(Gate):
```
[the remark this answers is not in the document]
BoxiLi: If it is only used in test, we can define it there and not expose it to the package name space.

src/qutip_qip/algorithms/qpe.py
```diff
@@ -0,0 +1,212 @@
...
+    return ControlledGate(
+        controls=controls,
+        targets=targets,
+        control_value=control_value,
+        target_gate=CustomGate,
+        U=U
+    )
+
+
+def _cphase_to_cnot(targets, controls, arg_value):
```
BoxiLi, line 54: We can import the `_cphase_to_cnot` directly from `qft.py`, even if it is not exposed to the namespace, no need to copy it. `from .algorithms.qft import ...` or something like this should work.

BoxiLi (changes requested):
Looks pretty good! Thanks a lot!

You can run `black` on `qpe.py` and `test_qpe.py` to clean the style. Codeclimate may have some additional complain such as trailing white space. You don't have to clean all of the code climate issues, if some of them become too tedious or irrelevant to the changes.

src/qutip_qip/algorithms/qpe.py
```diff
@@ -0,0 +1,212 @@
...
+    new_gates.append(Gate("RZ", targets=controls, arg_value=arg_value / 2))
+    
+    gate = decomposed_gates[7]
+    gate.arg_value += arg_value / 4
+    new_gates.append(gate)
+    
+    return new_gates
+
+
+def inverse_qft_gate_sequence(N=1, swapping=True, to_cnot=False):
```
BoxiLi, line 99: This looks nice, do you think maybe we should move it to `qft.py`? Add an additional keyword argument `inverse` to `qft_gate_sequence`?

tests/test_qpe.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        U = Qobj([[0, 1], [1, 0]])  
+        
+        controlled_u = create_controlled_unitary(controls=[0], targets=[1], U=U)
+        
+        assert_equal(controlled_u.controls, [0])
+        assert_equal(controlled_u.targets, [1])
+        assert_equal(controlled_u.control_value, 1)
+        
+        assert_(controlled_u.target_gate == CustomGate)
```
BoxiLi, line 40: target_gate is the same as the `CustomGate` class? Shouldn't it be a Gate initialised by `U`?

src/qutip_qip/algorithms/qpe.py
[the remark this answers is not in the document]
rochisha0: Is it possible to pass a random `Qobj` to gate class. I couldn't understand how to do that.

src/qutip_qip/algorithms/qpe.py
rochisha0: I transferred the function to the `qft` file. Let me now if that seems okay?

src/qutip_qip/algorithms/qft.py
```diff
@@ -133,6 +133,56 @@ def qft_gate_sequence(N=1, swapping=True, to_cnot=False):
                 qc.add_gate("SWAP", targets=[N - i - 1, i])
     return qc
 
+def inverse_qft_gate_sequence(N=1, swapping=True, to_cnot=False):
```
BoxiLi, line 136: Nice, is it possible to merge this into the `qft_gate_sequence`? They share the same input arguments. We could define `qft_gate_sequence(N=1, swapping=True, to_cnot=False, inverse=False)`, which returns the traditional qft if inverse=False and the inversed one if True.

The code also looks very similar, maybe we can even merge most of the code? I feel like there are only a few minus sign which can be defined according to the value of `inverse`. Please correct me if I and wrong.

src/qutip_qip/algorithms/qpe.py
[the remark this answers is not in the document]
BoxiLi: Ok... I get your point, usually the point of defining a gate class is to avoid storing the unitary every time and reuse them. So you are right, giving a different U as input verytime is not supported.

tests/test_qpe.py
BoxiLi: I got this now

src/qutip_qip/algorithms/qpe.py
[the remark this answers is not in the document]
BoxiLi: I feel like there should be a more elegant solution to this. But this is ok, for now.

src/qutip_qip/algorithms/qft.py
rochisha0: I tried doing this, the solution doesn't look elegant

BoxiLi: Ah that makes sense!

BoxiLi:
There are some inconsistencies in the last change. Should be easy fix, other than that I believe this is ready to be merged.

rochisha0:
Done @BoxiLi !

BoxiLi (approved):

BoxiLi:
Thanks! @rochisha0

PR #277, opened 2025-04-28
Add basic error correction codes in `qutip-qip` algorithms

rochisha0:
This PR aims to add basic error correction codes in `qutip-qip` algorithms.

- [x]  Create files for `phase-flip`, `bit flip`, and `shor code`.
- [ ] Write tests

### DIFF
--- a/src/qutip_qip/algorithms/error_correction/bit_flip.py
+++ b/src/qutip_qip/algorithms/error_correction/bit_flip.py
@@ -1,0 +1,112 @@
+# bit_flip_code.py
+
+import numpy as np
+from qutip import Qobj, tensor, basis, sigmax, sigmay
+from qutip_qip.operations import Gate
+from qutip_qip.circuit import QubitCircuit
+
+__all__ = ["BitFlipCode"]
+
+class BitFlipCode:
+    """
+    Implementation of the 3-qubit bit-flip code.
+    
+    The bit-flip code protects against X (bit-flip) errors by encoding
+    a single logical qubit across three physical qubits:
+    |0⟩ → |000⟩
+    |1⟩ → |111⟩
+    """
+
+    @staticmethod
+    def encode_circuit():
+        """
+        Create a circuit for encoding a single qubit into the 3-qubit bit-flip code.
+        
+        Returns
+        -------
+        qc : instance of QubitCircuit
+            Encoding circuit for the bit-flip code.
+        """
+        qc = QubitCircuit(3)
+        qc.add_gate("CNOT", controls=0, targets=1)
+        qc.add_gate("CNOT", controls=0, targets=2)
+        return qc
+    
+    @staticmethod
+    def syndrome_measurement_circuit():
+        """
+        Create a circuit for syndrome measurement of the 3-qubit bit-flip code.
+        
+        Returns
+        -------
+        qc : instance of QubitCircuit
+            Syndrome measurement circuit that uses two ancilla qubits.
+        """
+        qc = QubitCircuit(5)  # 3 data qubits + 2 syndrome qubits
+        
+        # First syndrome measurement: Parity of qubits 0 and 1
+        qc.add_gate("CNOT", controls=0, targets=3)
+        qc.add_gate("CNOT", controls=1, targets=3)
+        
+        # Second syndrome measurement: Parity of qubits 1 and 2
+        qc.add_gate("CNOT", controls=1, targets=4)
+        qc.add_gate("CNOT", controls=2, targets=4)
+        
+        return qc
+    
+    @staticmethod
+    def correction_circuit(syndrome):
+        """
+        Create a circuit for error correction based on syndrome measurement.
+        
+        Parameters
+        ----------
+        syndrome : tuple
+            Two-bit syndrome measurement result (s1, s2).
+            
+        Returns
+        -------
+        qc : instance of QubitCircuit
+            Correction circuit applying X gates as needed.
+        """
+        qc = QubitCircuit(3)
+        s1, s2 = syndrome
+        
+        # Syndrome interpretation:
+        # s1=0, s2=0: No error
+        # s1=1, s2=0: Error on qubit 0
+        # s1=1, s2=1: Error on qubit 1
+        # s1=0, s2=1: Error on qubit 2
+        
+        if s1 == 1 and s2 == 0:
+            # Error on qubit 0
+            qc.add_gate("X", targets=0)
+        elif s1 == 1 and s2 == 1:
+            # Error on qubit 1
+            qc.add_gate("X", targets=1)
+        elif s1 == 0 and s2 == 1:
+            # Error on qubit 2
+            qc.add_gate("X", targets=2)
+            
+        return qc
+    
+    @staticmethod
+    def decode_circuit():
+        """
+        Create a circuit for decoding the 3-qubit bit-flip code.
+        
+        Returns
+        -------
+        qc : instance of QubitCircuit
+            Decoding circuit for the bit-flip code.
+        """
+        qc = QubitCircuit(3)
+        qc.add_gate("CNOT", controls=0, targets=2)
+        qc.add_gate("CNOT", controls=0, targets=1)
+        
+        # Add a Toffoli gate to verify the parity
+        # If all qubits have the same value, the result is stored in qubit 0
+        qc.add_gate("TOFFOLI", controls=[1, 2], targets=0)
+        
+        return qc
+    
--- a/src/qutip_qip/algorithms/__init__.py
+++ b/src/qutip_qip/algorithms/__init__.py
@@ -1,1 +1,2 @@
 from .qft import *
+from .error_correction.bit_flip import *
--- a/src/qutip_qip/algorithms/error_correction/phase_flip.py
+++ b/src/qutip_qip/algorithms/error_correction/phase_flip.py
@@ -1,0 +1,130 @@
+import numpy as np
+from qutip import Qobj, tensor, basis, sigmax, sigmay, sigmaz
+from qutip_qip.operations import Gate
+from qutip_qip.circuit import QubitCircuit
+
+__all__ = ["PhaseFlipCode"]
+
+class PhaseFlipCode:
+    """
+    Implementation of the 3-qubit phase-flip code.
+    The phase-flip code protects against Z (phase-flip) errors by encoding
+    a single logical qubit across three physical qubits:
+    |0⟩ → |+++⟩
+    |1⟩ → |---⟩
+    
+    This is accomplished by applying Hadamard gates to transform the bit-flip code
+    into the phase-flip code, as phase errors in the Hadamard basis appear as bit-flips.
+    """
+    
+    @staticmethod
+    def encode_circuit():
+        """
+        Create a circuit for encoding a single qubit into the 3-qubit phase-flip code.
+        
+        Returns
+        -------
+        qc : instance of QubitCircuit
+            Encoding circuit for the phase-flip code.
+        """
+        qc = QubitCircuit(3)
+        
+        # First apply Hadamard to the input qubit
+        qc.add_gate("SNOT", targets=0)
+        
+        # Then use the bit-flip encoding structure
+        qc.add_gate("CNOT", controls=0, targets=1)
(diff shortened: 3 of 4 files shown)


rochisha0:
@BoxiLi this is ready for review!

BoxiLi:
Thanks a lot! This looks really cool. sorry for the delay.

The code looks pretty good to me. Two general questions before a more detailed review:

- Is class+3 static methods the best way to define this? I understand that the point here is because all of them have encoding, correction and decoding. So they share the same structure. However in this way one always has to first define a class instance and then use those methods. Do you know how other packages handle this?
- This could be for this RP or a separate PR, depending on your opinion. Is there a way to make the function more general. For instance, the bit flip encoding circuit is defined for encoding qubit 0 to qubit 0,1,2. Can we define such a function like `encode_circuit(qubit=0, ancilla=[1,2])` such that the resulting circuit will be constructed depending on the input parameters?. In this way one can generate multi-qubit error correction circuit based on those elementary functions.

BoxiLi:
Another question, usually, the dencode circuit has measurement and classically controlled operations (for example here https://quantumcomputing.stackexchange.com/questions/6501/2-ways-to-do-the-three-qubits-bit-flip-code), which I don't see in the code?

Could you also add some reference links/papers as inline comments to the code?

rochisha0:
A lot of the structure is taken from https://github.com/qiskit-community/qiskit-community-tutorials/blob/master/awards/teach_me_quantum_2018/intro2qc/10.Quantum%20error%20correction.ipynb?utm_source=chatgpt.com and Neilsen and Chuang.

rochisha0:
@BoxiLi , it's ready for review again!

BoxiLi:
Also, in case you are not aware of it, qutip-qip also allows adding measurement operation
`qc.add_measurement(...)`
and classically control the following circuit depending on the measurement output
`qc.add_gate(X, classical_control=..., classical_control_value...)`.
I can make a minimal example later. I don't have a coding environment for now.

The syndrome measurement and correction circuit can be defined using measurement and classically controlled operations I believe. Measurement operations should be added to the syndrome circuit correction circuit should be classically controlled gate instead of manually providing the syndrome measurement results to `correction_circuit(self, syndrome)`. The resuling circuit should be like
image

rochisha0:
Hello @BoxiLi , thank you for your suggestions. I will try to incorporate these suggestions in my code. 

I have a call going with @ajgpitch on every Tuesday, maybe you can join that one for a week or two?

rochisha0:
@BoxiLi  have made the discussed changes to `bit_flip`. If those look good I will make the same to other files and change the tests

src/qutip_qip/algorithms/error_correction/bit_flip.py
```diff
@@ -0,0 +1,70 @@
...
+        Encode a logical qubit into three physical qubits.
+    syndrome_and_correction_circuit(data_qubits, syndrome_qubits)
+        Extract error syndrome and apply correction via classical control.
+    decode_circuit(data_qubits)
+        Decode the logical qubit back to a single physical qubit.
+    """
+
+    def __init__(self):
+        self.n_data = 3
+        self.n_syndrome = 2
```
BoxiLi: I assume `self.n_data` and `self.n_syndrome` are things that are specific for this algorithm and we do not want people to change them, right? So I would suggest mark them as private members `self._n_data` and `self._n_syndrome`. In the Python convention, this means that those are private members and should not be accessed by users.

src/qutip_qip/algorithms/error_correction/bit_flip.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def encode_circuit(self, data_qubits):
+        assert len(data_qubits) == self.n_data
```
BoxiLi: We should raise an error instead of using assert. Assert is for internal checking. A user should in principle never see `assert` error. Raise a `ValueError` saying the number is wrong is more appropriate here.

src/qutip_qip/algorithms/error_correction/bit_flip.py
```diff
@@ -0,0 +1,70 @@
...
+        qc = QubitCircuit(max(data_qubits) + 1)
+        control = data_qubits[0]
+        for target in data_qubits[1:]:
+            qc.add_gate("CNOT", controls=control, targets=target)
+        return qc
+
+    def syndrome_and_correction_circuit(self, data_qubits, syndrome_qubits):
+        assert len(data_qubits) == self.n_data
+        assert len(syndrome_qubits) == self.n_syndrome
```
BoxiLi: Same, raise errors.

src/qutip_qip/algorithms/error_correction/bit_flip.py
```diff
@@ -0,0 +1,70 @@
...
+        sq = syndrome_qubits
+
+        # Syndrome extraction
+        qc.add_gate("CNOT", controls=dq[0], targets=sq[0])
+        qc.add_gate("CNOT", controls=dq[1], targets=sq[0])
+        qc.add_gate("CNOT", controls=dq[1], targets=sq[1])
+        qc.add_gate("CNOT", controls=dq[2], targets=sq[1])
+
+        # Measurements into classical bits
+        qc.add_measurement(sq[0], classical_store=0)
```
BoxiLi: Should be 
```python
qc.add_measurement(sq[0], sq[0], classical_store=0)
```
The first one is just the name of the measurement operation, the second one is the target qubit.

This triggers an error when plotting the circuit, you can use `qc.draw()` to test and see if it is what you expected.

BoxiLi: BTW I might find bug in classical controlled gate plotting. If you don't see the control line plotted, that is fine.

BoxiLi (commented):
Thanks! Yes it is good to fisrt look at one part of it. I had a look at the BitFlipCode part. 

I think you might build the code on the wrong base branch, because I do not see `qpe.py`, which you added last time. This could be the reason for the conflict. You can try to rebase the code upon the current upstream master.

It would be cool if we can combine three circuit part (with manually added error maybe?) and test it!

src/qutip_qip/algorithms/error_correction/bit_flip.py
rochisha0: Yes, makes sense to make them private

src/qutip_qip/algorithms/error_correction/bit_flip.py
rochisha0: The `qc.draw()` now works as expected

rochisha0:
@BoxiLi  I've made the required changes in bit flip code as you requested. If this looks good I will implement similar changes in the other files. I will sort out the merge conflict towards the end.

src/qutip_qip/algorithms/error_correction/bit_flip.py
rochisha0: Thanks, this is now working as expected.

src/qutip_qip/algorithms/error_correction/bit_flip.py
```diff
@@ -0,0 +1,71 @@
...
+        qc.add_gate("CNOT", controls=dq[2], targets=sq[1])
+
+        # Measurements into classical bits
+        qc.add_measurement(sq[0], sq[0], classical_store=0)
+        qc.add_measurement(sq[1], sq[1], classical_store=0)
+
+        # Classically controlled correction
+        qc.add_gate("X", targets=dq[0], classical_controls=[0, 1], classical_control_value=[1, 0])
+        qc.add_gate("X", targets=dq[1], classical_controls=[0, 1], classical_control_value=[1, 1])
+        qc.add_gate("X", targets=dq[2], classical_controls=[0, 1], classical_control_value=[0, 1])
```
BoxiLi:
```suggestion
        # Measurements into classical bits
        qc.add_measurement(sq[0], sq[0], classical_store=0)
        qc.add_measurement(sq[1], sq[1], classical_store=1)

        # Classically controlled correction
        qc.add_gate("X", targets=dq[0], classical_controls=[0, 1], classical_control_value=2)
        qc.add_gate("X", targets=dq[1], classical_controls=[0, 1], classical_control_value=3)
        qc.add_gate("X", targets=dq[2], classical_controls=[0, 1], classical_control_value=1)
```
The `classical_control_value` is an integer representation of the bit value. (I forgot which encoding we are using, so not sure if it is 2,3,1 or 1,3,2. But definitely an integer.

BoxiLi:
I used the following code to test, in case it is helpful for building tests.
```python
import qutip
from qutip_qip.algorithms import (
    BitFlipCode,
)
#%% Encode qubit
code = BitFlipCode()
initial_state = qutip.rand_ket([2])
qc = code.encode_circuit([0, 1, 2])
# should start from random state
state1 = qutip.tensor([initial_state, qutip.basis([2]*4, [0]*4)])
# state1 = qutip.basis([2]*5, [0]*5)
state2 = qc.run(state1)
#%% Add a bit-flip error
qc_error = QubitCircuit(5, num_cbits=2)
qc_error.add_gate("X", targets=0)
state3 = qc_error.run(state2)
#%% Apply correction
qc2 = code.syndrome_and_correction_circuit([0, 1, 2], [3, 4])
qc2.gates = qc2.gates
qc2.draw()
state4 = qc2.run(state3)
# %% Check final qubit 0 state
print(qutip.expect(qutip.projection(2, 0, 0), initial_state))
print(qutip.expect(qutip.projection(2, 0, 0), state4.ptrace([0])))
```

BoxiLi:
Now, the last question is, what is the use of `decode_circuit` here, and do we still need it?

rochisha0:
@BoxiLi I think it is helpful for completeness of the algorithm, and anyone who is trying to learn about error correction. Not so in circuit building.

BoxiLi:
So, what I am a bit confused about is, do people really use this decode circuit? It seems to me that after the measurement and controlled correction, the algorithm is complete? What does the decode algorithm do?

rochisha0:
@BoxiLi , the decode circuit is for reversing the error encoding, with `TOFFOLI` as parity check.

BoxiLi:
Ok, let's explain clearly in the documentation that this decode_circuit is not the decoder that processes the measurement data, but the one that reverses the encode circuit.
Other than this the bit flip looks good.

rochisha0:
@BoxiLi I have pushed the changes!

BoxiLi:
Looks nice! Will have a closer look and then merge. The failing tests are unrelated, the one with scipy is fixed already in qutip. The other one one from matplotlib can be fixed by restricting the version in another PR I think.

But there is a weird error message in the doctest when testing against existing old code.
```
qft.py:16: in <module>
    from qutip_qip.algorithms import qft_gate_sequence
/opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/qutip_qip/algorithms/__init__.py:3: in <module>
    from .error_correction.bit_flip import *
E   ModuleNotFoundError: No module named 'qutip_qip.algorithms.error_correction'
```

BoxiLi:
If interested, you can also try the Steane 7-qubit code, which is also quite popular, I think. Or write a notebook tutorial.

rochisha0:
I'll check the doctest one.

I'll also look into creating a new PR for Steane code specifically.

BoxiLi:
I have fixed the matplotlib error. The doctest one is really strange. Maybe it is because the test environment installed the wrong version?

rochisha0:
@BoxiLi , yes maybe.

rochisha0:
@BoxiLi can this be because `error_correction` folder doesn't have an init file?

BoxiLi:
Hmm, I don't know, but usually it should have an init. Maybe we can try?

rochisha0:
@BoxiLi This change fixes the issue.

BoxiLi:
Why does Shor code only have `encode_circuit`?

The new classes should also be added to the API doc under `qutip-qip\doc\source\apidoc`

Other than those looks good!

BoxiLi:
The error here is because an `__all__` variable  is needed in each file to define open interfaces. If this is not defined, all things imported will be visible, which causes this error message in the doc.

rochisha0:
@BoxiLi The `doc` is now solved. I left `decode_circuit` for `ShorCode` for simplicity.

BoxiLi:
All looks good to me ;)

PR #278, opened 2025-06-15
Fix classical control rendering for gate

gadhvirushiraj:
- Fixed rendering issue for classical control on gates for both renderers.
- Optimize: Removed `_draw_singleq_gate` for rendering in MatRenderer.

### DIFF
--- a/src/qutip_qip/circuit/mat_renderer.py
+++ b/src/qutip_qip/circuit/mat_renderer.py
@@ -374,14 +374,8 @@ class MatRenderer(BaseRenderer):
 
         cbridge_l = plt.Line2D(
             (
-                xskip
-                + self.style.gate_margin
-                + self._min_gate_width / 2
-                - self._cwire_sep,
-                xskip
-                + self.style.gate_margin
-                + self._min_gate_width / 2
-                - self._cwire_sep,
+                xskip - self._cwire_sep,
+                xskip - self._cwire_sep,
             ),
             (
                 c_pos * self.style.wire_sep + self._arrow_lenght,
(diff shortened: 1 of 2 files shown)


BoxiLi:
Thanks a lot for such a quick PR! 

I noticed two things: 
1. This classical control line should also be added when we compute the layer of the gate. 
```python
qc = QubitCircuit(2, num_cbits=2)
qc.add_gate("X", targets=0, classical_controls=0, classical_control_value=0)
qc.add_gate("X", targets=1, classical_controls=1, classical_control_value=1)
qc.draw()
```
image

2. I think there might be a misunderstanding of the `classical_controls` and `classical_control_value`. I didn't realize it when you asked that question on Discord. I think here we should read `classical_controls`, which should be a list that records the control classical bits. And then we draw a circuit on it, like in the quantum controlled gate

image

gadhvirushiraj:
> 1. This classical control line should also be added when we compute the layer of the gate.

We dont maintain a record for classical lines. But we just augment this by maintaining gaps in qlines. But it should be now fixed for MatRenderer.

I have modified it to read `classical_controls` too.


Edit: TextRenderer is yet to fix

BoxiLi:
Great! Looks good!

BoxiLi:
Looks good! Rendered nicely on my machine. Would be nice to add one test case for the text renderer. Just to be safe.

The failing test is unrelated, but the one on matplotlib seems to be a warning from the old matplotlib 3.9

BoxiLi:
Thanks! This looks good to me. Shall I merge? @gadhvirushiraj

gadhvirushiraj:
@BoxiLi, Sorry for the delay.
Yes, please merge - the failed test looks unrelated.

PR #280, opened 2025-07-15
Bug fix in the VQA module

BoxiLi:
- Fix a bug in vqa test
The overlap computes the inner product, which is unfortunately imaginary in this test case, so the cost function 1-overlap has always real part 1.
This should never work. However, since the test was initialized in the right answer, pi/2, the optimizer always stopped pretty close to the right answer. So it seldom reports an error. But the changes in SciPy 1.16.0 somehow make the optimizer bolder and explore large values in the beginning, causing the test to fail.
- Adapt the test example
This is a trivial example to test the plotting function. Before, it was a VQA with no parameters, but since Scipy 1.16 the minimize function raises an error if there is no value to optimize. So I adapted the test.
- Remove the VQA bitstring cost function
I do not understand how the cost function can be based on one bit string sampling. Sampling over a quantum state is a statistical process; information is needed for the whole distribution. Sampling one bit string does not tell anything.
The test was passing just because it was initialized in state 1.

### DIFF
--- a/src/qutip_qip/vqa.py
+++ b/src/qutip_qip/vqa.py
@@ -778,11 +778,7 @@ class OptimizationResult:
         display: bool, optional
             Display the plot with the pyplot plot.show() method
         """
-        try:
-            import matplotlib.pyplot as plt
-        except Exception:
-            print("could not import matplotlib.pyplot")
-            quit()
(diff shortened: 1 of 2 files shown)


hodgestar (approved):
Thanks for fixing, @BoxiLi.

PR #285, opened 2025-11-28
Removed v4 support

Mayank447:
qutip.Options() was depreciated, replace with a `dict`

### DIFF
--- a/src/qutip_qip/circuit/circuit.py
+++ b/src/qutip_qip/circuit/circuit.py
@@ -2,18 +2,14 @@
 Quantum circuit representation and simulation.
 """
 
-import os
 import inspect
-from io import BytesIO
-from itertools import product
-from functools import partialmethod
-from typing import Optional, Union, Tuple, List, Dict, Any
 from collections.abc import Iterable
+from math import pi
 
 import numpy as np
 from copy import deepcopy
 
-from .texrenderer import TeXRenderer, CONVERTERS
+from .texrenderer import TeXRenderer
 from ._decompose import _resolve_to_universal, _resolve_2q_basis
 from ..operations import (
     Gate,
--- a/src/qutip_qip/qasm.py
+++ b/src/qutip_qip/qasm.py
@@ -7,6 +7,7 @@ from copy import deepcopy
 import warnings
 
 import numpy as np
+from math import pi
 
 from .circuit import QubitCircuit
 from .operations import (


BoxiLi:
Hi! Thank you for the pull request, there are a few other places where qutip.Options() is used. Could you also update those?

Mayank447:
@BoxiLi I have updated it at all places.

BoxiLi:
Hmmm, seems that it is not that easy, in the code there are a few places where the old option class is used to make it copatible with old qutip-v4. That is why the tests with qutip@4.7 fail. But since qutip-v5 has been out for 1.5 years. I think it is the time now to drop the support of old qutip-v4.

If you want to just complete this PR, you will need to then check the version of qutip and use the option class if qutip is v4. You will find how to do this in several places such as `modelprocessor.py`

If you want to go deeper. Before merging this, we need to update the package to remove qutip-v4 support. Here or in a different PR. The task is to remove all the branching for qutip5 vs 4 (usually you will see a `is_qutip5`) variable. And also remove the qutip@4.7 GitHub action test.

Mayank447:
I want to do the latter. Since `qutip-qip` was separated from v5 onwards only and has been been decently long since it is stable, makes sense to remove v4 support. Anyone using v4.7 probably won't need this package since qip was an inbuilt library in qutip v4.

Is there some sort of a release document for changes in v5 in addition to the variable `is_qutip5` you mentioned

BoxiLi:
Good question, I don't think it is documented where all the changes are made, because most of them works for both versions. Most of the changes are introduced in #136. I think it should be sufficient to search for `parse_version`, this is the package that we use to read the version of the package, which is needed to distinguish qutip-v5 and qutip-v4.

Mayank447:
@BoxiLi I went through you migration PR. I have removed the code v4 completely based on that, `is_qutip5 `, `parse_version` and anything else I could find. Also the checks for `testSpline` and `testGetObjEvo` for Processor has been rewritten for v5, they were earlier being skipped.

Similarly testing workflow for v4 has been removed from github workflow testing file and same for tutorial generation for v4.

Mayank447:
@BoxiLi Linting issue has been fixed. As for the failing qiskit test, I will work on that (#287) in a separate PR once this is merged and #289 is completed.

BoxiLi:
Thank you, looks nice! 
Ignore the windows test failing. It is unrelated and is a github action machine issue.
Commenting out all the qiskit test is a bit weird here. If you leave it here, will the qiskit tests all fail?

Mayank447:
I have uncommented the qiskit tests. I had them commented out on my local machine because I also had qiskit (v2) installed in the same environment and `pytest .` would fail otherwise. Since `test_qiskit.py` only tests for `qiskit` being installed, not the version it.

BoxiLi (approved):

BoxiLi:
Nice, all passed now.

PR #290, opened 2025-12-02
Cleaning the code base

Mayank447:
- Removed unused import completely and made imports absolute (i.e. `qutip_qip.circuit import QubitCircuit`) instead of relative `..circuit import QubitCircuit`. This helps when files are moved around or refactored.
- Removed many of the undefined variables, there are still a few left. Those will be removed in subsequent PRs as there are deeper errors in the code.
- Unused variables removed, similar point as above.
- Code refactored: `qasm.py` has been broken down into a module since the no. of lines of code was over 1300 (the external import structure remains the same), separated `Model` into a separate file, renderers were refactored into a module name `draw` under `src/circuits`

@BoxiLi This PR isn't fully finished yet, its still a draft. But can you review it please since the code changes especially refactor is pretty huge.

**Update:** qasm.py refactoring and circuit draw would be done in separate PR.

### DIFF
--- a/src/qutip_qip/circuit/__init__.py
+++ b/src/qutip_qip/circuit/__init__.py
@@ -6,11 +6,6 @@ from .circuitsimulator import CircuitResult, CircuitSimulator
 from .circuit import QubitCircuit
 from qutip_qip.operations import Gate, Measurement
 
-__all__ = [
-    "CircuitSimulator",
-    "CircuitResult",
-    "QubitCircuit",
-]
 
 def _add_deprecation(fun, msg):
     def newfun(*args, **kwargs):
@@ -34,3 +29,10 @@ Measurement = _add_deprecation(
     "The class Measurement has been moved to qutip_qip.operations."
     "Please use update the import statement.\n",
 )
+
+
+__all__ = [
+    "CircuitSimulator",
+    "CircuitResult",
+    "QubitCircuit",
+]
--- a/src/qutip_qip/device/utils.py
+++ b/src/qutip_qip/device/utils.py
@@ -2,6 +2,7 @@ from collections.abc import Iterable
 import numbers
 import numpy as np
 
+
 def _to_array(params, num_qubits):
     """
     Transfer a parameter to an array.
--- a/src/qutip_qip/operations/gateclass.py
+++ b/src/qutip_qip/operations/gateclass.py
@@ -42,7 +42,6 @@ from qutip_qip.operations import (
 )
 
 
-
 """
 .. testsetup::
 


BoxiLi:
Nice thanks for doing this. I will have a look, but try to avoid exploding the PR. Let's keep the size as small as possible (At the moment it seems ok, maybe move the qasm and circuit drawer change to a different one.). I just noticed that in the last PR the test_qiskit was mistakely commented out in the end, which I missed.

Also to avoid having to manually slient the qiskit test, I would recomment to create a different branch instead of using master and create a new environment, either using conda or virtual env. So the development won't affect your existing environment. And you can have multiple version at the sametime. This is good practice in general.

Mayank447:
Sure, I will keep this in mind. I will move qasm, circuit to a different PR, those refactor are pretty huge.

Mayank447:
@BoxiLi I have removed the commits for refactoring qasm and circuit draw, I will open separate individual PRs for those. This PR is ready for review

src/qutip_qip/circuit/__init__.py
```diff
@@ -2,9 +2,9 @@
 
 import warnings
 
-from .circuit import *
-from .circuitsimulator import *
-from ..operations import Gate, Measurement
+from .circuitsimulator import CircuitResult, CircuitSimulator
+from .circuit import QubitCircuit
+from qutip_qip.operations import Gate, Measurement
```
BoxiLi, line 9: I think for the `__init__.py` files, it is fine to use the star importation. We already defined `__all__.py` in all the `.py` files. This simplifies the development. Otherwise one have to always rember to update `__init__.py` when a new function is added.

src/qutip_qip/circuit/__init__.py
```diff
@@ -29,3 +29,10 @@ def newfun(*args, **kwargs):
     "The class Measurement has been moved to qutip_qip.operations."
     "Please use update the import statement.\n",
 )
+
+
+__all__ = [
+    "CircuitSimulator",
+    "CircuitResult",
+    "QubitCircuit",
+]
```
BoxiLi, line 38: Same as above.

src/qutip_qip/circuit/__init__.py
BoxiLi: And is there a reason that relative import is kept here but removed elsewhere? I don't have strong preference but it should be consistent.

src/qutip_qip/compiler/circuitqedcompiler.py
```diff
@@ -1,10 +1,7 @@
 import numpy as np
 
-from ..operations import Gate
-from ..compiler import GateCompiler, Instruction
-
-
-__all__ = ["SCQubitsCompiler"]
```
BoxiLi, line 7: Why is `__all__` removed here?

I think typically we do in qutip family package is to keep this in all `.py` files. Not `__init__.py`. Unless there is some advantages?

BoxiLi: This applies to many other files.

src/qutip_qip/circuit/__init__.py
Mayank447: In Python by defining `__all__` in `__init__.py`, one is explicitly stating: "These are the tools intended for public use. Anything else you find in the submodules is internal logic and might change without warning." There were many cases of internal functions (starting with _) being exported through `__all__` in .py file which generally is not a good practice 

`__all__` in .py file is better, if we want to define the import at a file level instead of module level.

Mayank447: Generally relative imports are only kept at module level ( in `__init__.py`). Relative imports from parent are strongly discouraged for large packages because refactor breaks a lot of things. One keeps the relative imports in `__init__` only because some people map a package directory to a different name in pyproject.toml

Mayank447: For reference you can check: https://github.com/Qiskit/qiskit/blob/main/qiskit/quantum_info/__init__.py

Not defining `__all__` in `__init__.py` inherently means everything imported In `__init__.py` will be exported when does `from package import *`

BoxiLi:
Thanks!

PR #293, opened 2025-12-13
QiskitSimulator in QuTip for qiskit v2

### LINKED ISSUE #287: Broken Qiskit Backend
The Qiskit Backend is broken, this is evident from `pytest tests/test_qiskit.py`

`=========================================================== ERRORS ===========================================================
______________________________________________ ERROR collecting test_qiskit.py _______________________________________________
test_qiskit.py:18: in <module>
    from qutip_qip.qiskit.provider import (
../src/qutip_qip/qiskit/__init__.py:3: in <module>
    from .provider import Provider
../src/qutip_qip/qiskit/provider.py:4: in <module>
    from .backend import QiskitCircuitSimulator, QiskitPulseSimulator
../src/qutip_qip/qiskit/backend.py:18: in <module>
    from qiskit.providers.models import QasmBackendConfiguration
<frozen importlib._bootstrap>:1412: in _handle_fromlist
    ???
/Users/mayankgoel/miniconda3/envs/qiskit/lib/python3.12/site-packages/qiskit/providers/models/__init__.py:80: in __getattr__
    warnings.warn(
E   DeprecationWarning: qiskit.providers.models is deprecated since Qiskit 1.2 and will be removed in Qiskit 2.0. With the removal of Qobj, there is no need for these schema-conformant objects. If you still need to use them, it could be because you are using a BackendV1, which is also deprecated in favor of BackendV2.
================================================== short test summary info ===================================================
ERROR test_qiskit.py - DeprecationWarning: qiskit.providers.models is deprecated since Qiskit 1.2 and will be removed in Qiskit 2.0. With the re...
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!`

Mayank447:
There were several massive changes in Qiskit v2 on the Backend side

- Provider was removed, this decouples the Backend (QiskitCircuitSimulator and QiskitPulseSimulator from provider).
- Configuration dictionary was dropped in `BackendV2`
- target method was added which describes the number of qubits, basis gate set, coherent map etc. This removes the need for the parameter `allow_unitary_gate` as it can be added to the basis set if desired.
- Target method also allows for transpiling the qiskit circuit to the basis gate set, this was done as well in the run method.

There are still a few minor edits left, and documentation update which needs to be done. 

@BoxiLi This resolves #287. Could you please review this PR

### DIFF
--- a/src/qutip_qip/qiskit/backend.py
+++ b/src/qutip_qip/qiskit/backend.py
@@ -12,8 +12,6 @@ from qiskit.transpiler.target import Target
 from qutip_qip.qiskit import Job
 from qutip_qip.qiskit.utils import QUTIP_TO_QISKIT_GATE_MAP
 
-DEFAULT_BASIS_GATE_SET = QUTIP_TO_QISKIT_GATE_MAP.keys()
-
 
 class QiskitSimulatorBase(BackendV2):
     """
@@ -49,11 +47,11 @@ class QiskitSimulatorBase(BackendV2):
         self,
         name: str,
         description: str,
-        version: str = "0.1",
-        num_qubits: int = 10,
-        basis_gates: list[str] = list(DEFAULT_BASIS_GATE_SET),
-        max_shots: int = 1e6,
-        max_circuits: int = 1,
+        version: str,
+        num_qubits: int,
+        basis_gates: list[str],
+        max_shots: int,
+        max_circuits: int,
     ):
         super().__init__(
             name=name,
@@ -114,6 +112,8 @@ class QiskitSimulatorBase(BackendV2):
             `num_qubits`, `qubit_properties`, `basis_gates`,
             `concurrent_measurements` etc.
         """
+
+        DEFAULT_BASIS_GATE_SET = QUTIP_TO_QISKIT_GATE_MAP.keys()
 
         if basis_gates is not None:
             for gate in basis_gates:
--- a/src/qutip_qip/qiskit/circuit_simulator.py
+++ b/src/qutip_qip/qiskit/circuit_simulator.py
@@ -3,6 +3,7 @@ from qutip import basis
 from qutip_qip.circuit import CircuitResult, QubitCircuit
 from qutip_qip.qiskit import QiskitSimulatorBase
 from qutip_qip.qiskit.utils import (
+    QUTIP_TO_QISKIT_GATE_MAP,
     convert_qiskit_circuit_to_qutip,
     sample_shots,
 )
@@ -11,6 +12,8 @@ from qiskit import QuantumCircuit, transpile
 from qiskit.result import Result
 from qiskit.result.models import ExperimentResult, ExperimentResultData
 from qiskit.quantum_info import Statevector
+
+DEFAULT_BASIS_GATE_LIST = list(QUTIP_TO_QISKIT_GATE_MAP.keys())
 
 
 class QiskitCircuitSimulator(QiskitSimulatorBase):
@@ -48,7 +51,7 @@ class QiskitCircuitSimulator(QiskitSimulatorBase):
     def __init__(
         self,
         num_qubits: int = 10,
-        basis_gates: list[str] | None = None,
+        basis_gates: list[str] = DEFAULT_BASIS_GATE_LIST,
         max_shots: int = 1e6,
         max_circuits: int = 1,
         name: str = "circuit_simulator",
--- a/src/qutip_qip/qiskit/pulse_simulator.py
+++ b/src/qutip_qip/qiskit/pulse_simulator.py
@@ -5,6 +5,7 @@ from qutip_qip.circuit import QubitCircuit
 from qutip_qip.device import Processor
 from qutip_qip.qiskit import QiskitSimulatorBase
 from qutip_qip.qiskit.utils import (
+    QUTIP_TO_QISKIT_GATE_MAP,
     convert_qiskit_circuit_to_qutip,
     get_probabilities,
     sample_shots,
@@ -14,6 +15,8 @@ from qiskit import QuantumCircuit
 from qiskit.result import Result
 from qiskit.result.models import ExperimentResult, ExperimentResultData
 from qiskit.quantum_info import Statevector, DensityMatrix
+
+DEFAULT_BASIS_GATE_LIST = list(QUTIP_TO_QISKIT_GATE_MAP.keys())
 
 
 class QiskitPulseSimulator(QiskitSimulatorBase):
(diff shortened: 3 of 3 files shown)


BoxiLi:
Thanks! I haven't check all the details but the overall changes look good to me. We need to add new tests though to test the new backend.

BoxiLi:
Ah wait, it is already there. I missed the second part of the changes.

tests/test_qiskit.py
```diff
@@ -1,277 +1,258 @@
...
+
+        assert_allclose(qutip_sv, qiskit_sv)
+
+    def _run_pulse_processor(self, processor, circ):
+        qutip_backend = QiskitPulseSimulator(processor)
+        qutip_job = qutip_backend.run(circ)
+        return qutip_job.result()
+
+    def _init_pulse_test(self):
+        random.seed(1)
```
BoxiLi: Out of curiosity, the pulse simulation in qutip will actually return slightly different state verctors everytime (mostly because of numerical errors but the system Hamiltonian may also introduce deviation). So it seems that even if that is the case, as long as a random number is fixed, the final counts will be identical?

src/qutip_qip/qiskit/backend.py
```diff
@@ -278,174 +101,65 @@ def _default_options(cls):
...
+            Returns the ``Counts`` object sampled according to
+            the given probabilities and configured shots.
         """
-        zero_state = self.processor.generate_init_processor_state()
+        weights = []
+        for p in count_probs.values():
+            if hasattr(p, "item"):
+                weights.append(float(p.item()))  # For multiple choice
+            else:
+                weights.append(float(p))  # For a trivial circuit with output 1
```
BoxiLi: What does this mean? ` # For a trivial circuit with output 1` Only in this case is p a float number?

Mayank447: I had that trivial case (only one output state  with probability 1 e.g. |10> after applying X1 on |00>) because p.item() doesn't work on non-sequential datatypes

tests/test_qiskit.py
BoxiLi: Sorry, when I saied "everytime" I actually meant "different processor"

Mayank447: These tests are actually the same as previously coded by Shreyas Pradhan. Although `BackendV2` completely changes everything internally, I tried keeping the api interface most the same. Even I was surprised by this when I was going reading through these tests for the first time.

One possible explanation I found was "noise" or deviation different processor introduce is smaller than the "gap" between the random numbers. If done for larger number of shots, I believe the probability counts will slightly begin to differ.

Mayank447:
This PR is complete minus Documentation, Docstring update left.

@BoxiLi One interesting thing that qiskit `BackendV2` introduces is ability to pass list of qiskit circuits to the run function of backend. In the backend obviously we configure `max_circuits` which refers to the maximum no. of circuits which can be passed in a single job. Currently in run method of the Circuit, Pulse simulator

`
def run(self, run_input: list[QuantumCircuit], **run_options) -> Job:
    ....
    
    job = Job(
            backend=self,
            job_id=job_id,
            circuit=run_input,
        )
        job.submit()
        return job
`

We can potentially asynchronize this `job.submit()` call using threads. Do you want to pursue this route? If yes, I would probably want to do it in another PR, since this PR is already huge.

BoxiLi:
> We can potentially asynchronize this job.submit() call using threads. Do you want to pursue this route? If yes, I would probably want to do it in another PR, since this PR is already huge.

this could be interesting but I'm not sure if threads is the right choice. Since the is computation limited, not IO, multiprocessing seems more relevant here? This will be interesting for the use case with multiple small (number of qubits) but deep circuits.

I think there are plans to implement this systematically at the qutip level in the long term, but that will take quite some time to procede.

Mayank447:
@BoxiLi Docstring and Tutorial documentation updation done.

src/qutip_qip/qiskit/converter.py
```diff
@@ -152,21 +147,4 @@ def convert_qiskit_circuit_to_qutip(
...
-
-            unitary = np.array(Operator(qiskit_instruction))
-            qutip_circuit.user_gates[qiskit_instruction.name] = (
-                _make_user_gate(unitary, qiskit_instruction)
-            )
-            qutip_circuit.add_gate(
-                qiskit_instruction.name,
-                targets=_get_mapped_bits(qiskit_qregs, bit_map=qubit_map),
-            )
-
```
BoxiLi, line 192: I am not against removing it. I just forgot how and why this part exited before. This is no longer needed? Or no longer supported?

.gitignore
```diff
@@ -3,6 +3,10 @@ __pycache__/
 *.py[cod]
 *$py.class
 
+# Html and htm files
+*.html
+*.htm
+
```
BoxiLi, line 9: Are those intended?

src/qutip_qip/qiskit/converter.py
Mayank447: This part essentially converts an arbitrary qiskit gate to unitary Matrix and uses that in circuit simulation.

Since BackendV2 allows for defining the basis gate set, I am transpiling the input circuit to the basis set. This removes the need since that arbitrary qiskit gate will anyways be decomposed into the basis set.

.gitignore
Mayank447: Yes.

I didn't find any html files in the package. When I built the documentation on my local system, git was tracking the built documentation html files. Just to ensure those are not tracked in future, added these

Mayank447:
@BoxiLi This PR is ready to be merged. After your previous review I made few minor changes. There were minor bugs giving warning (resolved those) and pytests with mark `pytest.mark.repeat()` weren't being repeated actually since it requires a package `pytest-repeat` added that in the `test.yaml`

BoxiLi:
Thank you for this great update! @Mayank447 

BTW, are you interested in participating in the Google Summer of Code 2026? QuTiP applies for projects every year and the student spends 3 month coding and gets stipend from Google. See previous projects here: https://github.com/qutip/qutip/wiki. We usually post some topics in Feb but people are also free to propose their own interest. The only thing is that it is like a "full time" job between middle May and August, but with flexibility.

Mayank447:
Hi, I am aware about GSoC. I did it in 2023 with Open Chemistry. I am not sure whether I will do GSoC this year because of other commitments, but I will contribute to qutip irrespective.

Issue #292, opened 2025-12-10
Update to QASM 3.0


Mayank447:
`qutip-qip` currently supports reading and converting circuits to QASM 2.0

With QASM 3.0 being released in 2021 and long being stable, the `src/qutip-qip/qasm/` module should be updated to support QASM 3.0 which btw has backward compatability with QASM 2.0

PR #295, opened 2025-12-20
Qasm refactor

Mayank447:
Pending `qasm-refactor` from #290. There are ~couple~ lots of pre-existing error in the `qasm_processor.py` file. Couple of variables are undefined e.g. `curr_gate`, `reg_num` and some are unused e.g. `prev_token`, `gate_added` (not sure whether some logic is missing or they are simply not needed).

@BoxiLi I haven't looked at the conversion logic properly as my first aim was to refactor the 1200 line single `qasm.py` file. I believe instead of fixing these errors as per OpenQASM 2.0 circuit conversion, we can directly upgrade to OpenQASM 3.0 (in accordance with #292) in this or another PR. Since OpenQASM 3.0 supports a lot of new features like loops, extern etc. the conversion logic would change anyways.


BoxiLi:
Yes I agree there are some qutie complicated logic in qasm.py and many potential bugs. Basically we implemented our own parser for a qasm file. I think there should be a smarter way of doing it.

I am  surley happy to upgrade to qasm 3.0, also not against completely rewrite it if you find it a better approach. Would be good to keep the support of 2.0 if possible, because some platforms still only supports that. But this is open for discussion.

Mayank447:
Cool, I will work on resolving the bugs in openqasm 2. Unfortunately there is no parser for qasm 2 which converts `.inc` to an AST, so we will need to maintain a custom parser.

For qasm3 `openqasm` provides a python package for parsing, so that should make it easier to implement.

Mayank447:

This PR mainly split the monolithic `qasm.py` into tokenizer, converter, reader, writer. Fixing some of the bugs in converter is still left but I will do that in a separate PR since some of the conversion logic for QASM2 is missing and also the convertor structure as a whole can be made flexible so that adding support newer QASM versions becomes easier. Also I have added typehints.

src/qutip_qip/qasm/read_qasm.py
```diff
@@ -0,0 +1,78 @@
...
+    qasm_lines = [line.strip() for line in qasm_lines]
+    qasm_lines = list(filter(lambda x: x[:2] != "//" and x != "", qasm_lines))
+    # QASMBench Benchmark Suite has lines that have comments after instructions.
+    # Not sure if QASM standard allows this.
+    for i in range(len(qasm_lines)):
+        qasm_line = qasm_lines[i]
+        loc_comment = qasm_line.find("//")
+        if loc_comment >= 0:
+            qasm_line = qasm_line[0:loc_comment]
+        qasm_lines[i] = qasm_line
```
BoxiLi, line 60: Maybe we can merge the logic for removing comments?

First remove comments and then remove empty lines.

```
qasm_lines = [
    line.split("//", 1)[0].strip()
    for line in qasm_lines
]
qasm_lines = [line for line in qasm_lines if line]
```

src/qutip_qip/qasm/write_qasm.py
```diff
@@ -0,0 +1,234 @@
...
+        """
+        Return QASM gate name for corresponding QuTiP gate.
+
+        Parameters
+        ----------
+        gate_name: str
+            QuTiP gate name.
+        """
+
+        if gate_name in self.gate_name_map:
```
BoxiLi, line 195: It was like this before, but this can be simplified to 
`return gate_name_map.get(gate_name, None)`

BoxiLi:
Don't have to resolve the two comments of mine if you think they should be addressed in the next PR. Just give a signal and I can merge.

src/qutip_qip/qasm/write_qasm.py
Mayank447: Actually there is no need for return None in code. Python by default returns None for functions which don't return anything explicitly

Mayank447:
I will resolve these two comments in the next PR. In general some of the parsing logic needs to be reordered and made modular.

BoxiLi (approved):

PR #298, opened 2026-01-01
Pulse and noise refactor

Mayank447:
Pending refactor from #290 

- Refactored `pulse.py` and `noise.py` to their own separate dir, no API change
- Added typehints

### DIFF
--- a/src/qutip_qip/noise/zzcrosstalk.py
+++ b/src/qutip_qip/noise/zzcrosstalk.py
@@ -7,7 +7,7 @@ class ZZCrossTalk(Noise):
     """
     An always-on ZZ cross talk noise with the corresponding coefficient
     on each pair of qubits.
-    The operator acts only one the lowerest two levels and
+    The operator acts only one the lowest two levels and
     is 0 on higher level.
     Equivalent to ``tensor(sigmaz(), sigmaz())``.
 
--- a/src/qutip_qip/pulse/drift.py
+++ b/src/qutip_qip/pulse/drift.py
@@ -1,5 +1,5 @@
 from qutip import QobjEvo, Qobj
-from qutip_qip.pulse.evo_element import EvoElement, merge_qobjevo
+from qutip_qip.pulse.evo_element import _EvoElement, _merge_qobjevo
 
 
 class Drift:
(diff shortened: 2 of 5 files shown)


Mayank447:

BoxiLi:
Thanks for adding typehints. Are there particular advantages of spliting pulse.py and noise.py into many small files? The eixsting file is not very large.

Mayank447:
> Thanks for adding typehints. Are there particular advantages of spliting pulse.py and noise.py into many small files? The eixsting file is not very large.

The primary reason is for long term maintainability of the module and also making it easier for new contributors to navigate the project. Additionally as we add new noise models like SPAM (State preparation and Measurement), Leakage or pulses like Chirp pulses etc. at some point the split would become necessary.

For the performance point of view, lazy loading is faster. As I don't have to in import all the noise models (and their dependencies) if I only need RelaxationNoise

BoxiLi:
Interesting, now I can ask copilot to do the review. Not all the coments are relevant but it does find something.

I think the Variable in ZZ crosstalk should be address separately since this PR is about moving files and type hints. But the typehint of `tlist` should indeed be a list or more commonly a numpy array. About `_Evoelement` and `_merge_qobjevo`, Those should not be exposed to the users. Why removing the underscore?

Mayank447:
@BoxiLi Any idea about the failing test cases for `test_devices.py` and `test_vqa.py`. I also see the same tests failing for #300 which was recently merged, although before merging everything was passing by CI/CD.

On my local machine, the test cases pass (python 3.12.8)

*Edit:* I believe it has something to do with numpy on CI/CD. It is using `numpy 2.4.5`

BoxiLi:
`numpy 2.4.5`? I see the failing tests are using `numpy 2.4.1` and the old passing tests are using `numpy 2.4.0`.

This will be a bit hard to debug I think. Something in the new numpy must have triggered/caused a bug. I cannot see the reason immediately from the changelog https://numpy.org/devdocs/release/2.4.1-notes.html

In any case, this is a different issue, if you find this PR ready for merge, I will merge.

Mayank447:
Sorry I meant `2.4.1`, I will try to debug and resolve this issue in a different PR.

This PR is ready to be merged.

PR #300, opened 2026-01-01
Refactored circuit draw

Mayank447:
Last pending refactor from #290 

- Refactor the circuit rendering/draw logic .py files to a `src/circuit/draw/` dir
- Changed %s string to f-strings with raw support for easier readability
- Added missing typehints

### DIFF
--- a/src/qutip_qip/circuit/draw/base_renderer.py
+++ b/src/qutip_qip/circuit/draw/base_renderer.py
@@ -49,7 +49,7 @@ class StyleConfig:
         Align the layers of the gates across different wires. The default is False.
         Available to TextRender and MatRender.
 
-    theme : Optional[Union[str, Dict]], optional
+    theme : Union[str, Dict], optional
         Color theme of the circuit. The default is "qutip".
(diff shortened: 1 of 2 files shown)


Mayank447:

tests/test_renderer.py
```diff
@@ -2,7 +2,7 @@
 import numpy as np
 from unittest.mock import patch
 from qutip_qip.circuit import QubitCircuit
-from qutip_qip.circuit.text_renderer import TextRenderer
+from qutip_qip.circuit.draw import TextRenderer
```
BoxiLi, line 5: Out of curiosity, do you think it will be more convenient to include everything of  `draw` into `qutip.circuit` so that one can simpliy import everything from `qutip_qip.circuit`? Although the most common user interface is `QubitCircuit.draw(...)`

Mayank447: I wouldn't expose everything in `qutip_qip.circuit` given that we already have a `QubitCircuit.draw()` method. Since most people would actually be using the `.draw()` method

PR #302, opened 2026-01-09
Project specification in only pyproject toml

### LINKED ISSUE #294: Use only pyproject.toml for project configuration
Currently there is `pyproject.toml`, `requirements.txt`, `setup.py` and `setup.cfg` in the base directory (this is the older way to specify a project configuration, dependencies, tooling and build functions). As per [PEP 518](https://peps.python.org/pep-0518/) and [PEP 621](https://peps.python.org/pep-0621/) standard these should be done only in pyproject.toml serving as the single source of truth.

This would remove the redundancies of changing the package dependencies at several places and also declutter the package, tool dependencies.

Mayank447:
This PR fixes #294 

**Changes:**
- Moved project dependencies, metadata and build config from `setup.cfg` to `pyproject.toml`
- Removed `requirements.txt`
- Removed `setup.py` (it was only creating `/src/qutip_qip/version.py` by reading `VERSION` file and adding git hash) which was instead automated using `setuptools-git-versioning` in pyproject.toml and`metadata.version()` from standard `importlib` library.


Mayank447:

setup.py
```diff
@@ -1,99 +0,0 @@
...
-            git_hash = git_out.stdout.decode(sys.stdout.encoding).strip()
-            version_string += git_hash or "nogit"
-        # CalledProcessError is for if the git command fails for internal
-        # reasons (e.g. we're not in a git repository), OSError is for if
-        # something goes wrong when trying to run git (e.g. it's not installed,
-        # or a permission error).
-        except (subprocess.CalledProcessError, OSError):
-            version_string += "nogit"
-    options['version'] = version_string
-    return options
```
BoxiLi, line 71: This part actually does more than just the standard version encoding. It adds the corresponding git hash and additional version strings like "dev".

Personally I don't have a strong preference, but it exists across different qutip packages.

Mayank447: We can actually do this using `setuptools-git-versioning` in pyproject.toml. I didn't use it is because `test_family.py` isn't checking for git hashing.

Mayank447: If we using `dev` in versioning with git hash, we will need to set the new version in `VERSION` after every release like set `0.5.0` after release `0.4.1`.

If we use `post`in versioning with git hash, we will only need to set the new `VERSION` during the release, so we have the flexibility of `0.4.2` and `0.5.0`. @BoxiLi which would you prefer?

BoxiLi: Persontally I prefer `dev` since it indicates `working toward the next version`.

Mayank447: Cool, the commit for `dev` has already been pushed. You can review this PR

BoxiLi (approved):

PR #306, opened 2026-01-13
Refactor gateclass

### LINKED ISSUE #263: Refactor the Gate class objects
Currently, all QuTiP gate classes inherit from the Gate class, which has a long list of input parameters (e.g., control qubit, control value, etc.). To define a new simple gate, one always has to add `**kwargs` to be compatible with the `Gate` class. This makes it very hard to document and check if all the input is valid.

The gate class should be refactored such that elementary gates like X Y Z only accept necessary input parameters explicitly, avoiding **kwargs for unrelated arguments. Advanced gates should take those simple gates as one of the input parameters (see also other libraries like qiskit). For instance, the `ControlledGate` takes the unitary gate to be controlled, and the control qubit index and control values as input. This simplifies the definition of new gates and makes the type check much clearer.

Mayank447:
Fixes #263 

*Changes (until now):*
- Refactor `gateclass.py`
- Added a `GLOBALPHASE` Gate
- Removed usage of object declaration using`Gate` class in `_decompose.py`, instead replaced with existing gates e.g. `X`, `CNOT` etc.

### DIFF
--- a/doc/source/qip-basics.rst
+++ b/doc/source/qip-basics.rst
@@ -33,11 +33,11 @@ A circuit with the various gates and registers available is demonstrated below:
 .. testcode::
 
   from qutip_qip.circuit import QubitCircuit
-  from qutip_qip.operations import Gate
+  from qutip_qip.operations import Gate, SWAP
   from qutip import tensor, basis
 
   qc = QubitCircuit(N=2, num_cbits=1)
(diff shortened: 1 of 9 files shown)


Mayank447:
Clsoing this PR as superseeded by #348

PR #307, opened 2026-01-15
Remove gate class direct use

Mayank447:
- Removed direct usage of Gate class completely.
- Added a `GLOBALPHASE`, `IDLE` Gate
- Refactored `gateclass.py`
- Removed dependency of standard gates on `gates.py` for `get_compact_qobj` method
- Removed `get_compact_qobj` implementation from `Gate` class, only function defination remains.

### DIFF
--- a/doc/pulse-paper/customize.py
+++ b/doc/pulse-paper/customize.py
@@ -25,7 +25,7 @@ from qutip import (
     fock_dm,
 )
 from qutip_qip.circuit import QubitCircuit
-from qutip_qip.operations import Gate
+from qutip_qip.operations import Gate, GLOBALPHASE
 from qutip_qip.device import ModelProcessor, Model
 from qutip_qip.compiler import GateCompiler, Instruction
 from qutip_qip.noise import Noise
@@ -86,6 +86,7 @@ class MyCompiler(GateCompiler):
             Instruction (qutip_qip.compiler.instruction.Instruction): An instruction
             to implement a gate containing the control pulses.
         """
+
         pulse_info = [
             # (control label, coeff)
             ("sx" + str(gate.targets[0]), np.cos(phase) * coeff),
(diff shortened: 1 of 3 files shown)


Mayank447:

I ran `black` (previously I used to manually fix the linting errors since the number was small). Because of black quite a few unrelated files also got formatted. So it would be actually easier to review this PR before the run black commit or to do it commit-wise.

BoxiLi:
You can run black for a specific directory if that helps, or select to only add certain files to a git commit. But this is ok. I will go over commits.

BoxiLi:
> [Moved all the standard gates to std_gates.py](https://github.com/qutip/qutip-qip/pull/307/commits/6c4a6aecf2b14ef35544f6fe2889fa486a35347a)

Out of curiosity, what makes you spliting them into different files and then merging back to std_gates.py

src/qutip_qip/operations/gateclass.py
```diff
@@ -238,85 +207,12 @@ def get_compact_qobj(self):
...
-        elif self.name == "IDLE":
-            qobj = qeye(2)
-        elif self.name == "GLOBALPHASE":
-            raise NotImplementedError(
-                "Globalphase gate has no compack qobj representation."
-            )
-        else:
-            raise NotImplementedError(f"{self.name} is an unknown gate.")
-        return qobj
+        pass
```
BoxiLi, line 209: We could make this an `@abstractmethod`

doc/pulse-paper/qft.py
```diff
@@ -23,6 +23,7 @@
 state1 = qc.run(basis([2] * num_qubits, [0] * num_qubits))
 # Pulse-level simulation
 processor = LinearSpinChain(num_qubits)
+qc.gates = [g for g in qc.gates if g.targets is not None]
```
BoxiLi: Why do we now need this additional line? I feel like this should be handled internally. Is this because of GLOBALPHASE?

BoxiLi (commented):
Looks good, only some minor comments.

doc/pulse-paper/qft.py
Mayank447: Oh yes, this should have been removed as `GLOBALPHASE` is now handled internally. p
Previously added when I was trying to debug the failing tests.

src/qutip_qip/operations/gateclass.py
Mayank447: The plan is to make `Gate` an abstract class and yes `get_compact_qobj` should be an `@abstractmethod`. But making `Gate` an `ABC` class requires some deeper changes. So this is what we are going to do in the next PR

Mayank447:
> Out of curiosity, what makes you spliting them into different files and then merging back to std_gates.py

Well, I wanted to separate the standard gates from `gatesclass.py`. That's the reason for splitting. I wasn't sure about the nomenclature we should follow to split the standard gates into files - single, two qubits, special or parmaterized, controlled, multicontrolled etc. That's why merged them

doc/pulse-paper/qft.py
Mayank447: This has been fixed

BoxiLi:
> Well, I wanted to separate the standard gates from gatesclass.py. That's the reason for splitting. I wasn't sure about the nomenclature we should follow to split the standard gates into files - single, two qubits, special or parmaterized, controlled, multicontrolled etc. That's why merged them

Haha, I see, there also have some crossovers. Like CNOT can be a controlled gate and a two-qubit gates.

BoxiLi:
LGTM, should I merge?

Mayank447:
Yes

Issue #296, opened 2025-12-22
Depreciate SNOT and replace with H gate


Mayank447:
Every library uses `H` to represent Hadamard gate. Is there a particular reason `SNOT` is being used?

If not, ideally H should be added to operation (which it already is in the Gate class) and SNOT depreciated.

BoxiLi:
No, only historical reasons. Go for it!

Mayank447:
Closing this as superseded by #310

PR #309, opened 2026-01-16
Make Gate an abstract class

Mayank447:
*Changes:*
- Made `Gate` an abstract class and `get_compact_obj` an abstract method.
- Added `@staticmethod` for `get_compact_qobj` all non-parametrized standard gates.
- Removed all the unneccsary arguments from `Gate` class, it only takes targets, classical_control, class_control_value and style now. Removing these as well would be done in the next PR. 
- Removed `**kwargs` completely from all the gates and gate classes.
- Added `ParametrizedGate`, `ControlledParamGate` classes.
- Made `target_gate` in ControlledParamGate be initialized with arg_value like RX in CRX. 
- Made `latex_str` a class attribute instead of an object attribute for all the gates.
- Minor fixes to qpe, qiskit backend and vqa which were causing tests to fail

- Removed `user_gates` (this raises an error pointing to new way to define custom gates), `_get_gate_unitary()` method (since it was internal method no error, depreciation) from `QubitCircuit`.
- Restricted `add_gate()` to only accept `Gate` object, subclasses of it, `str` for standard gates.
- Added `CH`, `CQASMU` to standard gates and `CustomGate` which wraps a passed unitary (Qobj)

Things to keep track of:
- Need to update the newly added gates to documentation.
- Document how new gates can added instead of previous `user_gates` input.
- In VQA, `U=block.get_unitary(current_params)` saves the full unitary instead of a gate instance. Once gate refactor is merged to master, a GitHub issue should be created for the same.


Mayank447:
@BoxiLi This PR is ready for review. I will fix the docs towards the end of this refactor in a separate PR.

Mayank447:
@BoxiLi This PR is complete from my end.

doc/source/qip-basics.rst
```diff
@@ -208,14 +208,11 @@ gate function returning a :class:`qutip.Qobj` and save it in the attribute ``use
...
-      qc.user_gates = {"CTRLRX": user_gate1,
-                       "S"     : user_gate2}
-
       # qubit 0 controls qubit 1
-      qc.add_gate("CTRLRX", targets=[0,1], arg_value=np.pi/2)
+      qc.add_gate("CRX", controls=[0], targets=[1], arg_value=np.pi/2)
 
       # qubit 1 controls qubit 0
-      qc.add_gate("CTRLRX", targets=[1,0], arg_value=np.pi/2)
+      qc.add_gate("CRX", controls=[1], targets=[0], arg_value=np.pi/2)
```
BoxiLi, line 215: What we are demonstrating here is how to define a custom gate, but the user_gate part is removed. This example seems to lost its perpose then?

tests/test_circuit.py
```diff
@@ -364,6 +368,7 @@ def test_add_measurement(self):
         assert qc.gates[2].name == "TOFFOLI"
         assert qc.gates[4].classical_controls == [0, 1]
 
+    @pytest.mark.skip(reason="Changing the interface completely")
```
BoxiLi, line 371: This will be address in future PR?

tests/test_circuit.py
```diff
@@ -701,10 +711,10 @@ def test_latex_code_teleportation_circuit(self):
                 r"  &  \qw  &  \qw \cwx[2]  &  \ctrl{1}  &  \qw  & \qw \\ ",
                 r" & \lstick{\ket{0}} &  \qw  &  \targ  &  \qw  &  \qw"
                 r"  &  \qw  &  \qw  &  \gate{X}  &  \gate{Z}  & \qw \\ ",
-                r" & \lstick{\ket{0}} &  \gate{{\rm H}}  &  \ctrl{-1}  &"
+                r" & \lstick{\ket{0}} &  \gate{SNOT}  &  \ctrl{-1}  &"
                 r"  \targ  &  \qw  &  \qw  &  \meter &  \qw  &  \qw  & \qw \\ ",
                 r" & \lstick{\ket{q0}} &  \qw  &  \qw  &  \ctrl{-1}  &"
-                r"  \gate{{\rm H}}  &  \meter &  \qw  &  \qw  &  \qw  & \qw \\ ",
+                r"  \gate{SNOT}  &  \meter &  \qw  &  \qw  &  \qw  & \qw \\ ",
```
BoxiLi, line 717: It would be better to use H as the default symbol since it is the most widely used one.

tests/test_gates.py
```diff
@@ -421,59 +421,59 @@ def test_gates_class():
...
+    circuit1.add_gate("Y", targets=1)
+    circuit1.add_gate("Z", targets=2)
+    circuit1.add_gate("RX", targets=0, arg_value=np.pi / 4)
+    circuit1.add_gate("RY", targets=0, arg_value=np.pi / 4)
+    circuit1.add_gate("RZ", targets=1, arg_value=np.pi / 4)
+    circuit1.add_gate("H", targets=0)
+    circuit1.add_gate("SQRTNOT", targets=0)
+    circuit1.add_gate("S", targets=2)
+    circuit1.add_gate("T", targets=1)
     circuit1.add_gate("R", 1, arg_value=(np.pi / 4, np.pi / 6))
```
BoxiLi, line 434: Does this change means that we now force user to explicitly write `circuit1.add_gate("H", targets=0)`? Will `circuit1.add_gate("H", 0)` still work?

tests/test_qiskit.py
```diff
@@ -46,7 +47,7 @@ def _compare_args(self, req_gate, res_gate):
                 or type(res_gate.arg_value) is tuple
                 else [res_gate.arg_value]
             )
-            if res_gate.arg_value
+            if isinstance(res_gate, ParametrizedGate) and res_gate.arg_value
```
BoxiLi, line 50: If it is already checked taht `res_gate` is a `ParametrizedGate`. Is it sitll nessensary to check that `res_gate.arg_value` is not None?

src/qutip_qip/circuit/circuit.py
```diff
@@ -73,7 +63,6 @@ def __init__(
         input_states=None,
         output_states=None,
         reverse_states=True,
-        user_gates=None,
```
BoxiLi, line 76: Maybe we should keeps this for now add a deprecation? Depending on complicated it is, we can 
- supprot user_gate for now, define a gate class internally and raise a warning
- drop user_gate, but at least raise an error pointing people to the right way of using gate class.

src/qutip_qip/circuit/circuit.py
```diff
@@ -232,21 +210,64 @@ def add_gate(
...
+                    style=style,
+                )
+
+            else:
+                gate = gate_class(
+                    targets=targets,
+                    classical_controls=classical_controls,
+                    classical_control_value=classical_control_value,
+                    style=style,
+                )
```
BoxiLi, line 277: Will it be better if we define some sort of `kwarg` and remove all the input key whose value is `None`? And then feed it to the gate_class. In this way maybe we can get rid of the if branching?

src/qutip_qip/circuit/circuit.py
BoxiLi: Please also write it (deprecation or error, and the new API) also in the PR description so that I will remember to highlight in the release description

src/qutip_qip/circuit/circuit.py
```diff
@@ -861,33 +874,6 @@ def propagators(self, expand=True, ignore_measurement=False):
...
-                    )
-            elif isinstance(func_or_oper, Qobj):
-                qobj = func_or_oper
-            else:
-                raise ValueError("gate is neither function nor operator")
-        else:
-            qobj = gate.get_compact_qobj()
-        return qobj
-
     def compute_unitary(self):
```
BoxiLi, line 883: Great! we can finally get rid of this.

src/qutip_qip/circuit/circuit.py
```diff
@@ -310,7 +331,7 @@ def add_1q_gate(
         for q in qubits:
             self.add_gate(name, targets=q, **kwargs)
 
-    def add_circuit(self, qc, start=0, overwrite_user_gates=False):
+    def add_circuit(self, qc, start=0):
         """
         Adds a block of a qubit circuit to the main circuit.
         Globalphase gates are not added.
```
BoxiLi: > Globalphase gates are not added.

This this still true?

src/qutip_qip/compiler/instruction.py
```diff
@@ -76,4 +77,6 @@ def controls(self):
 
         :type: list
         """
-        return self.gate.controls
+        if isinstance(self.gate, ControlledGate):
+            return self.gate.controls
+        return None
```
BoxiLi, line 82: I see now that in several places we have to import some abstract gate class and check the class type before calling something like `gate.controls`.

Do you think it is better like this or better to give some default None value for those attributes? Same for `arg_values`.

src/qutip_qip/qasm.py
```diff
@@ -33,33 +28,33 @@ def __init__(self, name, gate_args, gate_regs):
...
+#     def cu3(args):
+#         return controlled_gate(qasmu_gate(args))
 
-    def ch():
-        return controlled_gate(snot())
+#     def ch():
+#         return controlled_gate(snot())
 
-    return {"ch": ch, "tdg": tdg, "id": id, "u2": u2, "sdg": sdg, "cu3": cu3}
+#     return {"ch": ch, "tdg": tdg, "id": id, "u2": u2, "sdg": sdg, "cu3": cu3}
```
BoxiLi: Those commented lines? They could be removed I assume

src/qutip_qip/vqa.py
```diff
@@ -140,11 +134,15 @@ def construct_circuit(self, angles):
...
                     n = block.get_free_parameters_num()
-                    circ.add_gate(
-                        block.name,
+
+                    current_params = angles[i : i + n] if n > 0 else []
+                    gate_instance = CustomGate(
+                        name=block.name,
                         targets=list(range(self.num_qubits)),
-                        arg_value=angles[i : i + n] if n > 0 else None,
+                        U=block.get_unitary(current_params),
```
BoxiLi, line 142: So here instead of saving the gate class instance we save the full unitary? This might be fine for small number of qubits, but may take a lot of memory if the size grows. This is not high priority though right now.

BoxiLi (commented):
Sorry for the delay, I have reviewed all files except for `gateclass.py` `std_gates.py` and `qasm.py`. LGTM in general.

doc/source/qip-basics.rst
Mayank447: Yes, infact this entire example needs to be removed actually and rewritten demonstrating how to define custom gates.

tests/test_circuit.py
Mayank447: Yes

tests/test_circuit.py
Mayank447: Actually once Gate refactor is completed. I will replace all instances of "SNOT" with "H" in the codebase and depreciate SNOT (#296).

Earlier in the codebase base `SNOT = H` (this also copied the name H), I replaced this and other redundant gates (#310) by inheriting. This didn't copy the `name`, `latex_str`.

tests/test_gates.py
Mayank447: In general it is good to define the argument as well in the codebase and tests, that's why I did this. `circuit1.add_gate("H", 0)` it will obviously work.

src/qutip_qip/circuit/circuit.py
Mayank447: I would prefer the second, drop `user_gate` and raise an error pointing people to the right way of using gate class.

src/qutip_qip/circuit/circuit.py
Mayank447: Actually we will be removing `targets`, `controls` etc. from Gate classes completely (it would be managed internally by `QubitCircuit`). So this piece of code will be removed actually in the next PR.

src/qutip_qip/circuit/circuit.py
Mayank447: I didn't touch `add_circuit` method at all. But `add_gate()` method is now able to add `GLOBALPHASE` gates. Handling of global phase needs to be done on a circuit level (like an class variable instead of Gates) in the next PR.

src/qutip_qip/compiler/instruction.py
Mayank447: Actually since `gate.controls` is temporary. This shall be removed in next PR.

src/qutip_qip/qasm.py
Mayank447: Yes, these can be removed.

src/qutip_qip/vqa.py
Mayank447: I will open an issue for this for us to keep track.

src/qutip_qip/circuit/circuit.py
BoxiLi: Sure

tests/test_qiskit.py
Mayank447: No we don't need to check. I will make this change

src/qutip_qip/vqa.py
Mayank447: I have added this issue in the PR description. Once gate_refactor branch is merged to master. We can open and resolve this issue.

Mayank447:
@BoxiLi I have made the changes as per suggestions.

src/qutip_qip/operations/gateclass.py
```diff
@@ -306,3 +273,210 @@ def get_compact_qobj(self):
...
+        if self.classical_controls:
+            err_msg = "Exporting controlled gates is not implemented yet."
+            raise NotImplementedError(err_msg)
+        else:
+            qasm_out.output(
+                qasm_out._qasm_str(
+                    q_name=qasm_gate,
+                    q_targets=self.targets,
+                    q_controls=self.controls,
+                    q_args=self.arg_value,
```
BoxiLi, line 431: I hope the _to_qasm function could be further simplified in future PRs?

BoxiLi (approved):
I will merge when you give me a final confirmation

src/qutip_qip/operations/gateclass.py
BoxiLi: Just seems to be some duplications

Mayank447: Infact it can be removed in future PRs as we remove targets, controls from gateclass.

Mayank447:
You can go ahead with merging.

src/qutip_qip/operations/gateclass.py
Mayank447: Wait.. we will still need _to_qasm especially for custom gates but yeah target, controls are not there in gate defination in qasm.

PR #312, opened 2026-01-19
Removing Redundant Gates

### LINKED ISSUE #310: Remove redundant Gates
Many Gates have double instances like (H, SNOT), (CX, CNOT), (CSIGN, CZ) etc. One of the Gate for each pair needs to be removed otherwise it will become harder when doing gate decomposition and transpilation.

mudit06mah:
## Summary
Removing Redundant Gate Pairs

## Issue
Fixes #310 

## Changes
Removed the following gates from GATE_CLASS_MAP (gateclass.py):
SNOT (Redundant with H)
CX (Redundant with CNOT)
CSIGN (Redundant with CZ)

## Note:
Not Ready to Merge Yet, since gates.py still contains redundant gate functions, need to confirm some changes before proceeding


mudit06mah:
Before Proceeding with this PR I wanted to confirm if I should rename the `snot()` function with `h_gate()`.
Also if I should completely remove the other functions that are redundant (hard remove) or deprecate them for now?

BoxiLi:
Hi @mudit06mah , thank you very much for this PR but right now it is probably not the best time to work on the gate module. As you may see there is a large scale refactoring of the gate class on #309. I belive these issues should have already been fixed there. If we merge this now. It will create conflicts when finishing that gate refactoring.

mudit06mah:
@BoxiLi  Understood! No Problem :)

PR #314, opened 2026-01-24
Feat: Add Grover Search algorithm

mudit06mah:
## Summary:
This PR adds Grover's Search Algorithm for amplification.

## Related issue:
This PR Addresses issue #264

## Changes:
- Added file `grover.py` that contains definition for class `OracleGate`  and functions to implement oracle and grover's search algorithim
- Added `test_grover.py` to test the algorithim
- Added `grover` to the `__init__.py`
- Added algorithm to doc

## Note:
LLMs were used to assist in improving the approach, writing test cases and documenting the algorithm

### DIFF
--- a/doc/source/qip-grover.rst
+++ b/doc/source/qip-grover.rst
@@ -10,28 +10,28 @@ It works by iteratively rotating a quantum state vector toward a "marked" soluti
 Overview
 ========
 
-Grover's algorithm is a fundamental quantum algorithm that finds the unique input to a black-box function that produces a particular output value. In ``qutip-qip``, the implementation consists of two main components:
+Grover's algorithm is a fundamental quantum algorithm that finds the unique input to a black-box function that produces a particular output value. The implementation consists of two main components:
 
-* **The Oracle**: A phase oracle that flips the sign of the amplitude for the "marked" or "target" states.
-* **The Diffusion Operator**: Also known as the "inversion about the mean" operator, which amplifies the probability of measuring the marked states.
+* **The Oracle**: A phase oracle that flips the sign of the amplitude for the "marked" states passed by the user.
+* **The Diffusion Operator**: This operator flips the current state about its previous state.
 
 The algorithm requires an optimal number of iterations, approximately :math:`\frac{\pi}{4}\sqrt{\frac{N}{M}}`, where :math:`N=2^n` is the size of the search space and :math:`M` is the number of marked states.
 
 Constructing the Algorithm
 ==========================
 
-In this module, you can construct the full search circuit using the :func:`.grover` function. To make the process easier, a utility function :func:`.grover_oracle` is provided to generate standard phase-flip oracles.
+In this module, you can construct the full Grover search circuit using the :func:`.grover` function. To make the process easier, a utility function :func:`.grover_oracle` is provided to generate standard phase-flip oracles if the marked states are known.
 
 The :func:`.grover` function requires the following:
 
 ====================  ==================================================
 Argument                           Description
 ====================  ==================================================
-``oracle``            A :class:`.QubitCircuit`, :class:`.Gate`, or :class:`~.Qobj` representing the phase oracle.
-``qubits``            List of qubit indices (or integer count) to run the search on.
-``num_solutions``     **Mandatory** integer representing the number of marked states :math:`M`.
+``oracle``            A :class:`.QubitCircuit` or :class:`.Gate` representing the phase oracle.
+``search_qubits``     List of qubit indices (or integer count) to run the search on.
+``num_solutions``     Integer representing the number of marked states :math:`M`.
 ``num_iterations``    Optional integer for manual control over the rotation count.
-``N``                 Optional total number of qubits when searching on a subset of a larger system.
+``num_qubits``        Optional total number of qubits when searching on a subset of a larger system.
 ====================  ==================================================
 
 Example: Searching for Multiple Targets
@@ -50,7 +50,7 @@ Let's simulate a search on 3 qubits (:math:`N=8`) where two states are marked: :
     >>> n_qubits = 3
     >>> marked = [3, 5]
 
-    We then use the utility function :func:`.grover_oracle` to create a phase-flip oracle for our target states.
+    We then use the utility function :func:`.grover_oracle` to create a phase-flip oracle for our marked states.
 
     >>> oracle = grover_oracle(n_qubits, marked)
 
--- a/src/qutip_qip/algorithms/grover.py
+++ b/src/qutip_qip/algorithms/grover.py
@@ -6,32 +6,6 @@ from qutip_qip.operations import Gate, Z, ControlledGate
 __all__ = ["grover", "grover_oracle"]
 
 
-class OracleGate(Gate):
-    """
-    Custom gate that wraps an arbitrary quantum oracle operator.
-    """
-
-    def __init__(self, targets, U, **kwargs):
-        """
-        Initialize an OracleGate.
-
-        Parameters
-        ----------
-        targets : int or list of int
-            The qubit(s) on which the oracle gate acts.
-        U : Qobj
-            The unitary operator representing the oracle.
-        **kwargs
-            Additional keyword arguments to pass to the parent Gate class.
-        """
-        super().__init__(targets=targets, **kwargs)
-        self.U = U
-        self.latex_str = r"Oracle"
-
-    def get_compact_qobj(self):
-        return self.U
-
-
 def grover_oracle(
     search_qubits: int | Sequence[int], marked_states: int | Sequence[int]
 ) -> QubitCircuit:
@@ -104,7 +78,7 @@ def grover_oracle(
 
 
 def grover(
-    oracle: QubitCircuit,
+    oracle: QubitCircuit | Gate,
     search_qubits: int | Sequence[int],
     num_solutions: int,
     num_iterations: int | None = None,
@@ -115,7 +89,7 @@ def grover(
 
     Parameters
     ----------
-    oracle : :class:`~.circuit.QubitCircuit` or :class:`~.operations.Gate` or :class:`qutip.Qobj`
+    oracle : :class:`~.circuit.QubitCircuit` or :class:`~.operations.Gate`
         The oracle that flips the phase of marked states.
     search_qubits : int or sequence of int
         The qubits to run the search on.
@@ -200,9 +174,9 @@ def grover(
         qc.add_gate("SNOT", targets=q)
 
     # Calculate optimal Iterations if none provided:
-    if num_iterations is not None and num_iterations <= 0:
+    if num_iterations is not None and num_iterations < 0:
         raise ValueError(
-            f"num_iterations must be a positive integer, got {num_iterations}."
+            f"num_iterations must not be a negative integer, got {num_iterations}."
         )
 
     if num_iterations is None:
@@ -220,8 +194,6 @@ def grover(
                     f"Oracle gate targets {oracle.targets} are not a subset of algorithm qubits {search_qubits}"
                 )
             qc.add_gate(oracle)
-        else:
-            qc.add_gate(OracleGate(search_qubits, oracle))
 
         # Diffusion (Inversion about the mean)
         for q in search_qubits:
--- a/tests/test_grover.py
+++ b/tests/test_grover.py
@@ -47,12 +47,7 @@ class TestGrover:
         oracle = grover_oracle(2, 3)
 
         with pytest.raises(
-            ValueError, match="num_iterations must be a positive"
-        ):
-            grover(oracle, 2, 1, num_iterations=0)
-
-        with pytest.raises(
-            ValueError, match="num_iterations must be a positive"
+            ValueError, match="num_iterations must not be a negative"
         ):
             grover(oracle, 2, 1, num_iterations=-3)
 
@@ -151,22 +146,3 @@ class TestGrover:
 
         fidelity = abs(psi_final.overlap(psi_expected)) ** 2
         assert fidelity > 0.9999
-
-    def test_grover_oracle_types(self):
-        """Test that the main function accepts Gate and Qobj oracles."""
-        qubits = [0]
-
-        # Qobj Oracle (Z gate for 1 qubit marks state |1>)
-        oracle_qobj = Qobj([[1, 0], [0, -1]])
-        qc_a = grover(oracle_qobj, qubits, 1)
-        assert len(qc_a.gates) > 0
-
-        # Gate Oracle
-        oracle_gate = Gate("Z", targets=[0])
-        qc_b = grover(oracle_gate, qubits, 1)
-        assert len(qc_b.gates) > 0
-
-        # Compare unitaries
-        assert np.allclose(
-            qc_a.compute_unitary().full(), qc_b.compute_unitary().full()
-        )


mudit06mah:
@BoxiLi  I'd like your opinion on this.

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,202 @@
...
+                qc.add_gate("X", targets=qubits[i])
+
+        # Change state by Multi Controlled Z Gate
+        ctrls = qubits[:-1]
+        tgt = qubits[-1]
+
+        ctrl_val = 2 ** len(ctrls) - 1
+
+        if len(ctrls) == 1:
+            qc.add_gate(CSIGN(controls=ctrls, targets=tgt))
```
BoxiLi: Let's just use `add_gate("CSIGN", controls=..., targets=...)` signature here for predefined gates. This will be compatible with the new gate class. THanks!

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,202 @@
...
+    qc = QubitCircuit(max(qubits) + 1)
+
+    # Superposition:
+    for q in qubits:
+        qc.add_gate("SNOT", targets=q)
+
+    # Calculate optimal Iterations if none provided:
+    if num_iterations is None:
+        N = 2**n_qubits
+        num_iterations = int((np.pi / 4) * np.sqrt(N / num_marked_solutions))
```
BoxiLi: This is probably very straightforward, but just for people who read this later, is there a reference?

tests/test_grover.py
```diff
@@ -0,0 +1,126 @@
...
+        Test 3 qubits with 2 marked states: |011> (3) and |101> (5).
+        """
+        n_qubits = 3
+        marked = [3, 5]
+
+        oracle = grover_oracle(n_qubits, marked)
+
+        # For N=8, M=2, theta = 30 deg. One iteration rotates to 90 deg (solution).
+        # Default calculation gives 2 iterations, so we must force 1.
+        qc = grover(oracle, n_qubits, num_iterations=1)
```
BoxiLi: What does this mean? The default calculation gives a "wrong" estimation?

BoxiLi: We should also test at least once the default iteration numbers.

src/qutip_qip/algorithms/grover.py
mudit06mah: @BoxiLi  should I wait for the gate refactor branch to merge with the main branch first?

tests/test_grover.py
mudit06mah: I did this as a temporary fix because it was estimating the no. of iterations for the search to be 2, I'll look into it again and  update the test case.

mudit06mah:
@BoxiLi  Have you tested this algorithm? Please let me know if you want me to change something (other than the review you have already given)

src/qutip_qip/algorithms/grover.py
Mayank447: `add_gate` method already supports `targets`, `controls` as arguments. Defining targets, controls within gate class won't be allowed after gate refactor

src/qutip_qip/algorithms/grover.py
Mayank447: Also instead of int this should be greatest integer function i.e. np.floor() for optimal number of iterations

BoxiLi:
Sorry for the delay! I tried it, and it worked nicely.

I used the following code, modified from your tests
```
n_qubits = 3
marked = [1, 5]

oracle = grover_oracle(n_qubits, marked)

qc = grover(oracle, n_qubits, num_iterations=None)

U_grover = qc.compute_unitary()
psi0 = tensor([basis(2, 0)] * n_qubits)
psi_final = U_grover * psi0

prob = 0
for i in marked:
    prob_i = abs(psi_final.overlap(basis(2**n_qubits, i))) ** 2
    prob += prob_i
print(prob)
```

The only thing I noticed is that sometimes when `marked` includes more than one index. The result may show quite low probability sometimes. I am not sure if it is because of a wrong calculation of `num_iterations` or because Grover does not guarantee the result with multiple marked indices.

src/qutip_qip/algorithms/grover.py
BoxiLi: I think we can merge it even before the gate refactor, just code it in a compatible way. There must be a way of doing that. Otherwise we should do the gate refactor in a more compatible way.

mudit06mah:
> The only thing I noticed is that sometimes when `marked` includes more than one index. The result may show quite low probability sometimes. I am not sure if it is because of a wrong calculation of `num_iterations` or because Grover does not guarantee the result with multiple marked indices.

This issue arised due to not not using `np.floor()` while calculating `num_iterations`, also the `num_marked_solutions` (variable name is now changed to `num_solutions`) should not be optional, otherwise it will default to 1, which might not be correct.
I have changed the algorithm to round off using `np.floor()` and also to make the arg `num_solutions` to be mandatory. I have also made other changes suggested by @Mayank447.
The algorithm should now work pretty neatly :)

doc/source/qip-grover.rst
```diff
@@ -0,0 +1,79 @@
...
+        >>> prob = abs(psi_final.overlap(basis(8, 3)))**2 + abs(psi_final.overlap(basis(8, 5)))**2
+        >>> print(f"Success Probability: {prob:.4f}")
+        Success Probability: 1.0000
+
+.. plot::
+    :context: close-figs
+
+    from qiskit.visualization import plot_histogram
+    # ... (after running simulation)
+    # plot_histogram(counts)
```
Mayank447: Remove these

doc/source/qip-grover.rst
```diff
@@ -0,0 +1,79 @@
+.. _qip_grover:
+
+**************************
+Grover's Search Algorithm
+**************************
+
```
BoxiLi, line 6: Nice that you added a doc. Would be better to add an `Algorithm` section instead of just Grover. But can be addressed latter with a different PR.

doc/source/qip-grover.rst
BoxiLi: In the documentation, it is more common to show the plotting code and also the plots.

doc/source/qip-grover.rst
mudit06mah: From what I have noticed after looking around in the repository is that we can open issues for:
1) Adding a separate algorithm section in the docs along with documentation for the other algorithms
2) Adding type hints in the algorithms (maybe a bigger parent issue can be opened that adds type hints in all files as well)

What do you think @BoxiLi? Should I open issues for these?

doc/source/qip-grover.rst
mudit06mah: Please take a look at the updated changes

doc/source/qip-grover.rst
```diff
@@ -0,0 +1,79 @@
...
+    >>> qc = grover(oracle, n_qubits, num_solutions=len(marked))
+
+    We can now simulate the circuit to obtain the final state vector[cite: 17].
+
+    >>> U_grover = qc.compute_unitary()
+    >>> psi0 = tensor([basis(2, 0)] * n_qubits)
+    >>> psi_final = U_grover * psi0
+
+    To visualize the results, we calculate the measurement probabilities for all possible states and plot them[cite: 17, 18].
```
Mayank447: Cite 17, 18 is invalid here

doc/source/qip-grover.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Let's simulate a search on 3 qubits ($N=8$) where two states are marked: $|011\rangle$ (index 3) and $|101\rangle$ (index 5).
+
+.. plot::
+    :context: close-figs
+
+    First, we import the necessary components and define our search space.
+
+    >>> import matplotlib.pyplot as plt
+    >>> from qiskit.visualization import plot_histogram
```
Mayank447: qiskit is not a visualization dependency for qutip-qip. Although it is kept as an optional dependency for user to import qiskit circuits directly into qutip-qip.

So this plotting should be done using matplotlib instead

doc/source/qip-grover.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Overview
+========
+
+Grover's algorithm is a fundamental quantum algorithm that finds the unique input to a black-box function that produces a particular output value. In ``qutip-qip``, the implementation consists of two main components:
+
+* **The Oracle**: A phase oracle that flips the sign of the amplitude for the "marked" or "target" states.
+* **The Diffusion Operator**: Also known as the "inversion about the mean" operator, which amplifies the probability of measuring the marked states.
+
+The algorithm requires an optimal number of iterations, approximately $\frac{\pi}{4}\sqrt{\frac{N}{M}}$, where $N=2^n$ is the size of the search space and $M$ is the number of solutions.
```
Mayank447:
```suggestion
The algorithm requires an optimal number of iterations, approximately $\frac{\pi}{4}\sqrt{\frac{N}{M}}$, where $N=2^n$ is the size of the search space and $M$ is the number of marked states.
```

Let's stick to one terminology

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,225 @@
...
+
+    Returns
+    -------
+    QubitCircuit
+        The oracle circuit.
+    """
+    if isinstance(qubits, int):
+        qubits = list(range(qubits))
+
+    if isinstance(marked_states, int):
```
Mayank447, line 30:
```suggestion
    if not isinstance(marked_states, Iterable):
```

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,225 @@
...
+        marked_states = [marked_states]
+
+    n_qubits = len(qubits)
+    # We need a circuit big enough to hold the highest qubit index
+    qc = QubitCircuit(max(qubits) + 1)
+
+    for state in marked_states:
+        # Safety check
+        if state >= 2**n_qubits:
```
Mayank447: We need to have a similar check to ensure state is not less than zero. Because even negative states will have a valid binary representation

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,225 @@
...
+        for i, char in enumerate(binary_rep):
+            if char == "0":
+                qc.add_gate("X", targets=qubits[i])
+
+    return qc
+
+
+def grover(
+    oracle: QubitCircuit,
+    qubits: Union[int, List[int]],
```
Mayank447: You can use the new Python 3.10 type hinting notation, this is what are following

```suggestion
    qubits: int | list[int],
```

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,225 @@
...
+    num_solutions: int,
+    num_iterations: Optional[int] = None,
```
Mayank447: Same as above

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,225 @@
...
+) -> QubitCircuit:
+    """
+    Construct the Grover search algorithm's circuit.
+
+    Parameters
+    ----------
+    oracle : QubitCircuit, Gate, or Qobj
```
Mayank447:
```suggestion
    oracle : :class:`~.circuit.QubitCircuit` or :class:`~.operations.Gate` or :class:`
    qutip.Qobj`
```

We need to reference the qutip-qip class for Sphinx documentation. There are some other places where this change needs to be made

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,225 @@
...
+    qc = QubitCircuit(max(qubits) + 1)
+
+    # Superposition:
+    for q in qubits:
+        qc.add_gate("SNOT", targets=q)
+
+    # Calculate optimal Iterations if none provided:
+    if num_iterations is None:
+        N = 2**n_qubits
+        if num_solutions >= N:
```
Mayank447: num_solutions >= N is not a valid case, should ideally raise warning for this.

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,225 @@
...
+            num_iterations = 0
+        else:
+            calc = (np.pi / 4) * np.sqrt(N / num_solutions)
+            num_iterations = max(1, int(np.floor(calc)))
```
Mayank447: Any particular reason we need to take max with 1?

N/num_solution >= 1, I think it should work even without the max

Also edge case num_solution being zero would throw an error

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,225 @@
...
+    # Superposition:
+    for q in qubits:
+        qc.add_gate("SNOT", targets=q)
+
+    # Calculate optimal Iterations if none provided:
+    if num_iterations is None:
+        N = 2**n_qubits
+        if num_solutions >= N:
+            num_iterations = 0
+        else:
```
Mayank447: This does not handle the case where `num_solution = 0`, which is feasible if we use Grover as a subroutine for another algorithm

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,225 @@
...
+            calc = (np.pi / 4) * np.sqrt(N / num_solutions)
+            num_iterations = max(1, int(np.floor(calc)))
+
+    # Grover Iterations:
+    for _ in range(num_iterations):
+        # Oracle
+        if isinstance(oracle, QubitCircuit):
+            qc.gates.extend(oracle.gates)
+        elif isinstance(oracle, Gate):
+            if set(oracle.targets) != set(qubits):
```
Mayank447:
```suggestion
            if set(oracle.targets) not in set(qubits):
```

This allows for grover to run on even a subset of my qubits

tests/test_grover.py
```diff
@@ -0,0 +1,125 @@
...
+        U_sim = qc_oracle.compute_unitary()
+
+        dims = [[2] * n_qubits, [2] * n_qubits]
+        N = 2**n_qubits
+        diag = np.ones(N)
+        for s in marked_states:
+            diag[s] = -1
+        U_expected = Qobj(np.diag(diag), dims=dims)
+
+        assert (U_sim - U_expected).norm() < 1e-6
```
Mayank447: Why are we checking the norm here instead of the matrix elements. Use np.allclose

Generally in papers by norm tend to mean spectral norm which is not the case in numpy

tests/test_grover.py
```diff
@@ -0,0 +1,125 @@
...
+        Full algorithm test: 2 qubits, searching for |11> (state 3).
+        Optimal iterations = 1.
+        """
+        n_qubits = 2
+        target_state = 3  # |11>
+
+        oracle = grover_oracle(n_qubits, target_state)
+        qc = grover(oracle, n_qubits, 1)
+
+        U_grover = qc.compute_unitary()
```
Mayank447, line 85: It upto you to follow, but can we instead add measurements to QC here after Grover is done and from the measurement statistics, test whether we get marked_state as desire

tests/test_grover.py
```diff
@@ -0,0 +1,125 @@
...
+        sys_qubits = 4
+        search_qubits = [1, 2]
+        target_local_state = 3  # |11> on the 2 search qubits
+
+        oracle = grover_oracle(search_qubits, target_local_state)
+
+        # N=4, M=1. Optimal iterations = 1.
+        qc = grover(oracle, search_qubits, 1)
+
+        assert qc.num_qubits == 3  # max(1,2) + 1 = 3
```
Mayank447: Shouldn't it be 5 (same as sys_qubits + 1)?

Original system - q0, q1, q2, q3

I search on q1, q2 and now need to add  an ancilla in grover iterate

tests/test_grover.py
```diff
@@ -0,0 +1,125 @@
...
+        qc_a = grover(oracle_qobj, qubits, 1)
+        assert len(qc_a.gates) > 0
+
+        # Gate Oracle
+        oracle_gate = Gate("Z", targets=[0])
+        qc_b = grover(oracle_gate, qubits, 1)
+        assert len(qc_b.gates) > 0
+
+        # Compare unitaries
+        assert (qc_a.compute_unitary() - qc_b.compute_unitary()).norm() < 1e-6
```
Mayank447: Same comment as above, we should ideally compare the matrix elements

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,225 @@
...
+        super().__init__(targets=targets, **kwargs)
+        self.U = U
+        self.latex_str = r"Oracle"
+
+    def get_compact_qobj(self):
+        return self.U
+
+
+def grover_oracle(
+    qubits: Union[int, List[int]], marked_states: Union[int, List[int]]
```
Mayank447: Instead of restricting to List, we can allow for Sequence here in typehint. Your code works in the general case.

doc/source/qip-grover.rst
BoxiLi: Sure, sounds good.

BoxiLi:
Thanks for the update, you can download the rendered HTML files in the test action summary https://github.com/qutip/qutip-qip/actions/runs/22573151626?pr=314

Please check the result. Right now, the plot renders well. But as mentioned, it would be good to use matplotlib. I also see that the math functions are not correctly rendered, and also citations are not defined; We have a bibliography section.

doc/source/qip-grover.rst
```diff
@@ -0,0 +1,79 @@
...
+    >>> n_qubits = 3
+    >>> marked = [3, 5]
+
+    We then use the utility function :func:`.grover_oracle` to create a phase-flip oracle for our target states[cite: 8].
+
+    >>> oracle = grover_oracle(n_qubits, marked)
+
+    Using the :func:`.grover` function, we construct the full circuit. Since we provide the mandatory ``num_solutions``, the algorithm automatically calculates the optimal number of iterations[cite: 7, 11, 15].
+
+    >>> qc = grover(oracle, n_qubits, num_solutions=len(marked))
```
BoxiLi, line 59: It would also be intuitive to render the circuit here using `qc.draw`

doc/source/qip-grover.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Constructing the Algorithm
+==========================
+
+In this module, you can construct the full search circuit using the :func:`.grover` function. To make the process easier, a utility function :func:`.grover_oracle` is provided to generate standard phase-flip oracles.
+
+The :func:`.grover` function requires the following:
+
+====================  ==================================================
+Property                           Description
```
BoxiLi: Arguments, not property

tests/test_grover.py
mudit06mah: Actually we do not use an ancillia qubit here instead we're using MCZ.
So it'll be 3 qubits only.

mudit06mah:
@Mayank447 @BoxiLi  PTAL at the updated changes.

doc/source/qip-grover.rst
mudit06mah: @BoxiLi  Had some doubts before opening this issue:

- Should the Algorithms section be separate like the Tutorials section, or just one big algorithm page with all of the algorithms?
- Should I open the type hint issue for only the algorithms section or the whole codebase? This issue should maybe then be done on another branch rather than `master` since it'll be better to create small PRs for each folder, merge that into a bigger branch and then merge the bigger branch with master? Please tell me your opinion on this.

tests/test_grover.py
Mayank447: Yeah I am a little confused by this test. You have a 4 qubit system - [0,1,2,3].

qubit 0 and qubit 3 should remain identity while qubit 1, 2 must become the state 0b11. For MCZ my controls will be qubit 1,2. What is my target, it has to be ancilla right? It can't be qubit 0, 3 since they must remain identity. So in total requires 5 qubits (4 originally from the system + 1 extra qubit required by grover, which I referred to as ancilla)

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,232 @@
...
+
+def grover_oracle(
+    qubits: int | Sequence[int], marked_states: int | Sequence[int]
+) -> QubitCircuit:
+    """
+    Constructs a Phase Oracle circuit for Grover's algorithm.
+
+    Parameters
+    ----------
+    qubits : int or list of int
```
Mayank447:
```suggestion
    qubits : int or Sequence of int
```

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,232 @@
...
+        The specific qubit indices the oracle acts on.
+    marked_states : int or list of int
```
Mayank447: Same as above

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,232 @@
...
+    # Calculate optimal Iterations if none provided:
+    if num_iterations is None:
+        N = 2**n_qubits
+
+        if num_solutions <= 0:
+            raise ValueError("num_solutions must be greater than 0.")
+        if num_solutions >= N:
+            warnings.warn(
+                "Number of solutions is equal/greater to the search space. 0 Grover Iterations applied."
+            )
```
Mayank447:
```suggestion
            raise ValueError(
                "Number of solutions is equal/greater to the search space. 0 Grover Iterations applied."
            )
```

doc/source/qip-grover.rst
Mayank447: For second point there already exists an issue #262.

BoxiLi:
There might still be a bug in the doc rendering. The circuit plot does not show up where it is supposed to be, but instead replaces the histogram.
image

mudit06mah:
@BoxiLi I have fixed the documentation now, both diagrams render as intended.
image
image

tests/test_grover.py
mudit06mah: @Mayank447 We do not need an ancillia qubit, the target is one of the search qubits itself!
If you look at the code, you can see that:
```
ctrls = qubits[:-1]
tgt = qubits[-1]
```
That means, for search qubits: [1,2], `ctrls = 1`, `tgt =2`.
But after you pointed out how it should be equal to the `sys_qubits`, I added argument `N` for the total qubits in the circuit, this will ensure that even if grover is conducted on a subset of the qubits, the circuit doesn't lose it's wiring for other qubits. I think this shall clear your confusion.
(Now `qc.num_qubits = sys_qubits`)

Mayank447: Grover search does need an ancilla (which is initialized in | - > state.

For reference you can check
Step 1 in https://quantum.cloud.ibm.com/learning/en/courses/utility-scale-quantum-computing/grovers-algorithm

I can't use my qubits (q0, q3) as that ancilla (even though grover search doesn't produce a dirty ancilla) because they need not be in the | - >

mudit06mah: @Mayank447 Actually there is a difference in implementation between the oracles here :)
The oracle in the IBM reference you have provided uses a boolean oracle using a MCX gate, which requires an ancillia qubit initialized to |-> to achieve the phase kickback, whereas, my implementation uses a phase oracle using a MCZ gate which does not require an ancilia qubit.

For your reference: #https://github.com/Qiskit/textbook/blob/main/notebooks/ch-algorithms/grover.ipynb

I hope this clears things up 👾

mudit06mah:
@BoxiLi I have made the necessary changes, can you PTAL again?

Mayank447:
> @Mayank447 Actually there is a difference in implementation between the oracles here :) The oracle in the IBM reference you have provided uses a boolean oracle using a MCX gate, which requires an ancillia qubit initialized to |-> to achieve the phase kickback, whereas, my implementation uses a phase oracle using a MCZ gate which does not require an ancilia qubit.

MCX = H (MCZ) H, these Hadamard being applied on the same qubits as Z. So my qubits count should not be affected.

> For your reference: #https://github.com/Qiskit/textbook/blob/main/notebooks/ch-algorithms/grover.ipynb

Screenshot 2026-03-07 at 3 23 36 AM

In the notebook you referenced they do have an ancilla (auxillary qubit) initialized in | - > state. And its being used in the oracle, even you are doing the same - `total_qubits  = num_qubits + 1`, check your `grover_oracle` method. That extra +1 is for the ancilla. But what you're testing in `test_grover_custom_qubit_indices ` is wrong.

Entire phase kickback in fact relies on my other qubit (ancilla) being in | - > state. People generally only care about ancilla count and usually don't represent it in circuit draw, but for circuit construction and simulation etc. it is required.

mudit06mah:
Hi @Mayank447 !
It appears that I did not provide a correct reference to prove my point, PTAL at this: https://www.nature.com/articles/s41467-017-01904-7

In my `grover_oracle()` function I have indeed done:
`qc = QubitCircuit(max(qubits) + 1)`, this is just for correct indexing, for example if `max(qubits)=2` then the total qubits are `3`: `[0,1,2]`.

I hope the reference clears everything this time :)

Mayank447:
> Hi @Mayank447 ! It appears that I did not provide a correct reference to prove my point, PTAL at this: https://www.nature.com/articles/s41467-017-01904-7

Quoting Fig 1 of this paper: The last qubit $q_a$ is the ancilla qubit

> In my `grover_oracle()` function I have indeed done: `qc = QubitCircuit(max(qubits) + 1)`, this is just for correct indexing, for example if `max(qubits)=2` then the total qubits are `3`: `[0,1,2]`.

I see what you're doing, you are essentially discarding the last qubit and considering it as your ancilla. This assumes that the user QubitCircuit's accounts for that ancilla qubit. In all your test cases you have marked states as odd e.g. 3, 5 etc., ie. the last qubit initialized to 1, so XH|1> = -|-> (the global phase -1 is not physically relevant and is anyways uncomputed). Try changing one of marked_states to even, my hypothesis is it should not work. A quick fix might be `qc = QubitCircuit(max(qubits) + 2)`.

A side question, did you use LLMs in this PR? There is nothing wrong with it, they are a great set of tools and the future of programming. Its just that it should ideally be be disclosed in the PR description. Because on the reviewer's end it makes the review process harder since the code maybe clean and well structured, but it may lack verifiable correctness.

mudit06mah:
> I see what you're doing, you are essentially discarding the last qubit and considering it as your ancilla. This assumes that the user QubitCircuit's accounts for that ancilla qubit. In all your test cases you have marked states as odd e.g. 3, 5 etc., ie. the last qubit initialized to 1, so XH|1> = -|-> (the global phase -1 is not physically relevant and is anyways uncomputed). Try changing one of marked_states to even, my hypothesis is it should not work. A quick fix might be `qc = QubitCircuit(max(qubits) + 2)`.

I have tested your hypothesis locally and all test cases pass correctly, I think you are still trying to visualize the boolean oracle, my implementation does not use any sort of ancillia qubit at all.

Let me give you a dry run of the oracle logic:
Let's consider qubits = [0,1] (2 qubit system)
target state = 2 (|10>)

Step 1:
|10> gets converted to |11> by X gate on second qubit (0)

Step 2:
MCZ is applied on circuit, resulting in a phase flip of |11> to -|11>

Step 3:
-|11> gets uncomputed to -|10> (we have achieved phase flip without any |-> qubit)

If you still have any doubts you can test my changes locally as well :)


> A side question, did you use LLMs in this PR? There is nothing wrong with it, they are a great set of tools and the future of programming. Its just that it should ideally be be disclosed in the PR description. Because on the reviewer's end it makes the review process harder since the code maybe clean and well structured, but it may lack verifiable correctness.

Since I am majoring in CS, my only exposure to quantum computation has been through my elective of Quantum Mechanics for Engineers, hence I did have to use LLMs to understand the algorithm and help rectify the code wherever I was stuck but I have personally written the algorithm myself. I used LLMs to help me improve my approach, write test cases and do the documentation of the algorithm.
But I completely understand your concern as a reviewer. I will add the necessary to the PR Description as a note.

I hope this has cleared your doubts regarding the algorithm.

Mayank447:
So I ran your code locally, this is the script (taken from what Boxi used)

```
from qutip_qip.algorithms import grover, grover_oracle

n_qubits = 3
marked = [0, 1, 2, 3]

oracle = grover_oracle(n_qubits, marked)

qc = grover(oracle, n_qubits, num_solutions=len(marked))

U_grover = qc.compute_unitary()
psi0 = tensor([basis(2, 0)] * n_qubits)
psi_final = U_grover * psi0

print(psi_final)

prob = 0
for i in marked:
    prob_i = abs(psi_final.overlap(basis(2**n_qubits, i))) ** 2
    prob += prob_i

print(prob)
```

(qiskit) mayankgoel@MayankGoel-2 mudit % python temp.py
Quantum object: dims=[[2, 2, 2], [1]], shape=(8, 1), type='ket', dtype=Dense
Qobj data =
[[-0.35355339]
 [-0.35355339]
 [-0.35355339]
 [-0.35355339]
 [ 0.35355339]
 [ 0.35355339]
 [ 0.35355339]
 [ 0.35355339]]
 
0.4999999999999993

This is the output, clearly an equal superposition state but with different phases. And total probability being 0.5 instead of 1. Try it for other combinations of marked states and you will find it fails terribly. There are some more deeper bugs in your implementation I came across

- len(qubits) == 1, ctrls = qubits[:-1] evaluates to an empty list []. So if you ran e.g. marked=[1] it fails (there is no test case for num_solutions=1).
- I didn't get why you added `N`? If a user explicitly provides `num_iterations`, the validation block checking if `num_solutions >= N` is skipped entirely. A user could input 5 solutions for a 2-qubit circuit (which only has 4 possible states), and the code will proceed without raising the necessary ValueError
- Also inside the if num_iterations is None: block, the code executes N = 2**n_qubits. This overwrites the N argument.
- The sequence of X gates, followed by a multi-controlled Z, followed by X gates, applies a relative phase of $-1$ to the $|00\dots0\rangle$ state. Mathematically, this implements the operator $I - 2|0\rangle\langle 0|$. The standard theoretical Grover diffusion operator is $2|0\rangle\langle 0| - I$.

There is a might be more issues in addition to the above.

> Let me give you a dry run of the oracle logic: Let's consider qubits = [0,1] (2 qubit system) target state = 2 (|10>)

[0,1] in qutip_qip corresponds to the state |01>, while qiskit will interpret it as |10>

> I think you are still trying to visualize the boolean oracle, my implementation does not use any sort of ancillia qubit at all.

Any sort of implementation requires a reference. From what I can see from your previous references IBM Qiskit Textbook, Nature paper doesn't mention the Phase Oracle logic or MCZ and use an ancilla instead. Providing a verifiable reference e.g. paper etc. would make it easier to verify.

mudit06mah:
@Mayank447  Thankyou for your thorough review, you have listed out some valid points so I have made the following changes in my code:
- Added edge case for n_qubits = 1  (along with test case)
- Moved `num_solutions` validation outside
- I was incorrectly using the same variable name: `N` for the search space as well as the total number of qubits in the circuit, I have corrected that as well.
- You have correctly noted that my implementation and the theoretical implementation have a phase difference, but that phase difference causes no differences in the physics.

Please see the implementation of the phase oracle here:
https://quantum.cloud.ibm.com/docs/en/tutorials/grovers-algorithm

Also for the case you have stated, this is actually a limitation of the algorithm itself! For your reference:
https://quantumcomputing.stackexchange.com/questions/33246/general-grovers-search-showing-wrong-results-on-a-particular-instance

I have tested my algorithm for other cases and it is working as intended, if you can find any such case where it is not I'll be happy to re-iterate on my implementation.

Mayank447:
Cool, the implementation now looks good and based on my preliminary testing it works well. I will review the code changes in a day or two.

So I was reading about Phase Oracles using MCZ gates. Now I see why it works in this case and why no one uses it in a theory paper or in practice since its works only in one setting. Generally people are interested in SAT problem (i.e. problems where they don't know the marked solutions), e.g. finding the values of a boolean variable which satisfies the boolean formula $f(x)$. If I just prepare my states in an equal superposition, f(x) can be written using reversible gates (you may search for implementation of digital gates like AND, OR, etc. in terms of CSWAP), using my ancilla I can ensure a phase kickback for any general $f(x)$ and let the Grover run. The math remains the same and I can get solutions to my SAT problems with a square root speedup compared to a linear search over the set of possible variables.

Now since we are specifying the solutions/marked states, my oracle is known (i.e. I can implement it using my marked states). Where as in the general case, I only have my implementation for $f(x)$, not the oracle $U_f$. I can construct that $U_f$ using $f$ using a single ancilla. In case of known oracle (known marked states), my oracle is just a diagonal matrix of size (NxN, where $N=2^n - 1$). Since my diagonal entries are either 1, -1. I can implement it by brute force using MCZ by setting my control value accordingly and iterating through marked solutions (-1 diagonal entries). I think I am clear with the Phase Oracle now.

I just have one question for you, what would an application scenario for `grover()` in qutip-qip look like or some sub routine where we can use it?

mudit06mah:
@Mayank447 
You're correct about the Phase Oracle now, actually the `grover_oracle()` is more of a helper to write unit tests or for studying textbook examples.

The main `grover()` function has been written to accept any Quantum Circuit or Gate as an argument, which enables it to run any circuit the user passes as the oracle for the algorithm. This can hence allow users to run SAT problems as you mentioned or any other simulation by passing it to the function!

Few implementations that I could think of:
- It can be used as a subroutine in amplitude amplification.
- It can be used to stress test hardware models (against t1/t2 decay) to note how much probability of marked states drops.

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,257 @@
...
+
+    return qc
+
+
+def grover(
+    oracle: QubitCircuit,
+    qubits: int | Sequence[int],
+    num_solutions: int,
+    num_iterations: int | None = None,
+    N: int | None = None,
```
BoxiLi: Let's use `num_qubits`, as in elsewhere in qutip-qip.

Maybe `qubits` -> `target_qubits` or `search_targets` ?

tests/test_grover.py
```diff
@@ -0,0 +1,172 @@
...
+        assert total_success_prob > 0.999999
+
+    def test_grover_custom_qubit_indices(self):
+        """
+        Integration check: Run Grover on qubits [1, 2] of a 4-qubit system [0, 1, 2, 3].
+        Qubits 0 and 3 should remain Identity.
+        """
+
+        sys_qubits = 4
+        search_qubits = [1, 2]
```
BoxiLi, line 127: If `search_qubits` is unsorted, will it still work??

src/qutip_qip/algorithms/grover.py
BoxiLi: And what is the use scenario for this `N`? If this is just a sub-block of a big algorithm, will this not automatically work when a small circuit is created here and concatenated to a larger circuit?

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,257 @@
...
+
+    # Validation check for num_solutions:
+    if num_solutions <= 0:
+        raise ValueError("num_solutions must be greater than 0.")
+    if num_solutions >= search_space_size:
+        raise ValueError(
+            "Number of solutions is equal/greater to the search space."
+        )
+
+    total_qubits = N if N is not None else (max(qubits) + 1)
```
BoxiLi: Why not rename `N` -> `total_qubits` then, or rather just `num_qubits`.

src/qutip_qip/algorithms/grover.py
mudit06mah: @BoxiLi yes, it will work fine when a small circuit is appended to the parent circuit, but I this this is a good to have for user convenience, for example when a user uses `qc.compute_unitary` to multiply with a tensor without building a parent circuit, user can ensure that the circuit has the correct dimensions.

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,259 @@
...
+            # uncompute by X
+            for i, char in enumerate(binary_rep):
+                if char == "0":
+                    qc.add_gate("X", targets=search_qubits[i])
+
+    return qc
+
+
+def grover(
+    oracle: QubitCircuit,
```
Mayank447:
```suggestion
    oracle: QubitCircuit | Gate,
```

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,259 @@
...
+    num_solutions: int,
+    num_iterations: int | None = None,
+    num_qubits: int | None = None,
+) -> QubitCircuit:
+    """
+    Construct the Grover search algorithm's circuit.
+
+    Parameters
+    ----------
+    oracle : :class:`~.circuit.QubitCircuit` or :class:`~.operations.Gate` or :class:`qutip.Qobj`
```
Mayank447:
```suggestion
    oracle : :class:`~.circuit.QubitCircuit` or :class:`~.operations.Gate`
```

There was a mismatch between the docstring and function typehint. Let's just restrict it to `QubitCircuit`, `Gate`.

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,259 @@
...
+        if isinstance(oracle, QubitCircuit):
+            qc.gates.extend(oracle.gates)
+        elif isinstance(oracle, Gate):
+            if not set(oracle.targets).issubset(search_qubits):
+                raise ValueError(
+                    f"Oracle gate targets {oracle.targets} are not a subset of algorithm qubits {search_qubits}"
+                )
+            qc.add_gate(oracle)
+        else:
+            qc.add_gate(OracleGate(search_qubits, oracle))
```
Mayank447:
```suggestion
```

This can be removed now. After gate refactor this is a single line function get_unitary_gate which converts the Qobj to a Gate along with some required input arguments and checks. Also search_qubits are targets won't be part of the Gate.

tests/test_grover.py
```diff
@@ -0,0 +1,172 @@
...
+            grover(oracle, [1, 2], 1, num_qubits=-1)
+
+    def test_grover_invalid_num_iterations(self):
+        """Test that grover raises errors for invalid num_iterations."""
+        oracle = grover_oracle(2, 3)
+
+        with pytest.raises(
+            ValueError, match="num_iterations must be a positive"
+        ):
+            grover(oracle, 2, 1, num_iterations=0)
```
Mayank447:
```suggestion
```

I think we only need to test for `negative num_iterations`. `num_iterations=0` just means we don't apply any Oracle, Diffuser Layer. If in some subroutine the state is already in one marked states, the number of iterations can be 0.

doc/source/qip-grover.rst
```diff
@@ -0,0 +1,86 @@
...
+==========================
+
+In this module, you can construct the full search circuit using the :func:`.grover` function. To make the process easier, a utility function :func:`.grover_oracle` is provided to generate standard phase-flip oracles.
+
+The :func:`.grover` function requires the following:
+
+====================  ==================================================
+Argument                           Description
+====================  ==================================================
+``oracle``            A :class:`.QubitCircuit`, :class:`.Gate`, or :class:`~.Qobj` representing the phase oracle.
```
Mayank447:
```suggestion
``oracle``            A :class:`.QubitCircuit`, :class:`.Gate` representing the phase oracle.
```

doc/source/qip-grover.rst
```diff
@@ -0,0 +1,86 @@
...
+``qubits``            List of qubit indices (or integer count) to run the search on.
+``num_solutions``     **Mandatory** integer representing the number of marked states :math:`M`.
+``num_iterations``    Optional integer for manual control over the rotation count.
+``N``                 Optional total number of qubits when searching on a subset of a larger system.
```
Mayank447:
```suggestion
``num_qubits``                 Optional total number of qubits when searching on a subset of a larger system.
```

In accordance with the recent changes.

src/qutip_qip/algorithms/grover.py
```diff
@@ -0,0 +1,259 @@
...
+            The unitary operator representing the oracle.
+        **kwargs
+            Additional keyword arguments to pass to the parent Gate class.
+        """
+        super().__init__(targets=targets, **kwargs)
+        self.U = U
+        self.latex_str = r"Oracle"
+
+    def get_compact_qobj(self):
+        return self.U
```
Mayank447:
```suggestion
```

This can be removed, its essentially wrapping around Qobj. `Gate` class definition has changed considerably now and some methods like `get_compact_qobj` have been renamed to `get_qobj`, `targets` have been removed and several new class attributes have been added.

tests/test_grover.py
```diff
@@ -0,0 +1,172 @@
...
+
+        # Gate Oracle
+        oracle_gate = Gate("Z", targets=[0])
+        qc_b = grover(oracle_gate, qubits, 1)
+        assert len(qc_b.gates) > 0
+
+        # Compare unitaries
+        assert np.allclose(
+            qc_a.compute_unitary().full(), qc_b.compute_unitary().full()
+        )
```
Mayank447:
```suggestion
```

These will anyways need to be removed after Gate refactor, better to do it in this PR itself.

doc/source/qip-grover.rst
```diff
@@ -0,0 +1,86 @@
...
+
+Grover's algorithm :cite:`grover1996fast` provides a quadratic speedup for searching an unstructured database of size :math:`N`.
+It works by iteratively rotating a quantum state vector toward a "marked" solution state using a phase oracle and a diffusion operator.
+
+Overview
+========
+
+Grover's algorithm is a fundamental quantum algorithm that finds the unique input to a black-box function that produces a particular output value. In ``qutip-qip``, the implementation consists of two main components:
+
+* **The Oracle**: A phase oracle that flips the sign of the amplitude for the "marked" or "target" states.
```
Mayank447: Just a general comment you need to define what you mean by marked, target beforehand.

doc/source/qip-grover.rst
```diff
@@ -0,0 +1,86 @@
...
+Grover's Search Algorithm
+**************************
+
+Grover's algorithm :cite:`grover1996fast` provides a quadratic speedup for searching an unstructured database of size :math:`N`.
+It works by iteratively rotating a quantum state vector toward a "marked" solution state using a phase oracle and a diffusion operator.
+
+Overview
+========
+
+Grover's algorithm is a fundamental quantum algorithm that finds the unique input to a black-box function that produces a particular output value. In ``qutip-qip``, the implementation consists of two main components:
```
Mayank447:
```suggestion
Grover's algorithm is a fundamental quantum algorithm that finds the unique input to a black-box function that produces a particular output value. The implementation consists of two main components:
```

doc/source/qip-grover.rst
```diff
@@ -0,0 +1,86 @@
...
+
+* **The Oracle**: A phase oracle that flips the sign of the amplitude for the "marked" or "target" states.
+* **The Diffusion Operator**: Also known as the "inversion about the mean" operator, which amplifies the probability of measuring the marked states.
```
Mayank447:
```suggestion
* **The Diffusion Operator**: This operator flips the current state about its previous state.
```

doc/source/qip-grover.rst
```diff
@@ -0,0 +1,86 @@
...
+
+The algorithm requires an optimal number of iterations, approximately :math:`\frac{\pi}{4}\sqrt{\frac{N}{M}}`, where :math:`N=2^n` is the size of the search space and :math:`M` is the number of marked states.
+
+Constructing the Algorithm
+==========================
+
+In this module, you can construct the full search circuit using the :func:`.grover` function. To make the process easier, a utility function :func:`.grover_oracle` is provided to generate standard phase-flip oracles.
```
Mayank447:
```suggestion
In this module, you can construct the full Grover search circuit using the :func:`.grover` function. To make the process easier, a utility function :func:`.grover_oracle` is provided to generate standard phase-flip oracles, if the marked states are known.
```

doc/source/qip-grover.rst
```diff
@@ -0,0 +1,86 @@
...
+
+The :func:`.grover` function requires the following:
+
+====================  ==================================================
+Argument                           Description
+====================  ==================================================
+``oracle``            A :class:`.QubitCircuit`, :class:`.Gate`, or :class:`~.Qobj` representing the phase oracle.
+``qubits``            List of qubit indices (or integer count) to run the search on.
+``num_solutions``     **Mandatory** integer representing the number of marked states :math:`M`.
```
Mayank447:
```suggestion
``num_solutions``     integer representing the number of marked states :math:`M`.
```

Since you're already already specifying some arguments as Optional, no need for mandatory.

doc/source/qip-grover.rst
```diff
@@ -0,0 +1,86 @@
...
+
+    First, we import the necessary components and define our search space.
+
+    >>> import matplotlib.pyplot as plt
+    >>> from qutip import basis, tensor
+    >>> from qutip_qip.algorithms.grover import grover, grover_oracle
+    >>> n_qubits = 3
+    >>> marked = [3, 5]
+
+    We then use the utility function :func:`.grover_oracle` to create a phase-flip oracle for our target states.
```
Mayank447:
```suggestion
    We then use the utility function :func:`.grover_oracle` to create a phase-flip oracle for our marked states.
```

mudit06mah:
@BoxiLi I think you can proceed with merging it if it's not a problem, after this is merged I will open an issue for adding documentation for Algorithms as well as tutorials for grover and other algorithms 👾

BoxiLi:
Thank you!

PR #316, opened 2026-01-24
Remove QubitCircuit.gates

Mayank447:
**Description:**

- This PR replaces the usage of `QubitCircuit.gates` with `QubitCircuit.instructions`. This is done in order to store the targets, controls and gate style (used by circuit draw) etc. within the `QubitCircuit` class itself. Next step will be to finally remove `targets`, `controls` from the `Gate` class.

- `QubitCircuit.instructions` is currently a list which is appended to by the `add_gate()` and `add_measurement()` methods. When adding gates to the circuit, the instructions is appended with a tuple (gate, qubits, classical_controls, classical_control_value, style). `qubits = controls + targets` and each gate now has the property `num_ctrl_qubits` which is by default 0 as defined in the `Gate` class and in case of `ControlledGate` inferred automatically. This removes the need of storing controls, target separately in a Circuit Instruction.

**Changes:**
- Moved circuitsimulator to its own dir (no API changes).
- Removed `add_1q_gate()` method in `QubitCircuit`
- Removed `add_gates(gates)` method in `QubitCircuit` as it was a 2 line function consisting of a for loop over `add_gate` and the argument didn't even have `**kwargs`.
- Removed `adjacent_gates()` method in `QubitCircuit` as superseded by `to_chain()` in `src/transpiler`
- Removed index argument from `add_gate`, `add_measurement` as it is very inefficient to be performed on a list. Also there was no check if the index is say like 100 and but our circuit only has 10 gates.
- Added named arguments to when `add_gate` method is being called in the codebase. This is purely from the point of a general good practice, no compulsion on user end.
- Added arguments (targets etc.) to lines like `qc.add_gate(gate)`. This will be properly finished once all gate arguments are removed in Gate class.
- Added `qubit_count`, `num_ctrl_qubits` to the `Gate` class.
- Minor change to qft, repetition code schemes (bit flip etc.)


Mayank447:
@BoxiLi I have removed the usage of `qc.gates` internally in `QubitCircuit` and also in `CircuitSimulator`, Pulse based simulator. There are still quite a bit of files at which `qc.gates` is being still used.

But before I move on to complete replacement I believe instead of leaving each circuit instruction as a tuple as it is we should make it a dataclass ,as we often need to do some operations like find the control qubits based the `gate.num_ctrl_qubits` and `qubits` which can be automated in a function. Also using instruction[0] as replacement for gate in the codebase feels a little unarticulated. Even for pulse simulation there is an `Instruction` class defined. Do you want me to it in this PR or a separate one?

doc/source/qip-basics.rst
```diff
@@ -37,13 +37,12 @@ A circuit with the various gates and registers available is demonstrated below:
   from qutip import tensor, basis
 
   qc = QubitCircuit(N=2, num_cbits=1)
-  swap_gate = SWAP(targets=[0, 1])
 
-  qc.add_gate(swap_gate)
+  qc.add_gate("SWAP", targets=[0, 1])
```
BoxiLi, line 41: Just double check, the old code
```
swap_gate = SWAP(targets=[0, 1])
qc.add_gate(swap_gate)
```
will not fail, right? It is just moved to a more suggested way of using.

src/qutip_qip/algorithms/qpe.py
```diff
@@ -63,22 +63,19 @@ def qpe(U, num_counting_qubits, target_qubits=None, to_cnot=False):
...
-            target_gate=CustomGate(
+        controlled_u = controlledGateFactory(
+            target_gate=customGate(
                 name="U^power gate",
-                targets=target_qubits,
                 U=U_power,
-            ),
-            controls=[i],
-            targets=target_qubits,
+            )
```
BoxiLi, line 70: Maybe we should keep the old interface `ControlledGate` instead of renaming it to `controlledGateFactory`? Given that the interfaces are the same. I know the latter is more appropriate and shows how this is implemented. Just less confusion for users.

BoxiLi (commented):
Quick feedback, haven't finished the full review:

I have a feeling that there are some people using `QubitCircuit.gates`. Can we provide the following:
- When calling `QubitCircuit.gates`, still construct and return a list of gates constructed from `QubitCircuit.instructions`?
- And if they try to modify `QubitCircuit.gates`, raising an error saying that this is no longer supported. And maybe point out the right way of doing things.

doc/source/qip-basics.rst
Mayank447: For now the old code will work. But once we remove targets, gates from Gate class it won't

src/qutip_qip/algorithms/qpe.py
Mayank447: We already have a `ControlledGate` class, that's why had to choose a different name

Mayank447:
> * When calling `QubitCircuit.gates`, still construct and return a list of gates constructed from `QubitCircuit.instructions`?

Yes we can do that the only issue is QubitCircuit.gates was a list of Gates and those gate object contained the target, controls, but targets, controls no longer would be present after gate refactor.

> * And if they try to modify `QubitCircuit.gates`, raising an error saying that this is no longer supported. And maybe point out the right way of doing things.

I will raise a warning for anyone trying to access `QubitCircuit.gates` (we can do it via setter, properties) once .gates is completely replaced in the codebase.

doc/source/qip-basics.rst
BoxiLi: Hmmm, I see, that makes sense. It seems that we kind of replaced gate with instruction.

src/qutip_qip/circuit/_decompose.py
```diff
@@ -6,383 +6,485 @@
...
     GLOBALPHASE,
     RX,
     RY,
     RZ,
-    CSIGN,
     CNOT,
-    ISWAP,
-    SQRTISWAP,
-    SQRTSWAP,
 )
```
BoxiLi, line 14: I am slightly confused by this. They are removed because they are not used? Then how come they were imported before and passed the tests?

src/qutip_qip/circuit/circuit.py
```diff
@@ -71,6 +64,7 @@ def __init__(
         self.N = N
         self.reverse_states = reverse_states
         self.gates = []
+        self.instructions = []
```
BoxiLi: Should we expose this attribute to the user? Are we sure? Maybe we will modify it in the near future?

src/qutip_qip/circuit/circuit.py
```diff
@@ -130,9 +124,7 @@ def add_state(self, state, targets=None, state_type="input"):
             for i in targets:
                 self.output_states[i] = state
 
-    def add_measurement(
-        self, measurement, targets=None, index=None, classical_store=None
-    ):
+    def add_measurement(self, measurement, targets=None, classical_store=None):
```
BoxiLi: Do the same as `add_gate`? Although this interface is not much used by users.
```
        if index is not None:
            raise ValueError("argument index is no longer supported")
```

tests/test_device.py
```diff
@@ -31,32 +31,30 @@
...
+    pytest.param(2, [_ry], [0], id="RY"),
+    pytest.param(2, [ISWAP], [0, 1], id="ISWAP"),
+    pytest.param(
+        2, [SQRTISWAP], [0, 1], id="SQRTISWAP", marks=pytest.mark.skip
+    ),
 ]
 
+# This test needs to be added like single_gate_tests but requires controls
+# pytest.param(2, CNOT, targets=[0], controls=[1], id="CNOT"),
+
```
BoxiLi: Add a TODO mark?

src/qutip_qip/circuit/circuit.py
BoxiLi, line 194: What we allow now is not an instance of class, right? it is the class itself.

tests/test_qpe.py
```diff
@@ -125,6 +122,7 @@ def test_qpe_controlled_gate_powers(self):
             gate = circuit.gates[num_counting + i]
             power = 2 ** (num_counting - i - 1)
 
+            print(gate.target_gate)
```
BoxiLi: ?

src/qutip_qip/operations/gateclass.py
```diff
@@ -433,39 +451,69 @@ def _to_qasm(self, qasm_out):
...
-class CustomGate(Gate):
+def customGate(name: str, U: Qobj) -> Gate:
     """
-    Custom gate that wraps an arbitrary quantum operator.
+    Gate Factory for Custom Gate that wraps an arbitrary unitary matrix U.
     """
 
-    latex_str = r"U"
+    class CustomGate(Gate):
+        latex_str = r"U"
```
BoxiLi, line 460: This should also be documented in the user guide at some point.

src/qutip_qip/algorithms/qpe.py
BoxiLi: Ok, now I get it. Since they are functions now, maybe one should use `controlled_gate_factory` and `custom_gate_factory`?

BoxiLi (commented):
Yes, we can give a warning in the `property` attribute (maybe return instructions list then, since new gates contain only incomplete information). And an error for `setter`.

I am not sure if you noticed, in `compiler.py`, a class Instruction was defined, which was actually used to include the pulse information etc. Here, right now, `circuit.instructions` is saving a list of tuples. What is the long-term plan?

src/qutip_qip/circuit/_decompose.py
Mayank447: We can pass a standard gate both as a class or a string

tests/test_qpe.py
Mayank447: A debug print (forgot to remove, will do so)

src/qutip_qip/operations/gateclass.py
Mayank447: Yes, I will do this in a PR once all the code changes are done

src/qutip_qip/circuit/circuit.py
Mayank447: For parametrized gate we are allowing instance of class, cause the argument can be modified like in VQAs

Mayank447: Since non-parametrized gates don't have an `__init__` i.e. a class constructor, there is one instance of class which is the class itself.

src/qutip_qip/circuit/circuit.py
Mayank447: Ideally we should expose this as a property

Mayank447:
> I am not sure if you noticed, in `compiler.py`, a class Instruction was defined, which was actually used to include the pulse information etc. Here, right now, `circuit.instructions` is saving a list of tuples. What is the long-term plan?

I did notice `Instruction` class being defined for pulses. The long-term plan is to have a base class `CircuitInstruction` on top of this class we define the `GateInstruction`, `MeasurementInstruction` (in future if someone wants, they can implement Barrier, reset similarly). `circuit.instructions` will be a list of `CircuitInstruction` class children. This is my plan for the next PR.

We need this cause all instructions don't necessarily follow the same tuple structure as Gate.

BoxiLi:
Looks good, just let me know when you are done with the final changes and I will merge.

Mayank447:
@BoxiLi I have made the changes

BoxiLi (approved):

PR #319, opened 2026-02-03
Add Circuit Instruction class and remove targets, controls from the gate class

Mayank447:
**Changes:**
- Based on our discussion in #308 implemented `CircuitInstruction`, `GateInstruction`, `MeasurementInstruction` classes.
- Added `num_qubits` and `num_ctrl_qubits` property to the Gate class.
- Added `_global_phase` attribute in `QubitCircuit` (exposed via property). Any `GLOBALPHASE` gate will just add to the circuit's global phase. Also `GLOBALPHASE` gate has been retained as it would be helpful in Gate decompositions.
- `QubitCircuit.gates` has been completely replaced by `QubitCircuit.instruction` in the codebase. Also added a warning via property for someone trying to access `QubitCircuit.gates` and returning `QubitCircuit.instructions` instead as discussed.
- Moved qasm logic from operations (Gates, Measurement) to `CircuitInstruction`.
- Removed `targets`, `controls`, `classical_controls`, `classical_control_values` and `style` from the Gate class. 
- Renamed `Instruction` to `PulseInstruction` in compiler and deprecated the former.

### DIFF
--- a/src/qutip_qip/circuit/circuit.py
+++ b/src/qutip_qip/circuit/circuit.py
@@ -26,7 +26,6 @@ from qutip_qip.circuit import (
 from qutip_qip.circuit.utils import _check_iterable, _check_limit_
 from qutip import qeye, Qobj
 
-
 try:
     from IPython.display import Image as DisplayImage, SVG as DisplaySVG
 except ImportError:
@@ -110,6 +109,7 @@ class QubitCircuit:
         return self._instructions
 
     gates.setter
+
     def gates(self):
         warnings.warn(
             "QubitCircuit.gates has been replaced with QubitCircuit.instructions",
@@ -371,10 +371,8 @@ class QubitCircuit:
                 )
 
             else:
-                raise TypeError(
-                    f"The circuit to be added contains unknown \
-                    operator {circuit_op[0]}"
-                )
+                raise TypeError(f"The circuit to be added contains unknown \
+                    operator {circuit_op[0]}")
 
     def remove_gate_or_measurement(
         self, index=None, end=None, name=None, remove="first"
@@ -394,20 +392,16 @@ class QubitCircuit:
         """
         if index is not None:
             if index > len(self.instructions):
-                raise ValueError(
-                    "Index exceeds number \
-                    of gates + measurements."
-                )
+                raise ValueError("Index exceeds number \
+                    of gates + measurements.")
 
             if end is not None and end <= len(self.instructions):
                 for i in range(end - index):
                     self._instructions.pop(index + i)
 
             elif end is not None and end > self.N:
-                raise ValueError(
-                    "End target exceeds number \
-                    of gates + measurements."
-                )
+                raise ValueError("End target exceeds number \
+                    of gates + measurements.")
 
             else:
                 self._instructions.pop(index)
@@ -568,10 +562,8 @@ class QubitCircuit:
             )
         )
         if num_measurements > 0:
-            raise NotImplementedError(
-                "adjacent_gates must be called before \
-                measurements are added to the circuit"
-            )
+            raise NotImplementedError("adjacent_gates must be called before \
+                measurements are added to the circuit")
 
         basis_1q_valid = ["RX", "RY", "RZ", "IDLE"]
         basis_2q_valid = ["CNOT", "CSIGN", "ISWAP", "SQRTSWAP", "SQRTISWAP"]
@@ -765,7 +757,7 @@ class QubitCircuit:
         U_list : list
             Return list of unitary matrices for the qubit circuit.
             The global phase of circuit is the last element of ``U_list``.
-            
+
         Notes
         -----
         If ``expand=False``, the global phase gate only returns a number.
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -11,4 +11,4 @@ jobs:
         with:
           options: "--check --diff"
           src: "./src"
-          version: "~= 24.0"
+          version: "~= 26.0"
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -145,9 +145,7 @@ warnings.filterwarnings("ignore")
 import numpy as np
 np.set_printoptions(precision=5)
 os_nt = {}
-""".format(
-    os_nt
-)
+""".format(os_nt)
 
 # -- Options for plot directive ---------------------------------------
 
--- a/src/qutip_qip/algorithms/qft.py
+++ b/src/qutip_qip/algorithms/qft.py
@@ -8,7 +8,6 @@ from qutip_qip.circuit import QubitCircuit
 from qutip import Qobj
 from qutip_qip.decompose import decompose_one_qubit_gate
 
-
 __all__ = ["qft", "qft_steps", "qft_gate_sequence"]
 
 
--- a/src/qutip_qip/circuit/_decompose.py
+++ b/src/qutip_qip/circuit/_decompose.py
@@ -7,7 +7,6 @@ individual gate classes.
 import numpy as np
 from qutip_qip.operations import RX, RY, RZ, CNOT, ParametrizedGate
 
-
 __all__ = ["_resolve_to_universal", "_resolve_2q_basis"]
 
 
--- a/src/qutip_qip/circuit/draw/base_renderer.py
+++ b/src/qutip_qip/circuit/draw/base_renderer.py
@@ -105,11 +105,9 @@ class StyleConfig:
         elif self.theme == "modern":
             self.theme = modern
         else:
-            raise ValueError(
-                f"""Invalid theme: {self.theme},
+            raise ValueError(f"""Invalid theme: {self.theme},
                 Must be selectec from 'qutip', 'light', 'dark', or 'modern'.
-                """
-            )
+                """)
 
         self.bgcolor = self.bgcolor or self.theme["bgcolor"]
         self.color = self.color or self.theme["color"]
--- a/src/qutip_qip/device/optpulseprocessor.py
+++ b/src/qutip_qip/device/optpulseprocessor.py
@@ -145,7 +145,7 @@ class OptPulseProcessor(Processor):
             setting_args = {}
 
         if isinstance(qc, QubitCircuit):
-            props = qc.propagators()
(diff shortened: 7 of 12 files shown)


Mayank447:
@BoxiLi This PR is ready for review. Also were the test settings changed, tests now require maintainer approval even for people who have an existing PR merged.

src/qutip_qip/circuit/draw/mat_renderer.py
```diff
@@ -585,16 +583,7 @@ def _draw_multiq_gate(self, gate: Gate, layer: int) -> None:
             )
 
         else:
-            adj_targets = [
-                i + self._cwires
-                for i in sorted(
-                    gate.targets
-                    if gate.targets is not None
-                    else list(
-                        range(self._qwires)
```
BoxiLi, line 594: Where does this _qwires logic go?

I kind of forget why it is here. When a gate has no targets, globalphase?

Mayank447:
This PR is complete from my end.

src/qutip_qip/circuit/instruction.py
```diff
@@ -0,0 +1,104 @@
...
+
+        if q < 0:
+            raise ValueError(f"{txt} indices must be non-negative, found {q}")
+
+
+@dataclass(frozen=True)
+class CircuitInstruction(ABC):
+    operation: Gate | Measurement
+    qubits: tuple[int] = tuple()
+    cbits: tuple[int] = tuple()
```
BoxiLi, line 22: I remember there used to be classical control and classical storage. They are merged into cbits now?

BoxiLi (commented):
Yes there was a discussion in the qutip admin meeting about this the conclusion leads to this change of policy. But for qutip-qip I guess we can make this easier for now.

src/qutip_qip/circuit/instruction.py
Mayank447: Yes

src/qutip_qip/circuit/draw/mat_renderer.py
Mayank447: _qwires was for Global Phase gate

src/qutip_qip/circuit/instruction.py
```diff
@@ -0,0 +1,150 @@
...
+        _validate_non_negative_int_tuple(self.qubits, "qubit")
+        _validate_non_negative_int_tuple(self.cbits, "cbit")
+
+        if len(self.qubits) != len(set(self.qubits)):
+            raise ValueError("Found repeated qubits")
+
+        if len(self.cbits) != len(set(self.cbits)):
+            raise ValueError("Found repeated cbits")
+
+    def is_gate_instruction(self) -> bool:
```
BoxiLi, line 41: Why having this here as a method instead of checking the instruction class type?

src/qutip_qip/circuit/instruction.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+@dataclass(frozen=True)
+class GateInstruction(CircuitInstruction):
+    operation: Gate
+    cbits_ctrl_value: int | None = None
+
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        if not self.is_gate_instruction():
```
BoxiLi: This is defined to return true in `GateInstruction`, not really checking if the operation is gate or not.

Mayank447:
> Sorry I will have to do this review piece by piece.

Sure, no problem. As a suggestion you can try doing it commit-wise for large PRs as the changes/logic tends to stack up.

> Wanted to add a note here that, as a migration period, we can let `circuit.gates` return `circuit.instructions` and then give a warning that the behavior is changed and `circuit.gates` will be removed soon completely in the next few releases. This will minimize the breaking of existing code, I think.

The moment we return in a function, any code below it won't execute so have to give the warning first. That's why in the code I am warning first and then returning circuit.instructions.

src/qutip_qip/circuit/instruction.py
Mayank447: Couple of reasons. `isinstance` in python is extremely slow and often the bottleneck. Since it has to be executed at the run time and isn't generally cached (because class object might change).

I wanted to avoid isinstance as much as possible and that's why went with this design. Also since I am doing a thorough validation check on my input parameters and my dataclass is frozen, user cannot overwrite this method.

src/qutip_qip/circuit/instruction.py
Mayank447: Nice catch, I was previously checking `isinstance(gate, Gate)` in the `is_gate_instruction()` method. After some updation forgot to put that check here, the same change needs to be done in `MeasurementInstruction`.

Also just a note, the reason we are doing this is because `isinstance` will only be called once while validating the parameters. Whereas `is_gate_instruction` which is called repeatedly in the codebase is free from `isinstance` call.

src/qutip_qip/compiler/instruction.py
```diff
@@ -33,15 +33,22 @@ class Instruction:
...
 
-    def __init__(self, gate, tlist=None, pulse_info=(), duration=1):
-        self.gate = deepcopy(gate)
+    def __init__(
+        self, circuit_instruction, tlist=None, pulse_info=(), duration=1
+    ):
+        self.gate = deepcopy(circuit_instruction.operation)
+        self._targets = list(circuit_instruction.targets)
+        self._controls = list(circuit_instruction.controls)
+
```
BoxiLi, line 42: We should probably rename this old Instruction class. I feel like the propertes it has are very close to the new `CircuitInstruction`. It has the gate target qubits, and on top of this some pulse informations such as pulse array and duration etc. Do you think if it would make sense to make it a `PulseInstruction`?

src/qutip_qip/circuit/circuit.py
```diff
@@ -724,16 +739,13 @@ def propagators(self, expand=True, ignore_measurement=False):
...
-            else:
-                qobj = gate.get_compact_qobj()
-                if expand:
-                    qobj = expand_operator(
-                        qobj, dims=self.dims, targets=targets
-                    )
+            qobj = gate.get_compact_qobj()
+            if expand:
+                qobj = expand_operator(qobj, dims=self.dims, targets=targets)
             U_list.append(qobj)
```
BoxiLi, line 784: What about the global phase here? The phase is not applied right?

src/qutip_qip/circuit/circuit.py
```diff
@@ -85,12 +92,34 @@ def __init__(
...
+            "QubitCircuit.gates has been replaced with QubitCircuit.instructions",
+            DeprecationWarning,
+            stacklevel=2,
+        )
         return self._instructions
 
-    @instructions.setter
-    def instructions(self, x):
-        self._instructions = x
+    gates.setter
```
BoxiLi, line 111: What is this `gates.setter` here? Do you mean to use the setter decoration?

BoxiLi:
> > Wanted to add a note here that, as a migration period, we can let circuit.gates return circuit.instructions and then give a warning that the behavior is changed and circuit.gates will be removed soon completely in the next few releases. This will minimize the breaking of existing code, I think.
> 
> The moment we return in a function, any code below it won't execute so have to give the warning first. That's why in the code I am warning first and then returning circuit.instructions.

Ohh yes, its all good. I was writing that without evening reading your new code of `circuit` yet. I was trying to first read the application part of the PR and then go back to the core design to see if it agrees with what I have in mind. We were thinking about the same thing. Sorry for the confusion. Looks all good!

src/qutip_qip/circuit/circuit.py
Mayank447: `QubitCircuit.gates` was a public attribute, so changing it was allowed that's why codebase had several instance of `qc.gates.append()`, `qc.gates.extend()`. I am just throwing a warning in case someone is trying to do `qc.gates = [...]` which was previously allowed.

src/qutip_qip/circuit/circuit.py
Mayank447: Global Phase is not being applied as it is `QubitCircuit` class property now.

src/qutip_qip/compiler/instruction.py
Mayank447: I have made this change now (renamed to `PulseInstruction` and deprecated the `Instruction` class via warning).

Mayank447:
> Looks good to me. Let me know when you find it ready.

The PR is ready, let me know if there are any other changes needed.

src/qutip_qip/circuit/circuit.py
BoxiLi: But this function returns the unitary representation of the full circuits. Basically this should give the same result as simulating the the circuit with a given input. In this case all possible basis states. Should the phase not be included here?

src/qutip_qip/circuit/circuit.py
BoxiLi: Wait I am a bit confused.
I see here 
```
@property
def gates(...)
    ...

gates.setter

def gates(...)
    ...
```

should it not be
```
@property
def gates(...)
    ...

@gates.setter
def gates(...)
    ...
```

Mayank447: Yeah, I didn't notice the extra line mb.

src/qutip_qip/circuit/circuit.py
Mayank447: I ignored global phase for two reasons. First they aren't physically relevant (obvious for density matrix and in case of statevector $<Ψ|U\dagger MU|Ψ>$ where $U|\psi>$ is our state after evolution, the global phase cancels). Second all the `GLOBALPHASE` gate now only add phase to the `QubitCircuit` they aren't appended to `QubitCircuit.instructions`.

We can still return the global phase of the circuit in propogators, might be helpful for equivalence checking. In that case do you want it to be first element of `U_list` or the last one.

BoxiLi: You are totally right. Global phase is not physically relevant. I was more thing of if one should get exactly the same state vector. Indeed it might be hepful for equivalence checking.

First or last doesn't make a difference to me. Maybe last then.

Mayank447: I have made the change and updated in the docstring.

PR #321, opened 2026-02-15
Add benchmark suite for circuit simulation performance evaluation

vishwas-droid:
This PR introduces a minimal and reproducible benchmark framework
for evaluating circuit simulation performance in qutip-qip.

Features:
- GHZ state preparation circuit
- Random layered circuit (single-qubit rotations + CNOT chain)
- |0...0> initial state preparation
- Warmup run to reduce first-call overhead effects
- Averaged runtime measurement over multiple runs
- Demonstrates scaling with increasing qubit count

The benchmark can be executed with:

    python -m benchmarks.run_benchmarks

No public API changes.
No behavioral modifications.

This benchmark infrastructure is intended to support future
performance comparisons between sparse and tensor-based
execution paths.

### DIFF
--- a/benchmarks/run_benchmarks.py
+++ b/benchmarks/run_benchmarks.py
@@ -8,9 +8,24 @@ def zero_state(n_qubits: int):
     return tensor([basis(2, 0) for _ in range(n_qubits)])
 
 
-def benchmark(circuit_fn, n_qubits: int, runs: int = 20):
-    """Benchmark average execution time for a circuit."""
-    qc = circuit_fn(n_qubits)
+def benchmark(circuit_fn, circuit_params: dict, runs: int = 20):
+    """
+    Benchmark average execution time for a circuit.
+
+    Parameters
+    ----------
+    circuit_fn : callable
+        Function that builds the circuit.
+    circuit_params : dict
+        Dictionary containing circuit inputs (e.g. n_qubits, depth, etc.).
+    runs : int
+        Number of timing runs.
+    """
+    qc = circuit_fn(**circuit_params)
+
+    n_qubits = circuit_params.get("n_qubits")
+    if n_qubits is None:
+        raise ValueError("circuit_params must include 'n_qubits'.")
 
     # Warmup run to avoid first-call overhead distortion
     qc.run(zero_state(n_qubits))
@@ -31,11 +46,10 @@ if __name__ == "__main__":
     print("Benchmarking circuit simulation performance\n")
 
     for n in [3, 4, 5, 6, 7, 8]:
-        ghz_time = benchmark(ghz_circuit, n)
-        rand_time = benchmark(random_layered_circuit, n)
+        ghz_time = benchmark(ghz_circuit, {"n_qubits": n})
+        rand_time = benchmark(random_layered_circuit, {"n_qubits": n})
 
         print(f"{n} qubits:")
         print(f"  GHZ circuit: {ghz_time:.6f} sec")
         print(f"  Random layered circuit: {rand_time:.6f} sec")
         print()
-


benchmarks/run_benchmarks.py
```diff
@@ -0,0 +1,41 @@
...
+from qutip import basis, tensor
+from .circuits import ghz_circuit, random_layered_circuit
+
+
+def zero_state(n_qubits: int):
+    """Return |0...0> initial state."""
+    return tensor([basis(2, 0) for _ in range(n_qubits)])
+
+
+def benchmark(circuit_fn, n_qubits: int, runs: int = 20):
```
BoxiLi: To make it general, what about using `circuit_params` argument, which is a dictionary that packs all the circuit input including n_qubits, depth etc. Ideally, this function should be compatible with different circuit test case.

benchmarks/README.md
```diff
@@ -0,0 +1,22 @@
+# Circuit Simulation Benchmark Suite
```
BoxiLi, line 1: Will it attract more attention if we add this page to the web documentation?

BoxiLi:
Also, is there a way to test this in GitHub action? We need to make sure that it is working. Maybe we can add it as a separate action?

hodgestar:
@BoxiLi We also add this to https://github.com/qutip/qutip-benchmark instead and re-use the infrastructure there (e.g. run nightly and plot results over time on the website).

benchmarks/README.md
vishwas-droid: Yes, I think adding a short section to the web documentation would make it more visible and useful for users interested in performance evaluation.
I can prepare a concise documentation section explaining how to run the benchmarks and interpret the results.

vishwas-droid:
@BoxiLi Thanks for raising this.

For CI, we could add a lightweight benchmark run (e.g., small qubit count) just to ensure it executes correctly without introducing significant overhead.

For more comprehensive performance tracking (nightly runs, historical plots, etc.), integrating with qutip-benchmark seems like the right long-term direction as suggested.

For now, this PR focuses on providing a minimal and reproducible benchmark framework within qutip-qip. I’m happy to follow up with CI integration or adapt it to the dedicated benchmarking repository based on your preference.

BoxiLi:
@hodgestar thanks this is a good point. 

Is it possible to add this instead to qutip-benchmark? Or, maybe the core test code here but call them from qutip-benchmark?

Adding a CI suite here will likely add a lot of duplications indeed. It is better if we can fit this into the qutip-benchmark workflow.

vishwas-droid:
That makes sense. In that case, I’m happy to move the benchmarking workflow to qutip-benchmark and adapt it to fit the existing infrastructure there.

Would you prefer keeping only the core utilities here and running the full benchmarks from the dedicated repository?

BoxiLi:
I think it is probably the best if we have all the benchmarking code in one place, to reduce duplication. Let's try and see we can add a new scaling plot page to qutip-benchmark, with exactly those circuit test cases you have here.

The benchmarking code in this PR looks good to me.

vishwas-droid:
Sounds good, thanks! I’ll prepare a PR for qutip-benchmark with these circuit test cases and scaling plots.

vishwas-droid:
@BoxiLi I’ve opened a corresponding PR in qutip-benchmark adding the GHZ scaling benchmark:

https://github.com/qutip/qutip-benchmark/pull/68

It integrates with the existing benchmarking workflow. Please let me know if any adjustments are needed.

Mayank447:
@vishwas-droid Can you close this PR since it has been moved to qutip-benchmark

PR #317, opened 2026-01-28
Replace n with num qubits

niranjannagumalli:
**Changes:**
- Depreciated N in `QubitCircuit` and replaced it with `num_qubits` instead.
- Replaced `qc.N` with `qc.num_qubits` in the codebase
- Depreciated `N` in Gatecompiler.
- Replaced N with `num_qubits` in TexRenderer (N was not an input argument btw)

PR #311, opened 2026-01-19
Removed unused variables and minor corrections

mudit06mah:
## Summary
Removed Unused Variables and corrected `projector1` and `projector2`.

## Issue
Fixes #304 

## Note
I also ran pytest tests but it appears a lot of tests unrelated to the `zzcrosstalk` are failing, should I open issues for each?

PR #325, opened 2026-02-22
Refactor Gate class (1)

### LINKED ISSUE #296: Depreciate SNOT and replace with H gate

Mayank447:
- Merge changes from master branch (#311 and #317)
- Replaced `get_compact_qobj` with `get_qobj` and deprecated the former.
- Added `is_clifford`, `self_inverse` class attributes to `Gate` class.
- Moved `num_ctrls` class attribute, from Gate class to `ControlledGate` Gate class.
- Renamed `ParametrizedGate` class to ParametricGate`.
- Added `num_params` class attributes, `validate_parameters` method  to `Parametric` Gate class.
- Removed `ControlledParamGate` class and included that functionality in `ControlledGate` class itself.
- Added a metaclass for `Gate` to disallow modification of Gate class attribution (since they are shared among all object instances).
- Added `__init_subclass__` in `Gate`, `ParametricGate`, `ControlledGate` class and several checks on class attribute type and value.
- Rename gate factory functions to `unitary_gate`, `controlled_gate`.
- Spilt standard gates into `operations/std/` dir, later renamed to `/operations/gates` dir in #328.
- Deprecated redundant gates (Closes #296)
- Deprecated string gate input, control_value, arg_value and arg_label in `add_gate` method.
- Deprecated `gates.py`
- Added typing.py
- Minor fixes to qiskit backend, algorithms.

### DIFF
--- a/doc/source/qip-basics.rst
+++ b/doc/source/qip-basics.rst
@@ -36,7 +36,7 @@ A circuit with the various gates and registers available is demonstrated below:
   from qutip_qip.operations import X, CX, SWAP
   from qutip import tensor, basis
 
-  qc = QubitCircuit(N=2, num_cbits=1)
+  qc = QubitCircuit(2, num_cbits=1)
 
   qc.add_gate(SWAP, targets=[0, 1])
   qc.add_measurement("M0", targets=[1], classical_store=0) # measurement gate
--- a/src/qutip_qip/qiskit/utils/converter.py
+++ b/src/qutip_qip/qiskit/utils/converter.py
@@ -101,7 +101,8 @@ def convert_qiskit_circuit_to_qutip(
         )
 
     qutip_circuit = QubitCircuit(
-        N=qiskit_circuit.num_qubits, num_cbits=qiskit_circuit.num_clbits
+        num_qubits=qiskit_circuit.num_qubits,
+        num_cbits=qiskit_circuit.num_clbits,
     )
 
     qutip_circuit.name = qiskit_circuit.name
--- a/doc/pulse-paper/customize.py
+++ b/doc/pulse-paper/customize.py
@@ -51,7 +51,7 @@ class MyModel(Model):
         elif label[:2] == "sy":
             return 2 * np.pi * sigmay() / 2, [targets]
         else:
-            raise NotImplementError("Unknown control.")
+            raise ValueError("Unknown control.")
 
 
 class MyCompiler(GateCompiler):
--- a/doc/source/qip-simulator.rst
+++ b/doc/source/qip-simulator.rst
@@ -16,7 +16,7 @@ examples of circuit evolution. We take a circuit from
     from qutip_qip.circuit import QubitCircuit
     from qutip_qip.operations import X, CX, CH, QASMU, TOFFOLI
     )
-    qc = QubitCircuit(N=3, num_cbits=3)
+    qc = QubitCircuit(3, num_cbits=3)
     qc.add_gate(QASMU, targets=[0], arg_value=[1.91063, 0, 0])
     qc.add_gate(CH, controls=[0], targets=[1])
     qc.add_gate(TOFFOLI, targets=[2], controls=[0, 1])
--- a/src/qutip_qip/algorithms/bit_flip.py
+++ b/src/qutip_qip/algorithms/bit_flip.py
@@ -84,7 +84,7 @@ class BitFlipCode:
 
         total_qubits = max(data_qubits + syndrome_qubits) + 1
         classical_bits = len(syndrome_qubits)
-        qc = QubitCircuit(N=total_qubits, num_cbits=classical_bits)
+        qc = QubitCircuit(num_qubits=total_qubits, num_cbits=classical_bits)
 
         dq = data_qubits
         sq = syndrome_qubits
--- a/src/qutip_qip/algorithms/phase_flip.py
+++ b/src/qutip_qip/algorithms/phase_flip.py
@@ -86,7 +86,7 @@ class PhaseFlipCode:
             raise ValueError("Expected 3 data qubits and 2 syndrome qubits.")
 
         total_qubits = max(data_qubits + syndrome_qubits) + 1
-        qc = QubitCircuit(N=total_qubits, num_cbits=2)
+        qc = QubitCircuit(num_qubits=total_qubits, num_cbits=2)
 
         dq = data_qubits
         sq = syndrome_qubits
--- a/src/qutip_qip/algorithms/shor_code.py
+++ b/src/qutip_qip/algorithms/shor_code.py
@@ -20,7 +20,7 @@ class ShorCode:
         Returns:
             QubitCircuit: Circuit that encodes one logical qubit into the Shor code.
         """
-        qc = QubitCircuit(N=self.n_qubits)
+        qc = QubitCircuit(num_qubits=self.n_qubits)
 
         # Step 1: Bit-flip encode qubit 0 → [0, 1, 2]
         bit_code = BitFlipCode()
--- a/src/qutip_qip/circuit/circuit.py
+++ b/src/qutip_qip/circuit/circuit.py
@@ -46,7 +46,7 @@ class QubitCircuit:
 
     Parameters
     ----------
-    N : int
+    num_qubits : int
         Number of qubits in the system.
     input_states : list
         A list of string such as `0`,'+', "A", "Y". Only used for latex.
@@ -60,31 +60,41 @@ class QubitCircuit:
 
     def __init__(
         self,
-        N,
+        num_qubits=None,
         input_states=None,
         output_states=None,
         reverse_states=True,
         dims=None,
         num_cbits=0,
         user_gates=None,
+        N = None
     ):
         # number of qubits in the register
-        self.N = N
+        self._num_qubits = num_qubits
+        if N is not None:
+            warnings.warn(
+                "The 'N' parameter is deprecated. Please use "
+                "'num_qubits' instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            self._num_qubits = N
+
         self.reverse_states = reverse_states
         self.num_cbits: int = num_cbits
         self._global_phase: float = 0.0
         self._instructions: list[CircuitInstruction] = []
-        self.dims = dims if dims is not None else [2] * N
+        self.dims = dims if dims is not None else [2] * self.num_qubits
 
         if input_states:
             self.input_states = input_states
         else:
-            self.input_states = [None for i in range(N + num_cbits)]
+            self.input_states = [None for i in range(self.num_qubits + num_cbits)]
 
         if output_states:
             self.output_states = output_states
         else:
-            self.output_states = [None for i in range(N + num_cbits)]
+            self.output_states = [None for i in range(self.num_qubits + num_cbits)]
 
         if user_gates is not None:
             raise ValueError(
@@ -121,6 +131,26 @@ class QubitCircuit:
     def instructions(self) -> list[CircuitInstruction]:
         return self._instructions
 
+    @property
+    def num_qubits(self) -> int:
+        """
+        Number of qubits in the circuit.
+        """
+        return self._num_qubits
+
+    @property
+    def N(self) -> int:
+        """
+        Number of qubits in the circuit.
+        """
+        warnings.warn(
+            "The 'N' parameter is deprecated. Please use "
+            "'num_qubits' instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self._num_qubits
+
     def __repr__(self) -> str:
         return ""
 
@@ -140,7 +170,7 @@ class QubitCircuit:
         state_type: str = "input", # FIXME Add an enum type hinting?
     ):
         """
-        Add an input or ouput state to the circuit. By default all the input
+        Add an input or output state to the circuit. By default all the input
         and output states will be initialized to `None`. A particular state can
         be added by specifying the state and the qubit where it has to be added
         along with the type as input or output.
@@ -180,7 +210,7 @@ class QubitCircuit:
         Parameters
         ----------
         measurement: string
-            Measurement name. If name is an instance of `Measuremnent`,
+            Measurement name. If name is an instance of `Measurement`,
             parameters are unpacked and added.
         targets: list
             Gate targets
@@ -298,8 +328,8 @@ class QubitCircuit:
         _check_iterable("classical_controls", classical_controls)
 
         # Checks each element is of given type (e.g. int) and within the limit
-        _check_limit_("targets", targets, self.N - 1, int)
-        _check_limit_("controls", controls, self.N - 1, int)
+        _check_limit_("targets", targets, self.num_qubits - 1, int)
+        _check_limit_("controls", controls, self.num_qubits - 1, int)
         _check_limit_(
             "classical_controls", classical_controls, self.num_cbits - 1, int
         )
@@ -376,7 +406,7 @@ class QubitCircuit:
         start : int
             The qubit on which the first gate is applied.
         """
-        if self.N - start < qc.N:
+        if self.num_qubits - start < qc.num_qubits:
             raise NotImplementedError("Targets exceed number of qubits.")
 
         for circuit_op in qc.instructions:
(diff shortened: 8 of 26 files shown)


BoxiLi:
Glad to see this coming! It seems like a very big collection. Are there ways to make it smaller? Maybe only the parts that resolve the master branch conflict?

Mayank447:
Hi,

Yes, the number of commits did considerably exceed my expectation for this PR. And there is still a little work left.

I can split this PR into two (one before merge master) and one after, this should help with reviewing. Any changes as per the code review for the first half, I will do in the second PR to avoid any merge conflicts. How does it sound?

Also the major code changes are operations/, circuit.py and test_gates.py, it's better to review those files first.

BoxiLi:
>  Any changes as per the code review for the first half, I will do in the second PR to avoid any merge conflicts. How does it sound?

Yes, that sounds better. The merging one will require less code review.

Mayank447:
@BoxiLi I have splitted this PR into two. The failing tests in this PR are because of pytest raising error for DeprecationWarning, you can ignore that. Everything passes in #328 (the second PR).

This PR is ready for review.

src/qutip_qip/operations/gateclass.py
```diff
@@ -257,6 +258,9 @@ def __init__(
                     "target_gate must be provided either as argument or class attribute."
                 )
 
+        if type(arg_value) is float:
+            arg_value = [arg_value]
+
```
BoxiLi, line 263: `if type(arg_value) is float or type(arg_value) is np.float64:`

BoxiLi: Can we find a more consistent way of checking? Maybe if arg_value is not iterable? What if someone gives an integer?

src/qutip_qip/operations/gateclass.py
```diff
@@ -146,14 +150,26 @@ def _repr_latex_(self):
...
 
 # Make this an abstract class
-class ControlledGate(Gate):
-    num_qubits = 2
+class ControlledGate(ABC, Gate):
+    num_qubits = 1
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+        if inspect.isabstract(cls): # Skip the num_ctrl_qubits check for an abstract class
+            return
```
BoxiLi, line 158: Why skip?

doc/source/qip-basics.rst
```diff
@@ -36,7 +36,7 @@ A circuit with the various gates and registers available is demonstrated below:
   from qutip_qip.operations import Gate
   from qutip import tensor, basis
 
-  qc = QubitCircuit(num_qubits=2, num_cbits=1)
+  qc = QubitCircuit(2, num_cbits=1)
```
BoxiLi, line 39: Do we really want people to explicitly state `num_qubits` in every circuit definition? I feel like this may be a bit too cumbersome, so I think we dropped it in the doc examples.

src/qutip_qip/operations/std_gates.py
```diff
@@ -992,6 +992,7 @@ class GLOBALPHASE(ParametrizedGate):
 
     latex_str = r"{\rm GLOBALPHASE}"
     num_qubits: int = 0
+    num_ctrl_qubits: int = -1
```
BoxiLi, line 995: > num_ctrl_qubits: int = -1

?

src/qutip_qip/operations/gateclass.py
```diff
@@ -233,6 +237,12 @@ def __init__(self, arg_value: float, arg_label: str = None):
         self.arg_value = arg_value
         # self.validate_params()
 
+    @property
+    @abstractmethod
+    def num_params(self) -> Qobj:
```
BoxiLi, line 242: > def num_params(self) -> Qobj:

Should be int

src/qutip_qip/typing.py
```diff
@@ -0,0 +1,19 @@
...
+
+Int: TypeAlias = int | np.integer
+Real: TypeAlias = int | float | np.integer | np.floating
+Number: TypeAlias = int | float | complex | np.number
+
+IntList = Sequence[Int]
+RealList = Sequence[Real]
+ScalarList = Sequence[Number]
+
+ArrayLike = Sequence[any] | np.ndarray
```
BoxiLi, line 19: `ArrayLike = Sequence[Any] | np.ndarray`

Capital Any

tests/test_renderer.py
```diff
@@ -3,21 +3,35 @@
...
+    qc.add_gate(ISWAP, targets=[2, 3])
+    qc.add_gate(CRX(np.pi / 2), targets=[0], controls=[1])
+    qc.add_gate(SWAP, targets=[0, 3])
+    qc.add_gate(BERKELEY, targets=[0, 3])
+    qc.add_gate(FREDKIN, controls=[3], targets=[1, 2])
+    qc.add_gate(TOFFOLI, controls=[0, 2], targets=[1])
+    qc.add_gate(CX, controls=[0], targets=[1])
+    qc.add_gate(CRX(0.5), controls=[2], targets=[3])
+    qc.add_gate(SWAP, targets=[0, 3])
     return qc
```
BoxiLi, line 35: If we want to keep string gate as a feature, we should keep at least one test for that.

Personally I think it is convenient as it avoids the importation of all gates explicitly.

src/qutip_qip/qiskit/utils/converter.py
```diff
@@ -51,14 +56,13 @@ def get_qutip_index(bit_index: int | list, total_bits: int) -> int:
...
-    and so on. Essentially the bit order is reversed.
+    the 0st bit is mapped to the 0th bit and (n-1)th bit to (n-q)th bit
+    and so on. Essentially the bit order stays the same.
     """
-
     if isinstance(bit_index, list):
         return [get_qutip_index(bit, total_bits) for bit in bit_index]
     else:
-        return total_bits - 1 - bit_index
+        return bit_index
```
BoxiLi, line 66: Why remove this? I remember there was a reason for this reversion. The qiskit uses a different order, and that breaks stuff.

tests/test_gates.py
```diff
@@ -5,40 +5,8 @@
...
-    MS,
-    TOFFOLI,
-    FREDKIN,
-    BERKELEY,
-    R,
-    RZX,
-    expand_operator,
-)
+from qutip_qip.operations import Gate, expand_operator
+from qutip_qip.operations.std import *
```
BoxiLi, line 9: The wildcard importation is a bit dangerous. But it is fine here in the test. This is why I suggest we keep the string input for add_gate, otherwise the users will just use wildcard import.

doc/source/apidoc/qutip_qip.operations.rst
```diff
@@ -38,7 +37,6 @@ qutip\_qip.operations
       TOFFOLI
       FREDKIN
       CNOT
-      CSIGN
       CRX
```
BoxiLi, line 40: Can we not keep them and simply define them to be identical (a copy?) to H and CPHASE? So that old code don't break.

doc/source/apidoc/qutip_qip.operations.rst
```diff
@@ -28,15 +28,14 @@ qutip\_qip.operations
...
+      CX
       SQRTSWAP
       SQRTISWAP
       BERKELEY
       SWAPALPHA
       MS
       TOFFOLI
       FREDKIN
-      CNOT
       CRX
```
BoxiLi, line 39: I really think we should keep CNOT thought, if it doesn't cause any conflict. Not a big fan or CX haha.

src/qutip_qip/operations/gateclass.py
Mayank447: This line is actually incorrect, the way to test a float is via `try: float(arg_value)`. Second we should test for iterable as you mentioned.

These changes is are already there in second PR

src/qutip_qip/operations/gateclass.py
Mayank447: num_ctrl_qubits need not be defined for an abstract class. Also since that abstract class will be used somewhere, so we are anyways checking for that final gate user will use

doc/source/qip-basics.rst
Mayank447: This change is from #317, yes you are correct about us dropping it because of that reason

src/qutip_qip/operations/std_gates.py
Mayank447: Ignore this, num_ctrl_qubits is moved to `ControlledGate` in the next PR. So this line is no longer required and has been removed.

src/qutip_qip/operations/gateclass.py
Mayank447: Already corrected in the next PR

src/qutip_qip/typing.py
Mayank447: From python 3.10, it allows from `any`.

src/qutip_qip/qiskit/utils/converter.py
Mayank447: https://quantum.cloud.ibm.com/docs/en/guides/bit-ordering We are actually following the same bit ordering. 

I added a test for 3 qubit Quantum Circuit with exact same gate and qubit sequence for `qiskit.QuantumCircuit` and `qutip_qip.QubitCircuit`, and it passes with the new ordering.

tests/test_gates.py
Mayank447: Actually in the next PR I changed it as

`import qutip_qip.operations.std as std` from there the user can use the gates like `std.X` etc.

doc/source/apidoc/qutip_qip.operations.rst
Mayank447: The old code still works, making them identical copies messes up with deprecation because of Python's object's referencing. I did remove the deprecated class from documentation so that anyone new to qutip_qip doesn't use it.

doc/source/apidoc/qutip_qip.operations.rst
Mayank447: Actually we can initialise `CX(control_value=0)`, `CX(control_value=1)`. This behaviour is not so apparent for `CNOT` gate, that's why the renaming and also to make it consistent with `CY`, `CZ`.

Mayank447:
> Another thing I find slightly inconsistent is parameteric gates and std gates. For a parameteric gate, one calls `parametricgate(argvalue).get_qobj()`, for other gates like std, it is simply `stdgate.get_qobj`. A static method. I was wondering if it will cause confusion.

Ok there is a deeper reason for it. A Parametric Gate needs to be initialised with the parameter RX(0.5) only then can any gate method on it be called e.g. `get_qobj`, decomposition method etc. Actually the same holds true for the `ControlledGate` with `control_value`. But a gate like `X` doesn't need any argument or initialisation for that fact. So `X()` is unnecessary, it just creates more object instances even though all of them are same, and each still occupies a different memory segment. So `X` gate should not be instantiable,  and in the next PR by default in `__init__` we raise an error if someone tries to do so. `__init__` is overwritten in case of `ControlledGate`, `ParametricGate` as they require arguments to mean a gate.

Also static method does mean a property so it is `stdgate.get_qobj()` not `stdgate.get_qobj`.

doc/source/apidoc/qutip_qip.operations.rst
BoxiLi: Oh, I meant more like we keep CSIGN, SNOT, and CNOT as an alias, instead of deprecating them. Is that possible?

doc/source/apidoc/qutip_qip.operations.rst
BoxiLi: Ok, I see, CX is more general in this sense.

Mayank447:
> I was wondering about the name `std` can we find something more user-friendly? Or simply encourage people to import from the gateclass module.
> 
> ```
> from qutip.operations import gateclass
> gateclass.X
> ```

Yes, we can always rename `std` to something better (coming for a C/C++ background I agree even more). Unfortunately `gateclass` is a file so it won't work. How about we rename the following  `gates.py` -> `old_gates.py` and `std/` -> `gates/`, `gateclass.py` remains the way it is.
 
```
from qutip.operations import gates

qc = QubitCircuit(2)
qc.add_gate(gates.X, 0)
```

> Though I still think it would be convenient to keep the string input.

I should have explained why I am in the favour of deprecating string input in the PR description. First this create inconsistency, "X" is fine but something like "RX" requires an `arg_value` which is really a property of the gate but is being handled in `add_gate` method, same holds true for `control_value`.  Since we now allow custom gates via inheritance, user has the freedom to have his own arguments for the gate. So instead of us trying to handle these nuances in the `add_gate` method for string input, we make the user handle it himself. Also in my opinion, `qc.add_gate(RX(0.5), 0)` is cleaner than `qc.add_gate("RX", 0, arg_value=0.5)` (the user will have to use `arg_value` here because in the argument ordering `controls` comes before it).

Second thing, I agree string inputs are convenient. But now the user can do something like
```
import qutip_qip.operations.std as std

qc = QubitCircuit(2)
qc.add_gate(std.X, 0)
qc.add_gate(std.RX(0.5), 1)
```
so his string inputs only change from "gate" to `std.gate`.

Third thing, in the second PR I introduce the concept of namespaces. So each gate must belong to a namespace (default is std, not to be confused with the dir `operations/std/`). There are several reasons for it, first at several places we use `gate.name` for comparison. A user can define his own `RX` gate (since different conventions are still prevalent) and it will mess up our current decomposition routines entirely. A namespace will prevent any sort of name collision, he can still define his own `RX` but in a different namespace and we will compare `RX` wrt. the std namespace. Second the concept  of namespaces is much broader and powerful, e.g. I can have std namespace for all these standard gates and another namespace for say noisy gates, or one for math components (Adders, Multipliers, etc.).

doc/source/apidoc/qutip_qip.operations.rst
Mayank447: Aliasing is possible. There is just one nuance from the user's perspective.

`qc.add_gate(SNOT, 1)` would render the gate as H in the circuit plot than SNOT. Same happens if he tries to print gate list etc.

BoxiLi:
> Yes, we can always rename `std` to something better (coming for a C/C++ background I agree even more). Unfortunately `gateclass` is a file so it won't work. How about we rename the following  `gates.py` -> `old_gates.py` and `std/` -> `gates/`, `gateclass.py` remains the way it is.

This might work, but let's make it in a separate PR at the end.

> I should have explained why I am in the favour of deprecating string input in the PR description....

Ok, I am convinced. `gates.X` should be sufficiently simple. I agree. Just one thing. We need to keep the backward compatibility for this in this PR. I think there are quite a few people using this interface. We should raise a deprecation warning for this first.

Mayank447:
> Ok, I am convinced. `gates.X` should be sufficiently simple. I agree. Just one thing. We need to keep the backward compatibility for this in this PR. I think there are quite a few people using this interface. We should raise a deprecation warning for this first.

Backward compatibility and the deprecation warning are already there in this PR.

doc/source/apidoc/qutip_qip.operations.rst
BoxiLi: > qc.add_gate(SNOT, 1) would render the gate as H in the circuit plot rather than SNOT. 

I think that is fine if they really want a different symbol. They can define a custom gate. I'd vote for keeping CSIGN, SNOT, and CNOT through aliasing.

BoxiLi:
Ok great, I missed this. If `circuit.add_gate("X", ... )` still works for this version and raises a warning, we are good to go.

Mayank447:
@BoxiLi you can go ahead with the merge

doc/source/apidoc/qutip_qip.operations.rst
Mayank447: Aliasing has been done in #328

src/qutip_qip/qiskit/utils/converter.py
Mayank447: You were right about this, the reason my test cases passed was because we started in |000> state, which is the default.

The change for the same has been made in #328

Issue #326, opened 2026-02-23
arg_label not being rendered properly


Mayank447:
Currently the argument label for Parametric Gate e.g. `RZ(arg_value=np.pi, arg_label='\pi')` isn't plotted as a symbol for the default `MatRenderer` as can be seen on the tutorial website https://nbviewer.org/urls/qutip.org/qutip-tutorials/tutorials-v5/quantum-circuits/quantum-gates.ipynb. Also the gate name is missing when arg_label is provided.

PR #327, opened 2026-02-24
Fix: render arg_label properly in Parametric gates and show gate name

suryanshcode2601:
Currently, in MatRenderer the arg_label of Parametric gates (e.g., RZ(arg_value=np.pi, arg_label='π')) is not displayed, and if arg_label is provided, the gate name sometimes disappears.

This PR fixes the issue by:

Ensuring the gate name is always rendered

Displaying arg_label if provided, falling back to arg_value otherwise

Wrapping symbolic labels (like π, π/2) in LaTeX formatting for proper rendering

Supporting both single-qubit and multi-qubit gates

### DIFF
--- a/src/qutip_qip/circuit/draw/mat_renderer.py
+++ b/src/qutip_qip/circuit/draw/mat_renderer.py
@@ -625,20 +625,7 @@ class MatRenderer(BaseRenderer):
                 horizontalalignment="center",
                 zorder=self._zorder["gate_label"],
             )
-            # gate_text = plt.Text(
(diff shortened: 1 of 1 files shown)


Mayank447:
Hi @suryanshcode2601

Thanks for this PR. As mentioned in the issue #326 currently it is not open as large scale `Gate` class refactoring is under way and will be finished in a week or 10 days. That refactor does lead to considerable change in code for circuit rendering.

suryanshcode2601:
okkkk, i would love to work on it if it opens again

PR #328, opened 2026-02-25
Refactor Gate class (2)

### LINKED ISSUE #301: Add a gate inversion function
Should return the inverse gate for basic gates like S, T, H etc.

For algorithm (composite of multiple gates) like QFT, should give the inverse by recursively utilising the above logic

Edit: This issue has already been resolved in a separate branch. It shall be closed when that branch is merged into master soon

Mayank447:
The original PR #325 was split into 2, for the previous set of changes look at that PR description.

- Defined `__slots__` for all the Abstract Gate classes and standard gate classes for better memory efficiency and faster lookup time.

- Added boolean `is_controlled`, `is_parametric` method to the `Gate` class.

- Added the concept and implementation of namespaces for operations, this is done because at several places earlier gate.name was being used for comparision. Each operation (gate for now) must now be defined in a namespace or if it is a temporary gate, namespace can be set to None. This allows the user to define faulty version of standard gates or follow their own convention for certain Parametric gates. The concept of namespace is much broader. Currently the parent namespace is "std", under it only "gates" is set as of now. In future we can make other namespaces under "std" like for Arithmetic Ops, Ansatz Ops etc. For the idea of Ops look at https://github.com/qutip/qutip-qip/issues/263#issuecomment-3753565184 Though it hasn't been implemented yet. In a way std namespace will work like a standard library for different operations and the user can also define his own implementation for operations in a different namespace.

- Added caching in `get_qobj` method for standard gates. This is done for two reasons.
```
@staticmethod
def get_qobj() -> Qobj:
    return Qobj([[1, 0], [0, -1]])
```
if you run it multiple times and check `id()` for each obj, they will be different. This leads to higher memory consumption for large circuit simulations, even though all of them represent the Z gate's qobj in this case. Additonally different qobjs will each need to be garbage collected. Memory and lookup is an even larger bottleneck in VQA circuits where several times many gates share the same parameter.  

- Added `dtype` to `get_qobj(dtype="dense")` method, . This might be useful if the circuit simulation in run on a different backend like JAX, CUDA.

- Made `Gate` non initialised by default i.e. `__init__` raise will raise an Error. Obviously`ParametricGate` and controlled version of them are initialisable. This makes sense because target, controls have been removed from gates and `x1=X()`, `x2=X()` now represent the same thing. Also good from memory point of view.

- Added `Sdag`, `Tdag`, `SQRTXdag`, `ISWAPdag`, `SQRTSWAPdag`, `BERKELEYdag` to the standard gates. They have been added to the documentation table and in circuit draw (colour config for them).

- Added `inverse` method for `Gate` class. Also implemented inverse for all standard gates and tests for the same. For controlled gates, the inverse is auto calculated based on the target gate. Closes #301 

- Renamed `unitary_gate` to `get_unitary_gate` and added checks for all the input arguments. 

- The control value for a gate is now baked in to the control gate itself. The user can always generate his own version of controlled gate e.g. `controlled(gates.X, num_ctrl_qubits=2, control_value=0b10)`. For more detail check out https://github.com/qutip/qutip-qip/discussions/308#discussioncomment-15950921

- For all Controlled gates, the (target_gate, num_ctrl_qubit, ctrl_value) is used as another key in the namespace to reference to that controlled gate. This is done in order to prevent regeneration of controlled gate class for a given target gate with known (num_ctrl_qubit, ctrl_value). Since namespace is used by default, there isn't any overhead.

- Renamed `controlled_gate` to `controlled_gate_unitary` and resolved the bug which didn't allow for generating the correct qobj if number of controls > 1.

- Added gate equivalence checking for `ParametricGate`, `ControlledGate`. And added `__eq__`, `__hash__` in Gate metaclass as well. Replaced gate.name = "X" search with gate == X now.

- Corrected inconsistency between Gate and Type[Gate] from previous PRs. Gate refers to an instantiated object while type[Gate] to a Gate class or subclass in accordance with Python's typehinting.

- Added several more checks in `__init_subclass__`, `add_gate` method.

- Added tests in `test_gates.py` and `test_circuit.py` mostly incorrect cases like num_qubits being negative etc. which should raise an Error. Since Gates are used through the codebase, their normal usage is already mostly covered.

**Maintainence and Bug Fixes:**

- `gate_product` had a clear bug that `tensor_list` was not defined but was being used, fixed that and also moved the entire `gate_product` logic to operations which was split in circuit-simulator utils, operations but was always being imported from operations as per `__init__` even in `mat_mul_simulator.py`.

- Removed depreciated arguments, and functions which were depreciated 3 years back. This was checked in the git commit history.

- Remove mutable default values from function arguments across the codebase, this is a common Python bug. Something like targets = [], can often lead to error. For details check out this blog https://medium.com/@matbrizolla/beware-of-the-hidden-trap-understanding-mutable-default-arguments-in-python-d10e8b0a7b72

- Corrected several `isInstance` checks in the codebase. `Iterable` and `Sequence` were being used interchangeably at several places especially in qasm. This led to several unexpected behaviour and errors during gate refactoring.

- Moved pytest.ini to the root, this is what pytest themselves recommend https://docs.pytest.org/en/stable/explanation/goodpractices.html#choosing-a-test-layout-import-rules . The reason was placing pytest.ini in `tests/` lead to certain sys path issues, for reference check https://docs.pytest.org/en/stable/explanation/pythonpath.html. This caused an error when standard gates were designated to a namespace and pytest was trying to redefine them and reassign them to the same namespace.

*Edit: Some more changes*

- Separated IDLE (meant for Pulse level simulation) and Identity gate (standard Identity matrix).

- In Pulse compiler instead of `gate.name`, `gates` are themselves being used as keys.

- Removed expand argument from `inverse()` in Parametric Gates.

- Renamed `controlled` to `get_controlled_gate`.

### DIFF
--- a/src/qutip_qip/operations/gates/single_qubit_gate.py
+++ b/src/qutip_qip/operations/gates/single_qubit_gate.py
@@ -605,6 +605,7 @@ class QASMU(_SingleQubitParametricGate):
         theta, phi, gamma = self.arg_value
         return QASMU(-theta, -gamma, -phi)
 
+
 class IDLE(AngleParametricGate):
     """
     IDLE gate.
@@ -633,4 +634,4 @@ class IDLE(AngleParametricGate):
         # Practically not required as this gate is only useful in pulse level
         # simulation, and the pulse compiler implementation of it will be
         # independent of get_qobj()
-        return qeye(2, dtype=dtype) 
+        return qeye(2, dtype=dtype)
--- a/src/qutip_qip/operations/namespace.py
+++ b/src/qutip_qip/operations/namespace.py
@@ -6,7 +6,8 @@ from dataclasses import dataclass, field
 
 class _SingletonMeta(type):
     """
-    Note this is not a thread-safe implementation of Singleton.
+    Metaclass to implement the Singleton design pattern.
+    Note that this is not a thread-safe implementation of a Singleton.
     """
 
     _instances = {}
@@ -19,21 +20,42 @@ class _SingletonMeta(type):
 
 
 class _GlobalNameSpaceRegistry(metaclass=_SingletonMeta):
+    """
+    Global registry to manage and store all active namespaces.
+
+    This class enforces a singleton pattern (using the metaclass) to ensure that only one global
+    registry exists during the runtime of the application.
+    """
+
     def __init__(self):
         self._registry: set[NameSpace] = set()
 
     def register_namespace(self, namespace: NameSpace) -> None:
-        """Safely adds an item to the specific namespace."""
+        """
+        Safely adds a new namespace to the global registry.
+
+        Note: This means that a gate (or operation) is never garbage
+        collected until the Namespace is destroyed. This is the desired
+        behavior for standard library gates.
+
+        Parameters
+        ----------
+        namespace : NameSpace
+            The namespace instance to be registered.
+
+        Raises
+        ------
+        ValueError
+            If the namespace already exists within the registry.
+        """
         if namespace in self._registry:
             raise ValueError(f"Existing namespace {namespace}")
 
-        # Note: This does mean that gate (or operation) is never garbage
-        # collected until the Namespace exists. This is fine for standard gates.
         self._registry.add(namespace)
 
         # Default behaviour for user defining his own gates is that namespace is None,
         # Thus those gates are considered temporary by default, we use the same logic in
-        # QPE for Controlled Unitary gates, VQA (until Ops i.e. composite gates are introduced).
+        # QPE for Controlled Unitary gates, VQA untils Ops are implemented.
 
 
 _GlobalRegistry = _GlobalNameSpaceRegistry()
@@ -41,11 +63,31 @@ _GlobalRegistry = _GlobalNameSpaceRegistry()
 
 @dataclass
 class NameSpace:
+    """
+    Represents a distinct, optionally hierarchical namespace for registering
+    quantum operations.
+
+    Parameters
+    ----------
+    local_name : str
+        The local identifier for the namespace. Must not contain periods ('.').
+    parent : NameSpace or None, optional
+        The parent namespace, if this is a nested sub-namespace. Default is None.
+    """
+
     local_name: str
     parent: NameSpace | None = None
     _registry: dict[str, any] = field(default_factory=dict)
 
     def __post_init__(self):
+        """
+        Validates the namespace name and registers it globally upon creation.
+
+        Raises
+        ------
+        ValueError
+            If `local_name` contains a dot, as dots are reserved for hierarchy.
+        """
         if "." in self.local_name:
             raise ValueError(
                 f"Namespace local_name '{self.local_name}' cannot contain dots. "
@@ -55,6 +97,10 @@ class NameSpace:
 
     @cached_property
     def name(self) -> str:
+        """
+        str: The fully qualified, hierarchical name of the namespace.
+        (e.g., 'std.gates').
+        """
         if self.parent:
             return f"{self.parent.name}.{self.local_name}"
         return self.local_name
@@ -62,9 +108,22 @@ class NameSpace:
     def register(
         self, name: str | tuple[str, int, int], operation_cls: any
     ) -> None:
-        """Safely adds an item to the specific namespace.
-        name is str for a non-controlled Gate
-        name is a tuple (target_gate.name, num_ctrl_qubits, ctrl_values) for a Controlled gate.
+        """
+        Safely adds an item to this specific namespace.
+
+        Parameters
+        ----------
+        name : str or tuple of (str, int, int)
+            The identifier for the operation. Use a string for a non-controlled
+            gate. Use a tuple `(target_gate.name, num_ctrl_qubits, ctrl_values)`
+            as a second key for controlled gates.
+        operation_cls : any
+            The operation class or object to register.
+
+        Raises
+        ------
+        NameError
+            If an operation with the given name already exists in this namespace.
         """
         if name in self._registry:
             raise NameError(
@@ -73,11 +132,37 @@ class NameSpace:
         self._registry[name] = operation_cls
 
     def get(self, name: str | tuple[str, int, int]) -> any:
+        """
+        Retrieves a registered item from the namespace.
+
+        Parameters
+        ----------
+        name : str or tuple of (str, int, int)
+            The identifier of the registered operation.
+
+        Returns
+        -------
+        any
+            The registered operation class or object, or None if it is not found.
+        """
         if name not in self._registry:
             return None
         return self._registry[name]
 
     def _remove(self, name: str | tuple[str, int, int]) -> None:
+        """
+        Removes an item from the namespace registry.
+
+        Parameters
+        ----------
+        name : str or tuple of (str, int, int)
+            The identifier of the operation to remove.
+
+        Raises
+        ------
+        KeyError
+            If the specified name does not exist in the namespace.
+        """
         if name not in self._registry:
             raise KeyError(
                 f"{name} does not exists in namespace '{self.name} "
@@ -91,13 +176,16 @@ class NameSpace:
         return str(self)
 
     def __eq__(self, other) -> bool:
+        """Checks equality based on the full namespace name."""
         if type(other) is not NameSpace:
             return False
         return self.name == other.name
 
     def __hash__(self) -> int:
+        """Hashes the namespace based on the full namespace name."""
         return hash(self.name)
 
 
+# NS stands for namespace, std for Standard
 NS_STD = NameSpace("std")  # DEFAULT NAMESPACE
 NS_GATE = NameSpace("gates", parent=NS_STD)  # Default Gate Namespace


Mayank447:
@BoxiLi Although this PR is still draft (the failing tests is due to an internal server error in Coveralls), I guess the service is down. All the tests do pass.

src/qutip_qip/operations/gateclass.py
```diff
@@ -324,7 +328,7 @@ class ControlledGate(Gate):
         * If the gate should execute when the 0-th qubit is $|1\rangle$, 
             set ``control_value=1``.
         * If the gate should execute when two control qubits are $|10\rangle$ 
-            (binary 10), set ``control_value=2``.
+            (binary 10), set ``control_value=0b10``.
```
BoxiLi: Will `control_value=2` and `control_value=0b10` give the same result? If so, it would be actually useful to show them both.

src/qutip_qip/operations/gates.py
```diff
@@ -1120,18 +1120,6 @@ def globalphase(theta, N=1):
...
-    Calculate the bit-wise Hamming distance of x from 0: That is, the number
-    1s in the integer x.
-    """
-    tot = 0
-    while x:
-        tot += 1
-        x &= x - 1
-    return tot
-
-
```
BoxiLi, line 1188: This is removed because it is not used anymore by any other functions?

src/qutip_qip/vqa.py
```diff
@@ -127,6 +127,7 @@ def construct_circuit(self, angles):
         circ = QubitCircuit(self.num_qubits)
         i = 0
         for layer_num in range(self.num_layers):
+            Gate.clear_cache("vqa")
             for block in self.blocks:
                 if block.initial and layer_num > 0:
```
BoxiLi, line 138: This is an interesting problem; in many generated algorithms, it is very common to define gate blocks. Maybe we should add some UID automatically instead of doing this cache and check?

src/qutip_qip/operations/gateclass.py
```diff
@@ -349,26 +365,34 @@ def __str__(self):
             arg_label={self.arg_label}),
         """
 
+    def __eq__(self, other) -> bool:
+        if type(self) is not type(other):
+            return False
+
+        if self.arg_value != other.arg_value:
+            return False
```
BoxiLi: Will this work for float number?

src/qutip_qip/operations/gateclass.py
```diff
@@ -547,57 +542,54 @@ def __eq__(self, other) -> bool:
         return True
 
 
-def custom_gate_factory(
+def unitary_gate(
```
BoxiLi: Do you think it will help to add a verb to the name? Like `generate_unitary_gate` or `get_unitary_get` so that it distinguishes itself with the gate name itself, like `Gate` / `CustomGate `

tests/test_gates.py
```diff
@@ -256,50 +222,52 @@ def test_single_qubit_rotation(self, gate, n_angles):
...
-            ),
+            pytest.param(std.CX(), 1, id="CX"),
+            pytest.param(std.CY(), 1, id="CY"),
+            pytest.param(std.CZ(), 1, id="CZ"),
+            pytest.param(std.CS(), 1, id="CS"),
+            pytest.param(std.CT(), 1, id="CT"),
+            pytest.param(std.SWAP, 0, id="SWAP"),
+            pytest.param(std.ISWAP, 0, id="ISWAP"),
+            pytest.param(std.SQRTSWAP, 0, id="SQRTSWAP"),
+            pytest.param(std.MS([0.5 * np.pi, 0.0]), 0, id="Molmer-Sorensen"),
```
BoxiLi: This looks a bit awkward. Why does CX need to be called while not SWAP? It is not straightforward for the user to figure out when to use a called function and when to use just the class.

BoxiLi (commented):
LGTM. Some small comments.

Ideally, the inverse changes should be a separate PR. But all of them are reviewed now.

src/qutip_qip/operations/gateclass.py
Mayank447: Yes, 0b prefix in python means my int after this is in the binary representation. Other example of this is 0x prefix for hexadecimal.

src/qutip_qip/operations/gates.py
Mayank447: Yes

src/qutip_qip/operations/gateclass.py
Mayank447: Yes

tests/test_gates.py
Mayank447: I absolutely agree with you. That's why I propose this https://github.com/qutip/qutip-qip/discussions/308#discussioncomment-15950921 to make it more intuitive for the user and us as well on the design end.

src/qutip_qip/operations/gateclass.py
Mayank447: Sounds good. How about `get_custom_gate`?

Mayank447:
@BoxiLi Thanks for the review.

This PR is draft and there might be quite a bit of changes before it gets to a mergable state, so you might have to review some parts again.

src/qutip_qip/operations/gateclass.py
BoxiLi: Sounds good

src/qutip_qip/vqa.py
Mayank447: So essentially in a namespace the gate.name class attribute acts as the uid. The reason I was previously clearing the cache was because the speed of loop was faster than garbage collection in python, even for `weakref`.

I have expanded the functionality to set namespace for a gate as None, which just means its a temporary gate. This removing the need of manually doing `clear_cache` here.

Mayank447:
@BoxiLi This PR is ready for review. You only need to review it from the commit [b54d6c2](https://github.com/qutip/qutip-qip/pull/328/commits/b54d6c2f0d91a1222ab62cc100814abeeff9833a) as you already reviewed the commits before that. Rebasing it on top #325 did cause the git hashes to change wasn't aware of that earlier. That's why it shows your review comment at the top even before any commits.

Gate refactoring is complete, in the next set of PRs I will only be fixing the documentation.

src/qutip_qip/device/processor.py
```diff
@@ -380,7 +380,7 @@ def coeffs(self, coeffs):
         self.set_coeffs(coeffs)
 
     def _generate_iterator_from_dict_or_list(self, value):
-        if isinstance(value, dict):
+        if type(value) is dict:
             iterator = value.items()
```
BoxiLi, line 384: Why do we want to be so strict on dict and int using `type.. is ..`? I cannot come up with anything  that violates ` isinstance(value, dict)`

src/qutip_qip/operations/gates/single_qubit_gate.py
```diff
@@ -398,7 +397,8 @@ def _compute_qobj(arg_value: tuple[float]) -> Qobj:
         )
 
     def inverse(self) -> Gate:
-        return RX(theta=-self.theta)
+        theta = self.arg_value[0]
+        return RX(-theta)
```
BoxiLi: It is slightly counterintuitive that in RX class, nothing like arg_value is defined, but it is used in the inverse function.

What is the motivation to rename all `arg_values` argument to `phase`, `theta` etc?

src/qutip_qip/operations/controlled.py
```diff
@@ -185,22 +185,27 @@ def get_qobj(cls_or_self) -> Qobj:
...
+            )
+            inverse = controlled(
+                inverse_gate_class,
+                cls_or_self.num_ctrl_qubits,
+                cls_or_self.ctrl_value,
+            )
+            return inverse(*param)
+
+        else:
+            raise NotImplementedError
```
BoxiLi: Why this change? It seems to make the inversion more complicated. For some gate the expanded option exists for some not. Maybe add a private interface for this use case?

Also, `expanded` is a bit confusing, sounds a bit like expanding the Qobj to a big unitary.

src/qutip_qip/circuit/draw/mat_renderer.py
```diff
@@ -15,6 +15,7 @@
 from qutip_qip.circuit import QubitCircuit
 from qutip_qip.circuit.draw import BaseRenderer, StyleConfig
 from qutip_qip.operations import Gate
+from qutip_qip.operations import gates as std
```
BoxiLi, line 19: I assume `as std` can be removed now?

src/qutip_qip/circuit/instruction.py
```diff
@@ -17,7 +18,7 @@ def _validate_non_negative_int_tuple(T: any, txt: str = ""):
 
 @dataclass(frozen=True, slots=True)
 class CircuitInstruction(ABC):
-    operation: Gate | Measurement
+    operation: Gate | Type[Gate] | Measurement
```
BoxiLi, line 21: Maybe we should define a custom type for `Gate | Type[Gate]`

src/qutip_qip/namespace.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def register(self, namespace: NameSpace, name: str, item: any) -> None:
+        """Safely adds an item to the specific namespace."""
+        if namespace not in self._registry:
+            self._registry[namespace] = {}
+
+        name_upper = name.upper()
+        if name_upper in self._registry[namespace]:
+            raise ValueError(
+                f"'{name_upper}' already exists in namespace '{namespace}'"
```
BoxiLi: Maybe add some more inline docstrings here or texts in the doc later explaining the design here?

src/qutip_qip/operations/gates/other_gates.py
```diff
@@ -19,6 +20,7 @@ class GLOBALPHASE(AngleParametricGate):
     """
 
     __slots__ = "phase"
+    _namespace = NS_GATE
```
BoxiLi: What does NS stands for?

src/qutip_qip/algorithms/qpe.py
```diff
@@ -66,7 +66,7 @@ def qpe(U, num_counting_qubits, target_qubits=None, to_cnot=False):
         controlled_u = controlled(
             gate=get_unitary_gate(
                 gate_name=f"U^{power}",
-                namespace="qpe",
+                # namespace="qpe",
                 U=U_power,
             ),
         )
```
BoxiLi: What will happen if one defines several different unitary operators as gates in an algorithm? One must use different gate names to distinguish them?

src/qutip_qip/algorithms/qpe.py
```diff
@@ -55,7 +55,7 @@ def qpe(U, num_counting_qubits, target_qubits=None, to_cnot=False):
     for i in range(num_counting_qubits):
         qc.add_gate(H, targets=[i])
 
-    Gate.clear_cache("qpe")
+    # Gate.clear_cache("qpe")
```
BoxiLi: Should be removed

tests/test_vqa.py
```diff
@@ -241,3 +242,8 @@ class fakeRes:
         S = [1, 2, 3]
         result = OptimizationResult(fakeScipyRes, qutip.basis(8, 1))
         assert result._label_to_sets(S, "|010>") == "{1, 3} {2}"
+
+        print(NS_USER_GATES._registry)
+        print()
+        print(NS_GATE._registry)
+        assert False
```
BoxiLi: What are those for?

src/qutip_qip/operations/gateclass.py
```diff
@@ -92,19 +97,12 @@ def __eq__(cls, other: any) -> bool:
         if not isinstance(other, type):
             return False
 
+        # 'is' keyword in Python checks check if two variables refer to
+        # the exact same object in memory, since non-parametrized gate classes
+        # are not meant to be parametrized it works, ParametrizedGate class
+        # has its own __eq__ method which acts on an instance.
         if cls is other:
             return True
```
BoxiLi: This seems to define a default behavior that, if two gates are of the same class, they are the same gate. This might trigger unwanted behavior if someone defines a custom gate without paying attention to `__eq__`.

This can be the default for our predefined gate, but maybe not the default for all inherited.

src/qutip_qip/operations/utils.py
```diff
@@ -0,0 +1,531 @@
...
+            "Qobj with the same input and output dimensions."
+        )
+
+    # If operator dims matches the target dims
+    if dims is not None and targets is not None:
+        targ_dims = [dims[t] for t in targets]
+        if oper.dims[0] != targ_dims:
+            raise ValueError(
+                f"The operator dims {oper.dims[0]} do not match "
+                "the target dims {targ_dims}."
```
BoxiLi: Need an f for the second line

src/qutip_qip/operations/controlled.py
```diff
@@ -0,0 +1,290 @@
...
+
+    class _CustomControlledGate(ControlledGate):
+        __slots__ = ()
+        namespace = gate_namespace
+        name = gate_name
+        num_qubits = n_ctrl_qubits + gate.num_qubits
+        num_ctrl_qubits = n_ctrl_qubits
+        ctrl_value = control_value
+        target_gate = gate
+        latex_str = r"{gate_name}"
```
BoxiLi: f"{gate_name}" or r with formatting

src/qutip_qip/operations/controlled.py
```diff
@@ -0,0 +1,290 @@
...
+        if self.is_parametric() and self._target_inst != other._target_inst:
+            return False
+
+        return True
+
+    def __hash__(self) -> int:
+        return super().__hash__()
+
+
+def controlled(
```
BoxiLi: `get_controlled_gate`?

src/qutip_qip/operations/gates/single_qubit_gate.py
Mayank447: So for gates with multiple like CQASMU its hard for the user to figure out what is the order of different parameters if we are passing `arg_values`.

In standard gates like RX, `__init__` passes it to `ParametricGate`'s `__init__`, which runs the check on valid `arg_values `.

src/qutip_qip/circuit/draw/mat_renderer.py
Mayank447: matrenderer had a lot of variables similar to gates. So in order to avoid confusion, I just imported it as a alias named import std.

src/qutip_qip/circuit/instruction.py
Mayank447: Sorry I didn't get this point? X -> type i.e. Type[Gate], X() ->object i.e. Gate in this case.

src/qutip_qip/operations/gates/other_gates.py
Mayank447: NameSpace

tests/test_vqa.py
BoxiLi: Ok they are removed later, I missed.

src/qutip_qip/algorithms/qpe.py
Mayank447: Yes, but only if they are defined in the same NameSpace. If those gates are temporary, namespace can be set to None (default behaviour).

tests/test_vqa.py
Mayank447: These debug print statements were removed in a later commit.

src/qutip_qip/operations/gateclass.py
Mayank447: `is` keyword in Python checks if they are at the same memory location, this is the most strict equivalence check in Python.

They don't have to define a `__eq__` for any gateclass, `__eq__` is defined in the metaclass itself. So any gate by default inherits this `__eq__`.

src/qutip_qip/device/processor.py
Mayank447: isinstance(x, y) returns `True` is 'x' is a subclass of y. Collections has tons of these dict subclasses `Counter `, `ChainMap ` etc.

Also `type.. is ..` is much faster because it doesn't have to traverse the inheritance tree.

src/qutip_qip/circuit/instruction.py
BoxiLi: Oh, I just mean that I see quite a few `Gate | Type[Gate]` combinations in the user interface. I was wondering if we want to merge them. Give a clear instruction on what is expected. Maybe that will just cause more confusion.

BTW, if people call `qc.add_gate(X(), ...)` instead of `qc.add_gate(X,...)`, it will still work right?

src/qutip_qip/algorithms/qpe.py
BoxiLi: Just a random thought, I have seen in qiskit (?) that a small circuits (say on two qubits) can be wrapped up as a gate (block?) in a bigger circuit with multiple qubits. Seems convenient when building large circuit.

src/qutip_qip/namespace.py
BoxiLi: Not just those lines, but the general namespace and use instruction.

src/qutip_qip/device/processor.py
BoxiLi: So, for good practice in general, `instance (value, dict)` should be discouraged?

Mayank447:
> * What is the design purpose of having a compute_qobj and get_qobj. Do we expect users to call `compute_qobj` explicitly? Or only `RX(angle).get_qobj()`

We only expect the user to use `RX(angle).get_qobj()`. `compute_qobj` needs to be only defined if a custom parametrized gate is defined. The reason is `get_qobj` can't cached because one of the parameters would be `self` itself. Having a static method `compute_qobj` allows us to do so. By default we are not lru caching parametrized gates except the standard gates, but a user can do it now. This is just an advanced option on the user end if they want to make use of it.

> * In ParametricGate, `__eq__` is done by comparing the class and argument value, but **hash** is done by comparing id I think? Will this cause weird behaviors?

If two objects are equal (according to `__eq__`), they must have the same hash value (according to `__hash__`), so we should change the `__hash__` to (cls, self.arg_value) instead of id().

> I fed the gate class part to the AI to try to understand it better. It gives me the following caveats, which I cannot properly evaluate. Posting them here. What's your opinion?
> 
> * Controlled-gate class registration is non-atomic and can leak failed classes
>   Metaclass registration does two inserts: first by class name, then by (target, ctrl_qubits, ctrl_value). If the second insert fails, the first remains registered even though class creation failed, leaving stale namespace entries.

Yes, it is non-atomic since we are adding two keys for a controlled gate. Nice catch regarding leakage, I will swap my order of insertion that will prevent any leakage even in the extreme case.

> * Namespace uniqueness is not stable after mutation
>   NameSpace is a dataclass where generated **eq** includes mutable _registry, but **hash** uses only name. After registering operations, equality changes. Global duplicate checks rely on set membership (if namespace in self._registry) and can be bypassed, allowing duplicate namespaces with the same name.

I will change the `__eq__` to use name instead.

> * Cached Qobj instances are shared and mutable (medium severity, design risk)
>   @cache/@lru_cache return shared Qobj objects. If any caller mutates returned objects in place, cache contents are corrupted globally. This is fine only if immutability is a hard contract.
> About the last one. In principle, some part of Qobj can be muted. But we never encourage users to do this. I can cause many problem. The matrix is mutable for sure.

Yes this did strike my mind previously. We can have a `Note` in the documentation that if you're modifying use deepcopy. Regarding Qobj mutability to be honest in Python we can actually modify almost everything by hook or crook.

src/qutip_qip/device/processor.py
Mayank447: `instance (value, dict)` works fine for almost most of the cases. Its just that `type.. is ..` is faster and more strict.

src/qutip_qip/circuit/instruction.py
Mayank447: Doing X() would throw an error even before it goes to `add_gate`. Non-parameterized gates can't be initialised now.

Mayank447: I should have used the example of RX() instead of X() tbh

src/qutip_qip/algorithms/qpe.py
Mayank447: I agree, checkout #334

src/qutip_qip/operations/controlled.py
Mayank447: I have reverted to the previous one.

Mayank447:
@BoxiLi I have made the changes. Please re-review.

src/qutip_qip/circuit/draw/color_theme.py
```diff
@@ -90,7 +89,6 @@
     "SWAPALPHA": "#CDC1E8",  # Light Purple
     "MS": "#CDC1E8",  # Light Purple
     "RZX": "#CDC1E8",  # Light Purple
-    "CNOT": "#E0E2F7",  # Very Light Indigo
     "CX": "#E0E2F7",  # Very Light Indigo
```
BoxiLi, line 92: Not related really to this change, but just to CNOT. I understand that we know default to CX, and CNOT is an alias. If someone wants to change the latex representation to "CNOT" or to "CX". Will they be able to customize it and change this latex_str property of CX?

src/qutip_qip/circuit/circuit.py
```diff
@@ -758,7 +758,7 @@ def resolve_gates(self, basis=["CX", "RX", "RY", "RZ"]):
                     targets=targets,
                 )
 
-            elif isinstance(gate, std.RZ) and "RZ" not in basis_1q:
+            elif type(gate) is std.RZ and "RZ" not in basis_1q:
                 qc_temp.add_gate(
```
BoxiLi, line 762: I realized that to identify the gate type, one would now need different methods. For RX gate, it is `type(gate) is std.RX` but for X gate `gate is std.X`. Am I right?  Is there a way to unify them?

src/qutip_qip/circuit/draw/color_theme.py
BoxiLi: Probably not? And they need to define a custom gate?

Since `latex_str` is Final `latex_str: Final[str] = r"{\rm CNOT}"`

Mayank447: For all gates, the class attributes like `num_qubits`, `latex_str` are read only, they can't be modified after defining them in the class. One way to change the latex_str is to inherit from CX and in that inherited class define the new `latex_str`.

src/qutip_qip/circuit/circuit.py
Mayank447: At the barebone Python level, I don't think we can do it because of the difference in type, objects.

One way we can partially overcome this is by having a function `is_type_of` which takes the gate input and returns the type for a parametric gate and the gate itself for non-parametric case. And then doing the comparision using `is` or `==`.

BoxiLi: I feel like maybe then we should encourage people (at least in doc and tutorial) to use the interface `gate.name == "X"`? That should work right? I believe quite some users will be not that familar with the difference between gate class and instance. We should not make this level of understanding as a requirement of using the pacakge smoothly.

Mayank447: There is one case where `gate.name == "X" ` will fail, if the user defines his own gate named X in a different namespace, highly unlikely though. For sake of correctness, I kept like that in the codebase.

I agree with the last point completely. We can have this in the docs and the tutorials for sure.

BoxiLi: Hmm, interesting, what will the name be then? Something like `"namespace:X"`?

Mayank447: Currently `name` is just the local gate name. But we can actually have a property like `full_name` which represents `namespace:X`, this can be then be used for comparison.

Or maybe just rename `name` to `local_name` and let `name` be the fullname?

BoxiLi: I think gate.name == "X" is sufficient and straightforward. If a user defines anthor gate with the name "X", they should already expect some weird behaviors.

BoxiLi:
I shall merge if the PR is ready.

Mayank447:
You can go ahead with the merge

src/qutip_qip/compiler/gatecompiler.py
```diff
@@ -117,7 +118,7 @@ def idle_compiler(self, circuit_instruction, args):
         """
         idle_time = None
         gate = circuit_instruction.operation
-        if gate.is_parametric_gate():
+        if gate.is_parametric():
             idle_time = gate.arg_value
```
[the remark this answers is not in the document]
Mayank447: Fixed in the pulse paper PR.

PR #335, opened 2026-03-11
Pulse paper

Mayank447:
Changes:
- Move the pulse-paper codes to the root (previously was under `/doc`)
- Separate the test in the github workflows (previously was under doctest), also fix the bug which didn't test the pulse-paper code in the CI.
- Minor fixes to the pulse-paper code based on the changes in gate-refactor.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -118,7 +118,8 @@ jobs:
     - name: Test code examples for the pulse paper
       run: |
         cd pulse-paper
-        for f in *.py; do python "$f"; done
+        for f in *.py; do python - W error "$f"; done
+        # '-W error' will throw an error in case any warning is raised 
 
   finalise:
     name: Finalise coverage reporting


Mayank447:

.github/workflows/test.yml
```diff
@@ -99,11 +99,26 @@ jobs:
...
+        python -m pip install --upgrade pip
+        pip install .[graphics,control]
+
     - name: Test code examples for the pulse paper
       run: |
-        python -m pip install joblib pytest pytest-custom_exit_code
-        cd doc/pulse-paper
-        pytest *.py --suppress-no-test-exit-code
+        cd pulse-paper
+        for f in *.py; do python "$f"; done
```
BoxiLi, line 123: Could you add an argument such that the test will fail if a warning is raised?

I think it was not configured to be like that, but it is best to also detect warnings.

.gitignore
```diff
@@ -7,6 +7,10 @@ __pycache__/
 *.html
 *.htm
 
+# Pdf files (from plots in /pulse-paper)
+*.pdf
+/pulse-paper/deutsch-jozsa-qutip.qasm
```
BoxiLi, line 12: Why is this ignored?

Mayank447: The plots are generated as pdf and for one case we are also generating a .qasm file. Do we want to keep them?

Previously under /doc they were also ignored.

BoxiLi: Ah ok. They are generated. I thought they were part of the doc input. Got it.

.github/workflows/test.yml
Mayank447: Done.

Issue #334, opened 2026-03-09
Add Ops


Mayank447:
We should have a 2 level mechanism for people to write a Quantum Algorithm.

1. Gates - these are the primitives (the most basic building blocks)

2. Ops - think of them as composite gates which are reusable. For example a Quantum Adder can be constructed using only CNOTs and Tofollis, other examples would be QFT, QPE, Encoders etc. These Ops (subroutines) are meant to be reusable in other Ops (e.g. QPE would internally used in QFT, while HHL would use QPE) or these Ops can directly be used in a Quantum Circuit. These Ops should allow for num_qubits as input, since we don't want to have define QFT for 3 qubit, 4 qubit, 5 qubit, etc. separately and ideally can have their own parameter (would be useful for Oracles).

This is my preliminary sketch of how the interface would look like
```
qc = QuantumCircuit(3)
qc.add_gate(H, 0)
qc.add_gate(CX, 0, 1)
qc.add_op(QFT(3), qubits=(0, 1, 2))
```

I think we need this level of abstractness because one this is intuitive, even we think of certain subroutines like Oracle, QFT etc. to just work. And second for many useful Quantum Algorithms, they are extremely hard to construct from a gate level point of view. Like QSVT (relies on Block encoding, PCPhase), Shor's (heavily relies on modular exponentiation which is also the current bottleneck).

From the Pulse simulator point of view, since Ops are just composite gates. Since my compiler works for primitive gates, it also works for the composite gates. Same goes for circuit simulator (based on matrix multiplication), since Ops are just a fancy collection of primitive gates.

mudit06mah:
Hi! Can I work on this issue? Also I'm assuming that work on this issue will be continued from the `gate_refactor` branch only?

Mayank447:
Sure, you are free to work on this. But unlike other issues this is primarily design work - 70% of work is just design and rest 30% is implementation.

Also the Issue text is a very rough sketch, there are various other constraints/factors not considered yet. How to handle ancilla qubit in this, recycling of clean ancilla qubit, is there a mechanism to simply swap my implementation of certain Ops on the run time like QFT if there is a more version of it let say via Qubitization.

You're right about the second part. The work on this can only commence after `gate_refactor` and moving decomposition APIs to gateclasses is completed.

mudit06mah:
Here's a draft of the implementation I have in my mind:

We can have Op as an abstract class (in `operations/ops` as you mentioned):
- `num_qubits`
[...]

PR #343, opened 2026-03-17
Gatefactor: Add docstrings, fix doctests

Mayank447:
Code Changes:
- Removed compute_qobj from ParametricGate, move it to helper classes for standard Parametrized gates
-  Removed default __init__ for standard parametrized gates to be consistent with `arg_value` from user point of view. The parameter order, info for standard param gates has been moved to docstrings.
- There was some minor bugs as pointed out by copilot in #328, fixed those.

Others:
- Fixed all the failing doctest
- Added whitespace normalization by default to doctests
- Added docstring for all standard gates. Note: I used LLMs for docstring generation, although I cross checked. It could be that I might have missed something.
- Removed the deprecated code from old gates.
- Resolved qutip.qip import issue.

### DIFF
--- a/src/qutip_qip/operations/gates/other_gates.py
+++ b/src/qutip_qip/operations/gates/other_gates.py
@@ -55,10 +55,18 @@ class GLOBALPHASE(AngleParametricGate):
         return f"Gate({self.name}, phase {self.arg_value[0]}) -> Qobj:"
 
     def get_qobj(self, dtype: str = "dense") -> Qobj:
+        r"""
+        Return the scalar Qobj corresponding to the global phase factor
+        :math:`e^{i \theta}`.
+        Returns
+        -------
+        qobj : qutip.Qobj
+            A scalar quantum object with data equal to :math:`e^{i \theta}`.
+        """
         phase = self.arg_value[0]
-        return Qobj(phase, dtype=dtype)
+        return Qobj(np.exp(1.0j * phase), dtype=dtype)
 
-    def get_expanded_qobj(self, num_qubits=None, dtype: str = "dense") -> Qobj:
+    def get_expanded_qobj(self, num_qubits: int, dtype: str = "dense") -> Qobj:
         r"""
         Get the QuTiP quantum object representation expanded for a given number of qubits.
 
@@ -89,7 +97,7 @@ class GLOBALPHASE(AngleParametricGate):
 
     def inverse(self):
         theta = self.arg_value[0]
-        return GLOBALPHASE((theta,))
+        return GLOBALPHASE((-theta,))
 
 
 class TOFFOLI(ControlledGate):
--- a/src/qutip_qip/operations/gates/two_qubit_gate.py
+++ b/src/qutip_qip/operations/gates/two_qubit_gate.py
@@ -382,7 +382,7 @@ class SQRTISWAPdag(_TwoQubitGate):
 
 class BERKELEY(_TwoQubitGate):
     r"""
-    Two-qubit Berkeley (B) gate.
+    Two-qubit Berkeley gate.
 
     The Berkeley gate is a universal two-qubit entangling gate. 
     It is optimized for hardware architectures where the natural two-qubit 
@@ -440,7 +440,7 @@ class BERKELEY(_TwoQubitGate):
 
 class BERKELEYdag(_TwoQubitGate):
     r"""
-    Inverse (hermitian conjugate) of BERKLEY gate.
+    Inverse (hermitian conjugate) of BERKELEY gate.
 
     .. math::
 
--- a/pulse-paper/customize.py
+++ b/pulse-paper/customize.py
@@ -35,7 +35,7 @@ class ROT(AngleParametricGate):
         super().__init__(arg_value)
 
     def get_qobj(self, dtype) -> Qobj:
-        # This is not required, because Pules levele implementation
+        # This is not required, because Pulse level implementation
         # of this gate is already provided.
         pass
 
--- a/src/qutip_qip/circuit/circuit.py
+++ b/src/qutip_qip/circuit/circuit.py
@@ -297,10 +297,10 @@ class QubitCircuit:
 
         if control_value is not None:
             warnings.warn(
-                "'control_value' is no longer a valid argument and has been deprecated and will be removed in the future version. "
-                "from qutip_qip.operations import get_controlled_gate"
-                "from qutip_qip.operations.gates import X"
-                "Example: gate = get_controlled_gate(X, num_ctrl_qubits=1, control_value=0) instead",
+                "'control_value' is no longer a valid argument and has been deprecated and will be removed in the future version.\n"
+                "from qutip_qip.operations import get_controlled_gate\n"
+                "from qutip_qip.operations.gates import X\n"
+                "Example: gate = get_controlled_gate(X, num_ctrl_qubits=1, control_value=0)",
                 DeprecationWarning,
                 stacklevel=2,
             )
--- a/src/qutip_qip/operations/gateclass.py
+++ b/src/qutip_qip/operations/gateclass.py
@@ -59,7 +59,7 @@ class _GateMetaClass(ABCMeta):
             # step would go through, but wrt. second key won't and will throw an error.
             # This will lead to leakage in the namespace i.e. classes which don't exist but are in the namespace.
 
-            existing_gate = namespace.get(name)
+            existing_gate = namespace.get(cls.name)
             if existing_gate is not None:
                 try:
                     # Check if both classes originate from the exact same physical file.
--- a/src/qutip_qip/operations/parametric.py
+++ b/src/qutip_qip/operations/parametric.py
@@ -73,11 +73,11 @@ class ParametricGate(Gate):
                 f" but it takes {len(inspect.signature(validate_params_func).parameters)}."
             )
 
-        # get_qobj method must take only two arguments arg_value, dtype
+        # get_qobj method must take only two parameters: arg_value, dtype
         get_qobj_func = getattr(cls, "get_qobj")
         if len(inspect.signature(get_qobj_func).parameters) != 2:
             raise SyntaxError(
-                f"Class '{cls.name}' method 'get_qobj_func()' must take exactly 2 "
+                f"Class '{cls.name}' method 'get_qobj()' must take exactly 2 "
                 f"arguments (only the implicit 'self, dtype'),"
                 f" but it takes {len(inspect.signature(get_qobj_func).parameters)}."
             )


Mayank447:

src/qutip_qip/operations/gates/other_gates.py
```diff
@@ -33,19 +33,15 @@ def __init__(self, phase: float = 0.0):
...
 
-    def compute_qobj(arg_value, dtype):
-        raise NotImplementedError
-
-    def get_qobj(self, num_qubits=None, dtype: str = "dense") -> Qobj:
+    def get_qobj(self, num_qubits=None) -> Qobj:
         phase = self.arg_value[0]
         if num_qubits is None:
-            return Qobj(phase, dtype=dtype)
+            return Qobj(phase)
```
BoxiLi: I missed this before. What does this mean? It returns a Qobj of dimension [[1], [1]], which cannot be used to multiply with any matrix, I think.

BoxiLi: I remember I tried before to make it return just a number, but failed.

BoxiLi: Also, having num_qubits here as input for `get_qobj` is slightly inconsistent with other gates

doc/source/qip-basics.rst
```diff
@@ -68,18 +68,18 @@ full dimension of the circuit:
 .. testcode::
 
   U_list = qc.propagators(ignore_measurement=True)
-  print(U_list)
+  print(U_list[:-1])
```
BoxiLi, line 72: Why does the last one have to be skipped? I am lost here

doc/source/qip-basics.rst
```diff
@@ -385,14 +385,14 @@ QuTiP-QIP offers three distinct methods for visualizing quantum circuits. Below
...
-                                            ║     
-    c0 :════════════════════════════════════╩═════
+           ┌────┐  ┌───┐  ┌───────┐  ┌───┐      
+    q1 :───┤ CX ├──┤ H ├──┤       ├──┤ M ├───
+           └──┬─┘  └───┘  │       │  └─╥─┘   
+              │           │       │    ║     
+    q0 :──────█───────────┤ ISWAP ├────║─────
+                          └───────┘    ║     
+                                       ║     
+    c0 :═══════════════════════════════╩═════
```
BoxiLi, line 395: What change in the code leads to this? Looks all fine, but just out of curiosity.

src/qutip_qip/operations/gates/two_qubit_gate.py
```diff
@@ -491,7 +488,7 @@ def compute_qobj(arg_value: tuple[float], dtype: str) -> Qobj:
 
     def inverse(self) -> Gate | tuple[Type[Gate], tuple[float]]:
         alpha = self.arg_value[0]
-        return SWAPALPHA(-alpha)
+        return SWAPALPHA((-alpha))
```
BoxiLi: If you want a tuple, you need a comma.

src/qutip_qip/operations/utils.py
```diff
@@ -277,33 +277,6 @@ def _mult_sublists(tensor_list, overall_inds, U, inds):
...
-
-    Then, we can use _mult_sublists to produce a new list of unitaries by
-    multiplying TOFFOLI (and expanding) only on the qubit indices involving
-    TOFFOLI gate (and any multiplied gates).
-
-    >>> U_list, overall_inds = _mult_sublists(tensor_list, overall_inds, U, U_inds)
-    >>> np.testing.assert_allclose(U_list[0]) == Z.get_qobj())
-    >>> toffoli_xy = toffoli() * tensor(X.get_qobj(), Y.get_qobj(), identity(2))
-    >>> np.testing.assert_allclose(U_list[1]), toffoli_xy)
-    >>> overall_inds = [[2], [0, 1, 3]]
```
BoxiLi, line 306: I know they don't show in the doc because they are private functions. They are more just for our own usage. Maybe keep them?

src/qutip_qip/operations/utils.py
```diff
@@ -387,26 +360,6 @@ def _gate_sequence_product(U_list, ind_list):
...
-
-    >>> tensor_lst = [X.get_qobj(), Y.get_qobj(), Z.get_qobj(), TOFFOLI.get_qobj()]
-    >>> overall_inds = [[0], [1], [2], [0, 1, 3]]
-
-    Then, we can use gate_sequence_product to produce a single unitary
-    obtained by multiplying unitaries in the list using heuristic methods
-    to reduce the size of matrices being multiplied.
-
-    >>> U_list, overall_inds = gate_sequence_product(tensor_lst, overall_inds)
     """
```
BoxiLi, line 363: Maybe keep them?

doc/source/apidoc/qutip_qip.operations.rst
```diff
@@ -14,37 +14,52 @@ qutip\_qip.operations
...
+      gates.CTdag
+      gates.CRX
+      gates.CRY
+      gates.CRZ
+      gates.CPHASE
+      gates.CQASMU
+      gates.TOFFOLI
+      gates.FREDKIN
+      gates.GLOBALPHASE
+      gates.IDLE
```
BoxiLi, line 62: One has to include `gates` here?

src/qutip_qip/operations/gates/two_qubit_gate.py
BoxiLi: What makes you introduce arg_values back?

doc/source/apidoc/qutip_qip.circuit.rst
```diff
@@ -15,3 +15,6 @@ qutip\_qip.circuit
       QubitCircuit
       CircuitResult
       CircuitSimulator
+      CircuitInstruction
+      GateInstruction
+      MeasurementInstruction
```
BoxiLi, line 20: I am not sure if we are ready to expose them as user level API but ok.

BoxiLi:
I noticed one thing. We used to be able to use qutip_qip as a plug-in for qutip, so both

`from qutip_qip.operations import gates`

and

`from qutip.qip.operatoions import gates`

should work.

But now the second one does not work because of a namespace conflict

doc/source/qip-basics.rst
Mayank447: Last one is the global phase.

doc/source/qip-basics.rst
Mayank447: Actually previously these were CNOT, SNOT (not CX, H as see in changelog). Since the gate name has shortened that leads to this change. We weren't testing documentation for last 10 PRs so this has clearly been out of loop.

src/qutip_qip/operations/utils.py
Mayank447: They are completely out of date, as in these examples need a complete rewrite.

src/qutip_qip/operations/gates/other_gates.py
Mayank447: > Also, having num_qubits here as input for `get_qobj` is slightly inconsistent with other gates

This is not there in the latest commit.

> I missed this before. What does this mean? It returns a Qobj of dimension [[1], [1]], which cannot be used to multiply with any matrix, I think.

This is how qutip defines scalars by default. Regarding multiplication, I don't know why they don't support it. But it can be fixed actually.

src/qutip_qip/operations/utils.py
Mayank447: Same as above.

doc/source/apidoc/qutip_qip.operations.rst
Mayank447: The standard gates are exposed in `operations/gates` not `operations/`. But yeah, this should be moved to a different file` doc/source/apidoc/qutip_qip.operations.gates.rst`

Mayank447:
> I noticed one thing. We used to be able to use qutip_qip as a plug-in for qutip, so both
> 
> `from qutip_qip.operations import gates`
> 
> and
> 
> `from qutip.qip.operatoions import gates`
> 
> should work.
> 
> But now the second one does not work because of a namespace conflict

Thanks's for letting me know I will try to fix this.

src/qutip_qip/operations/gates/two_qubit_gate.py
Mayank447: Couple of reasons

- `arg_value` always existed. One less change for the user during migration.
- Consistent design, `gate=RX(theta=pi)` works but during updating it needs to be `gate.arg_value = pi/2`. This feels a little inconsistent.
- My reason for introducing `__init__` in standard parametrized gates was primarily for gates like R, MS, QASMU which take multiple parameters and the order is not defined in the `arg_values`. So instead of fixing this issue it in `__init__` it is much easier to resolve in the docstring.

BoxiLi:
Also, the old gate function is broken because we dropped the argument `N` in `expand_operator` completely. They are not tested. Just to make it work for now before we remove them completely, we should give `[2]*N`
```
In [3]: csign(N=3, control=0, target=1)
<ipython-input-3-2cb64a249209>:1: DeprecationWarning: csign has been deprecated and will be removed in future version.         Use CZ.get_qobj() instead.
  csign(N=3, control=0, target=1)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[3], line 1
----> 1 csign(N=3, control=0, target=1)

File ~src\qutip_qip\operations\old_gates.py:598, in csign(N, control, target)
    596 if N is not None:
    597     _deprecation_warnings_gate_expansion()
--> 598     return expand_operator(csign(), N, (control, target))
    599 return Qobj(
    600     [[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, -1]],
    601     dims=[[2, 2], [2, 2]],
    602 )

File ~src\qutip_qip\operations\utils.py:205, in expand_operator(oper, dims, targets, dtype)
    202 oper = oper.to(dtype)
    204 if not isinstance(dims, Iterable):
--> 205     raise ValueError(f"dims needs to be an interable {not type(dims)}.")
    207 N = len(dims)
    208 targets = _targets_to_list(targets, oper=oper, N=N)

ValueError: dims needs to be an interable False.
```


EDIT: Or maybe we can already remove those supports because they are already deprecated in the released version?

Mayank447:
> Also, the old gate function is broken because we dropped the argument `N` in `expand_operator` completely. They are not tested. Just to make it work for now before we remove them completely, we should give `[2]*N`

I removed that because argument `N` was deprecated three years back.

BoxiLi:
So we can remove all the `if N is not None:` in the old gate functions.

doc/source/apidoc/qutip_qip.operations.rst
BoxiLi: I see, makes sense. Will it cause trouble (confusion? pollute the module namespace? to expose them in `operations`?

src/qutip_qip/operations/utils.py
BoxiLi: Ok

src/qutip_qip/operations/gates/other_gates.py
BoxiLi: I think by default qutip does not expect one to give a scalar as Qobj input. Not if you want to treat it as a real scalar and multiply it with matrices/vectors.

BoxiLi:
Just some tiny clearification, in the PR description: "Removed compute_qobj from ParametricGate, move it to helper classes for standard Parametrized gates". I am slightly confused as compute_qobj is still there in `PrametricGate`

Mayank447:
> So we can remove all the `if N is not None:` in the old gate functions.

Yes, we should these were deprecated long back. I didn't remove N in the old gates, because I was like no one will really use them any longer. But since we removed deprecation from `expand_operator`, makes sense to do the same with old gate functions as well.

Mayank447:
> I am slightly confused as compute_qobj is still there in `PrametricGate`

It isn't there, I just verified. You can check the first commit of the PR [Move compute_qobj to standardised gates](https://github.com/qutip/qutip-qip/pull/343/commits/6fbcb2f3a668a290d6c20af079cd35c5ef3a9b48)

doc/source/apidoc/qutip_qip.circuit.rst
Mayank447: 99% of the users won't need it. But for the 1% who want to add custom stuff, there's no harm in exposing these.

doc/source/apidoc/qutip_qip.operations.rst
Mayank447: I think wouldn't expose them. Pollution is the main reason, whenever we add Ops (#334) sometime in future, the reusable components would need to be exposed in `/operations`.

Doing it in a different module, gives a nice organisation
gates -> `operations/gates/`
Ops -> `operations/ops/`
Initialization -> `operations/intialize/`

BoxiLi:
> It isn't there, I just verified. You can check the first commit of the PR [Move compute_qobj to standardised gates](https://github.com/qutip/qutip-qip/pull/343/commits/6fbcb2f3a668a290d6c20af079cd35c5ef3a9b48)

My bad, I was checking in the existing gate_refactor branch, not this PR.

Mayank447:
@BoxiLi This PR is ready for a re-review

- I have resolved the `qutip.qip` import issue via inspect. The more robust solution to it would be to make the import aliasing proper in the qutip repo.
- I have removed the deprecated code from old gates.

BoxiLi:
What would be the more robust solution in qutip? Maybe it is something that should be applied to all familary package.

Mayank447:
So the core problem with sys.modules["qutip.qip"] = sys.modules["qutip_qip"] hack used in qutip is that it only aliases the top-level module. Not any of the submodules. That's why `qutip.qip.operations` threw an error.

We can write a custom `MetaPathFinder` using the default `importlib` library to alias all the submodules too.

BoxiLi:
There is also a suppressed comment from Copilot in its reply regarding BERKELEYdag

BoxiLi:
Let me know when you are done. Glad to see the tests passing.

Mayank447:
You can go ahead with the merge

BoxiLi:
Just found a possible bug

```python
gates.MS(0.,0.)
```
raise an error. Some gates are not tested?

```
ValueError: Requires 2 parameters, got 1
```

BoxiLi:
Oh, I see, we are back to a single tuple as input. Instead of two values. So it should be `MS((0., 0.))`

PR #344, opened 2026-03-17
fix: showarg style option now displays gate argument value in matplotlib renderer

### LINKED ISSUE #340: Bug in circuit plotting: `showarg` arguments is not working

frankgoldfish:
## Summary

Fixes #340

The `showarg` style option was being stored but never used to actually change the rendered gate label. The gate text was unconditionally set to `gate.arg_label or gate.name` *before* `showarg` was even read, so setting `style={"showarg": True}` had no visible effect.

## Root Cause

In `mat_renderer.py`, the order of operations was:
1. Set `self.text = gate.arg_label if gate.arg_label is not None else gate.name`
2. Read `self.showarg = style.get("showarg", False)`

`showarg` was read after `self.text` was already fixed, and was never used to modify `self.text`.

## Fix

Read `showarg` first, then conditionally build `self.text`:
- `showarg=True` + `arg_value` present → displays `"<name>(<arg_value>)"` (e.g. `"RZ(1.5707...)"`)
- `showarg=False` or no `arg_value` → original behaviour preserved (`arg_label` or gate `name`)

## Reproduction

```python
from math import pi

q = QubitCircuit(1)
q.add_gate("RZ", targets=0, arg_values=pi / 2, style={"showarg": True})
q.draw()  # Before fix: shows only "RZ". After fix: shows "RZ(1.5707...)".
```

## Test

Added `test_showarg_displays_gate_argument` to `tests/test_renderer.py` — verifies the argument value appears in the label and that the full draw pipeline runs without error.

### DIFF
--- a/src/qutip_qip/circuit/draw/mat_renderer.py
+++ b/src/qutip_qip/circuit/draw/mat_renderer.py
@@ -786,9 +786,21 @@ class MatRenderer(BaseRenderer):
 
             if isinstance(gate, Gate):
                 style = gate.style if gate.style is not None else {}
-                self.text = (
-                    gate.arg_label if gate.arg_label is not None else gate.name
-                )
+                self.showarg = style.get("showarg", False)
+                if self.showarg and gate.arg_value is not None:
+                    arg_str = str(gate.arg_value)
+                    base = (
+                        gate.arg_label
+                        if gate.arg_label is not None
+                        else gate.name
+                    )
+                    self.text = f"{base}({arg_str})"
+                else:
+                    self.text = (
+                        gate.arg_label
+                        if gate.arg_label is not None
+                        else gate.name
+                    )
                 self.color = style.get(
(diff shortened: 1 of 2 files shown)


BoxiLi:
Thank you for the fix.

There is a large refactor in the development branch, and we might have to rebase it after that one is merged. Sorry about this.

One more request. Could you add a truncation to the string? Right now it is very long. Truncate after two digit would be nice.
image

frankgoldfish:
Closing this PR — found that [PR #342](https://github.com/qutip/qutip-qip/pull/342) already addresses this issue. Apologies for the duplicate.

Issue #288, opened 2025-12-01
Undefined variable in QASM


Mayank447:
File `src/qutip-qip/qasm.py`

def _regs_processor(self, regs, reg_type):
        """
        Process register tokens: map them to the :class:`.QubitCircuit` indices
        of the respective registers.

        Parameters
        ----------
        regs : list of str
            Token list corresponding to qubit/cbit register invocations.
        reg_type : str
            reg_type can be "measure" or "gate" to specify type of required
            processing.

        Returns
        -------

        regs list : list of (list of regs)
                list of register sets to which circuit operations
                are applied.

        """

        # turns messy tokens into appropriate form
        # ['q', 'p', '[', '0', ']'] -> ['q', 'p[0]']
        regs = [reg.replace(" ", "") for reg in regs]
        if "[" in regs:
            prev_token = ""
            new_regs = []
            open_bracket_mode = False
            for token in regs:
                if token == "[":
                    open_bracket_mode = True
                elif open_bracket_mode:
                    if token == "]":
                        open_bracket_mode = False
                        reg_name = new_regs.pop()
                        new_regs.append(reg_name + "[" + reg_num + "]")
                    elif token.isdigit():
                        reg_num = token
                    else:
                        raise SyntaxError("QASM: incorrect bracket formatting")
                else:
                    new_regs.append(token)
            if open_bracket_mode:
                raise SyntaxError("QASM: incorrect bracket formatting")
            regs = new_regs

The variable `reg_num` is undefined

satzgits:
Hi @Mayank447, I noticed #288 is currently self-assigned.

I’ve prepared a small fix on top of gate_refactor for the `reg_num` undefined-variable path in QASM parsing:
- malformed indexed registers like `measure q[] -> c[0];` now raise a proper `SyntaxError` instead of an internal `UnboundLocalError`
- added a regression test for this case

If this issue is still open for contribution, I can open the PR right away.

PR #355, opened 2026-03-24
Fix malformed QASM register index handling

satzgits:
## Summary
Fix malformed indexed register handling in QASM parsing to avoid internal `UnboundLocalError` and return a user-facing syntax error.

## Changes
- In `_regs_processor`, validate bracket content before using `reg_num`.
- Raise `SyntaxError("QASM: incorrect bracket formatting")` for malformed indexed registers like `q[]`.
- Add regression test for malformed measurement register index:
  - `measure q[] -> c[0];`

## Validation
- Ran focused QASM tests related to parser errors and the new regression case.
- All targeted tests passed locally.

## Notes
- Small scoped fix, no API changes.
- Based on `gate_refactor`.

addresses #288

### DIFF
--- a/src/qutip_qip/qasm.py
+++ b/src/qutip_qip/qasm.py
@@ -428,11 +428,17 @@ class QasmProcessor:
             prev_token = ""
             new_regs = []
             open_bracket_mode = False
+            reg_num = None
             for token in regs:
                 if token == "[":
                     open_bracket_mode = True
+                    reg_num = None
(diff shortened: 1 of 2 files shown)


Mayank447:
Hi, thanks for the PR! There are a lot of hidden bugs in `qasm.py` and the parsing logic does need some refinement. Nice to see, someone working on it.

Can you add the test case in the PR description which would earlier incorrectly fail but passes now.

PR #356, opened 2026-03-25
Add group dependencies, test for Python 3.13, 3.14

Mayank447:
Currently `pip install qutip-qip[test]` installs `pytest`, `pytest-repeat` which are dev dependencies, unlike optional dependencies like visualization, control etc which add a new feature set. Also dev dependencies are not meant to be exposed to the user. Earlier pyproject.toml didn't have the separating concept of optional, dev dependencies but now they do.

- [PEP 735](https://peps.python.org/pep-0735/) was accepted around 1.5 years back which separated dev (group) dependencies from optional dependencies. Now a contributor can simply do `pip install --group test` to install all the testing dependencies for the project. Followed the same for test, lint dependency in qutip-qip.
- Deleted `doc/requirements.txt` and moved those dependencies to a doc group in `pyproject.toml`. `pip install --group doc` will install all the dependencies needed to build documentation.
- Updated the GitHub workflow based on the above change.
- Updated the github action version from v4/v5 to v6.
- Added testing for Python version 3.13, 3.14 (earlier we were only testing upto Python 3.12)


pyproject.toml
```diff
@@ -64,21 +64,33 @@ Issues = "https://github.com/qutip/qutip-qip/issues"
...
 qutip_qip = "qutip_qip.family"
 
+[dependency-groups]
+lint = [
+    "black==26.1.0",
+]
+test = [
+    "pytest>=5.2",
+    "pytest-repeat>=0.9",
+    "qiskit_aer>=0.17.0",  # Some tests in test_qiskit.py use Aer for verification
```
BoxiLi, line 82: qiskit_aer is a very heavy package. I am wondering if we really want to include it here. Instead of just in the qiskit group

Mayank447: We can't include it as an optional-dependency because it isn't really used anywhere in the codebase except testing.

My assumption was every contributor does `pip install -e .[full]` while installing the package for development. This also includes qiskit. So doing `pip install --group test` won't really add much in dependency.

We should ideally add these two commands in the development doc.

.github/workflows/test.yml
```diff
@@ -17,30 +17,41 @@ jobs:
...
-            qiskit-version: '==2.2.*'
             python-version: '3.10'
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
     - name: Set up Python
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
```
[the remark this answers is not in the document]
Mayank447: Copilot is incorrect here. Zero-indented sequences are actually valid YAML

PR #360, opened 2026-03-26
Switch PyPI release workflow to trusted publishing (PEP 740)

DeconBear:
This PR updates the release workflow to use PyPI trusted publishing (OIDC) instead of long-lived API tokens.

  Changes:
  - Add `id-token: write` permission to the deploy job.
  - Replace `twine upload` with `pypa/gh-action-pypi-publish@release/v1`.
  - Collect wheel/sdist artifacts into `dist/` and publish from there.
  - Keep existing manual confirmation and non-dev-version checks.

  Refs: qutip/qutip#2868

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,10 +121,21 @@ jobs:
     if: ${{ github.event.inputs.confirm_ref != '' }}
     needs: [deploy_test, build_sdist, build_wheels]
     runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      id-token: write
 
     steps:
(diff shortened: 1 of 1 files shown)


BoxiLi:
Sorry, this needs to be discussed among the admin maintainers first, and I don't think this script will work by itself because something must be done with PyPI by the project admins.

Mayank447:
> I don't think this script will work by itself because something must be done with PyPI by the project admins.

This is correct, for reference: https://docs.github.com/en/actions/how-tos/secure-your-work/security-harden-deployments/oidc-in-pypi

PR #364, opened 2026-03-30
Minor maintenance work

Mayank447:
**Description**
Minor maintenance work:

- Moved .dev tag from pyproject.toml to VERSION file. This is because read the docs shows 0.5 instead of 0.5.dev which might be confusing for new contributors.
- Change log for 0.4.1 was not present in the master branch so was not present in `docs/src/changelog.rst` and doesn't reflect in[ latest version](https://qutip-qip.readthedocs.io/en/latest/changelog.html) in RTD. It was committed to 0.4.X only (probably accidently). cherry-picked the commit for the same.
- Updated the version numbers in pyproject.toml for dependencies. Also updated few of the classifier tags.
- Added a small dependabot for updating github actions version, black formatter version.
- Replaced codeclimate.yaml to qlty.toml

**Related issues or PRs**
None

**AI Usage Disclosure**
None


pyproject.toml
```diff
@@ -1,8 +1,7 @@
 [build-system]
 requires = [
-    "setuptools>=42",
+    "setuptools>=77.0.3",
```
BoxiLi, line 3: Where does this minimum version come from?

Mayank447: I remember mentioning the reason briefly in the commit description. Here is the full expanded version.

Essentially wheels needed to be dropped from build dependencies. Wheels are not required while making editable install now. Wheels used to be a dependency for setuptools particularly used in bdist. That logic was moved out of wheels and into setuptools in v70 (released around 2 years back). So in the edge case if a user has a recent version of wheels but an older version of setuptools (which still depends on wheels for that logic), the editable install/build would fail.

Now setuptools 78 removed the non-PEP 517 config format which was deprecated since 2021 but there are still some old package which haven't been updated. Luckily all our dependencies do follow PEP 517. But for sake of keeping some buffer I stuck with setuptools 77 as our baseline. Btw the main qutip repo also follows the same versioning for setuptools I believe for the same reason.

Mayank447:
@BoxiLi This PR is complete. I have also replaced `codeclimate.yml` with `qlty.toml`

BoxiLi:
Thanks!

PR #372, opened 2026-04-02
Improve the Github action worflow

Mayank447:
**Changes:**

- Currently we are running the entire test suite on everything. A simple change in README.md runs all the pytest on different Python version and different OS. This is redundant and shouldn't trigger in the first place. *Added paths to our workflow* to run selectively only on dependent file changes.
- If I push a typo fix, and then 30 seconds later push another commit to the same Pull Request, currently GitHub will run both test suites concurrently. *Set `cancel-in-progress: true`* in our workflow to automatically cancel the old, outdated run except on the master branch.
- Set read-only permission for Github actions, since no writing to the repo is involved for these actions. This is recommended by github and good from security pov.
- Run test on push to release branches `qutip-qip-*`
- Dropped `build_documentation.yaml` as it was replaced by Read the Doc Pull request build workflow.
- Added CI for Spelling check.
- Separated doctest, pulse-paper jobs from `test.yaml` into individual `.yaml` files
- Renamed some of the Github workflow files to better reflect their working.
- Renamed `.yml` extensions to `.yaml`
- Added sponsor button.

**Related issues or PRs**
None

**AI Usage Disclosure**
Gemini 3.1 Pro: Reviewing and suggesting changes


.github/workflows/test.yaml
```diff
@@ -13,56 +23,59 @@ jobs:
...
-            qutip-version: ''
-            python-version: '3.10'
-
     steps:
-    - uses: actions/checkout@v6
+    - name: Clone the repo
+      uses: actions/checkout@v4
+
     - name: Set up Python
       uses: actions/setup-python@v6
```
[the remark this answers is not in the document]
Mayank447: This is wrong

Mayank447:
@BoxiLi Do you want coverage reporting and uploading to coveralls for each (python, os version) or for anyone of it?

BoxiLi:
I think before we merged the coverage report for all the different python and os versions?

Mayank447:
> I think before we merged the coverage report for all the different python and os versions?

Yes, I have made the change to do the same. But its showing the status for matrix (python, os version) coverage reports as well. Could you go to the dashboard on [coveralls.io](https://coveralls.io/) and uncheck Leave Statuses for Parallel Jobs box? This will ensure we only see the final merged report.

BoxiLi:
Got it, I check the 
```
Limit status updates to overall build?
Don't send a status update for each individual job or subproject.
(Only applies for parallel builds.)
```
box

BoxiLi (approved):

BoxiLi:
Maybe we should keep the name `build_documentation.yml`? Changing the name creates a new panel in GitHub Actions.
image

Mayank447:
> Maybe we should keep the name `build_documentation.yml`? Changing the name creates a new panel in GitHub Actions.

I changed back to the old name. You will need to delete all the 11 workflow runs for [documentation.yml](https://github.com/qutip/qutip-qip/actions/workflows/documentation.yml), so that it doesn't showup in the GitHub actions panel anymore

Mayank447:
@BoxiLi Similar to the main qutip repo could you enable [Pull requests RTD build](https://docs.readthedocs.com/platform/stable/guides/pull-requests.html). This will allow us to delete `build_documentation.yaml`, Read the Doc will itself handle the build and documentation view for the PR e.g. [https://app.readthedocs.org/projects/qutip/builds/32249652/](https://app.readthedocs.org/projects/qutip/builds/32249652/)

Additional configuration can always be specified in the existing `.readthedocs.yaml` file in the root.

BoxiLi:
Done, maybe open a new PR or a new commit to see if we can see that?

> Pull requests opened before enabling pull request builds will not trigger new builds automatically. Push a new commit to the pull request to trigger its first build.

Mayank447:
> Done, maybe open a new PR or a new commit to see if we can see that?

RTD build test didn't trigger.

BoxiLi:
Try again? There were some github permission issue.

Mayank447:
> Try again? There were some github permission issue.

It works now.

Mayank447:
@BoxiLi I have added a Spell check CI, this should help prevent spelling mistakes/typos in future.

The current list of typos https://github.com/qutip/qutip-qip/actions/runs/24454026149/job/71449857919?pr=372 I will correct in another PR.

Mayank447:
> Got it, I check the
> 
> ```
> Limit status updates to overall build?
> Don't send a status update for each individual job or subproject.
> (Only applies for parallel builds.)
> ```
> 
> box

Could you enable this, I still see coverage for each python/os version instead of a single marked report and coverage number.

BoxiLi:
Sorry I thought I checked it, maybe I forgot to click save. Now it should work.

Mayank447:
> Sorry I thought I checked it, maybe I forgot to click save. Now it should work.

Ahh, no problem

.github/workflows/pulse-paper.yaml
```diff
@@ -0,0 +1,44 @@
...
+          run: |
+            python -m pip install --upgrade pip
+            python -m pip install scipy==1.14.0  # Due to Issue #39 in qutip-qtrl
+            pip install .[graphics,control]
+
+        - name: Test code examples for the pulse paper
+          working-directory: pulse-paper
+          run: |
+            for f in *.py; do python -W error "$f"; done
+            # '-W error' will throw an error in case any warning is raised
```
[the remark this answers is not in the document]
Mayank447: Although not required is preferred as per yaml indentation rules.

Mayank447:

BoxiLi:
One question here about the spelling check. How is that correctness defined? Sometimes in physics, we do use some words that are deliberately "wrong". Will that break the test, or is there a configuration file that one has to edit?

Edit: I see, so `.typos.toml` is the config file.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,109 @@
+name: Tests
+permissions:
+  contents: read
+
+on:
+  push:
+    branches: [ master ]
+    paths: &trigger_paths
+      - "src/**/*.py"
```
BoxiLi: Should we really limit to py files? Right now it seems fine. I was wondering if we may introduce Cython or other things at some point.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,109 @@
...
+      - "tests/**"
+      - ".github/workflows/test.yaml" #Run when this exact pipeline file changes
+      - "pyproject.toml" # Run when any dependency changes
```
BoxiLi: Maybe also `pytest.ini`

Mayank447: Good catch!

BoxiLi:
I know that it was like this before, but do you think we should include other branches for testing on push?
Right now, it includes only `branches: [ master ]`. But we also use `qutip-qip-...` as release branches.

BoxiLi:
Also, there are two coveralls reports. I think it has been like this for a while.

.github/workflows/test.yaml
Mayank447: There are two ways to handle this

- We run it on any file in `src/**`. This will take care of all cases. [I didn't do this because some people add readme, other files in src to explain the module internals/working. But this is the most robust way.]
- If we introduce Cython, we can add another line for `.pyx` file and `setup.py` to this workflow. [But introducing Cython will require quite a lot of change in workflows especially build/packaging]

Mayank447:
> Also, there are two coveralls reports. I think it has been like this for a while.

Yeah, this is not due to the CI workflow. I tried to removing the `finalise-coverage-reporting` job and it didn't show any coverage number. Also both coveralls reports have the same build number, so its more of a notification issue. Online this is what I found, disable `USE STATUS API` on coveralls.io since coveralls is also added as a GitHub app, using both is causeing two coverage reports.

BoxiLi (commented):
Since we replaced build_documentation.yml with readthedocs, we should enable `fail_on_warning: true` to `.readthedocs.yaml` to test for warnings.

https://docs.readthedocs.com/platform/stable/config-file/v2.html

Mayank447:
> I know that it was like this before, but do you think we should include other branches for testing on push? Right now, it includes only `branches: [ master ]`. But we also use `qutip-qip-...` as release branches.

Nice suggestion, we should be doing this.

Edit: The same change should be made in other qutip packages as well.

.github/workflows/test.yaml
BoxiLi: Yes, I just wanted to make sure that we don't forget it next time when we add something. Like in the tests, we have a few qasm files. I cannot come up with any specific example right now. Maybe running it on any file to just be secure?

Mayank447: > Maybe running it on any file to just be secure?

Sounds good.

Mayank447:
@BoxiLi I have made all the changes. Also could you try disabling `Use Status API` on coveralls.io, it might resolve the double coverage report issue

BoxiLi:
Ok, I unchecked. This is the closest I found.
> **Send status updates**
Send commit statuses (aka. status updates) to your SCM.

BoxiLi (approved):

BoxiLi:
So, I am supposed to merge this first, and then you update the spell check one?

Mayank447:
Yes

Issue #331, opened 2026-03-04
Add general Measurement


Mayank447:
Currently qutip-qip only supports measurement in Pauli Z i.e. the standard basis. Expanding this to a general operator basis would be helpful as several algorithms/protocols measure in other basis e.g. Pauli X while other like OTOCs (Out of Time Ordered Correlators) inherently rely randomised measurements.

Ideally along side general measurement we should add easier methods to `QubitCircuit` e.g. `qc.measure_all()` and also expand (or refactor) the `Measurement` class as target, cbits are being moved to `MeasurementInstruction` in gate refactor. There are couple of more things I have in mind which can be added alongside controlled measurements, operations on cbits and POVM measurements.

mudit06mah:
Hi @Mayank447 
Are you working on this or is it okay for me to? Since this is now blocking the Ops feature issue i can work on this first.

Mayank447:
Sure go ahead, I haven't started work on this. Also before starting with code implementation, open a discussion tab for this and propose your implementation design there.

mudit06mah:
Further Discussion in #373

mudit06mah:
Checklist:
- [x] Shift `classical_store` to Qubit Circuit
- [x] Remove `targets` and `classical_store` from `measurement.py`
- [ ] Expand Measurement Class: `num_qubits`, `get_qobj()`, `apply()`, `get_measurement_ops()`
- [ ] Implement operator subclasses (Mx, My, Mz)
- [ ] Update simulator to use arbitrary bases and map multi-qubit results
- [ ] Add `measure_all()` method to `circuit.py`
- [ ] Add basis conversion (Mx, My -> Mz) to `to_qasm()` in `instructions.py`

Mayank447:
Adding this as a future note:

**Naimark's (Neumark's) Dilation Theorem**
The physical realization of a POVM is explained by Naimark's Dilation Theorem. The theorem states that any generalized POVM on a principal system can be physically realized by: 

1. Attaching an ancilla system.
2. Applying a unitary transformation to the joint system (entangling the principal system and ancilla).  
3. Performing a standard orthogonal projective measurement (PVM) on the larger joint system.

This is needed for simulating POVMs on the Pulse Simulator, though adding projective measurement to Pulse Simulator is pending and will need quite a bit of work. But once Projective Measurements are implemented for Pulse Simulator, this implies POVMs are automatically as well.

PR #386, opened 2026-04-17
Measurement: Remove targets and classical_store

mudit06mah:
**Description**
This PR removes `targets` and `classical_store` from `Measurement` class.

**Changes**
- Deprecates(Removes) `targets` and `classical_store` from `Measurement` class
- Changes `measurement_comp_basis()` to accept `targets` as arg
- Changes `_apply_measurement` to `accept targets` and `classical_store` as arg
- Modifies related test cases 

**Related issues or PRs**
Part of #331

**AI Usage Disclosure**
To fix docstring grammar


tests/test_qasm.py
```diff
@@ -98,11 +98,13 @@ def test_qasm_teleportation():
     filename = "teleportation.qasm"
     filepath = Path(__file__).parent / "qasm_files" / filename
     teleportation = read_qasm(filepath)
-    final_measurement = Measurement("start", targets=[2])
-    initial_measurement = Measurement("start", targets=[0])
+    final_measurement = Measurement("start")
+    initial_measurement = Measurement("start")
```
mudit06mah, line 102: I was curious as to why both of these measurements have the name "start", it is the same in other tests as well so It is intentional I think.

Mayank447: The `name` argument in Measurement object doesn't matter and should be removed. 

Ideally both of these should be

```suggestion
    final_measurement = Mz
    initial_measurement = Mz
```

If someone wants to define a custom measurement operator, it should be via inheritance

Mayank447:
Hi, thanks for this PR. I admit I am a bit lost. All the changes seem to be squashed in a single commit, could you explain a bit more the reasoning behind the different changes.

mudit06mah:
Hey @Mayank447, since the changes weren't that big I only made one commit.

This PR focuses on removing `targets` and `classical_store` from the Measurement class.
Since `measurement_comp_basis` was accessing Measurement.targets I had to pass it from `_apply_measurement`which uses the 'classical_store` so I passed both targets and classical_store from the step function.
I changed the test cases accordingly.

tests/test_qasm.py
mudit06mah: I haven't yet introduced measurement operators, but we can remove the name prop from the class I guess

BoxiLi:
Glad to see this coming, just wanna add that, please let me know if you find it better to create a separate dev branch for this. E.g. it is hard or cumbersome to keep the tests passing for this refactor in every single PR.

mudit06mah:
@BoxiLi I did fix all the tests for this PR, didn't appear to be too cumbersome but maybe when I introduced measurement operators which might cause a lot of changes in tests, though the PRs could very well just be bigger in size. I guess you can make that call :)

tests/test_qasm.py
Mayank447: Sounds good.

mudit06mah: @Mayank447  I have left the name prop but removed assigning the passed the name prop in the `__init__` method, so when operators are introduced they can have the name prop to "MZ', "MX", etc.

Mayank447: Instead of defining the measurement operators "Mz", "Mx" as name prop, we can make `Measurement` class an abstract class and define any measurement operators by inheriting the `Measurement`. This is what we follow for `Gate` class and in future plan to do the same for device classes. You don't need to make any change for this in this PR.

I see `Measurement()` being repeated at quite a few places. What you can do is to define `Mz = Measurement()` in `measurement.py`, import that wherever `qc.add_measurement` is being called use `Mz` instead of `Measurement()`. This will fix the interface for the other files, we won't have to touch them. Later all our future changes like making `Mz` inherit from `Measurement` etc. would only need to be made in `measurement.py`

mudit06mah: > we can make `Measurement` class an abstract class and define any measurement operators

Yes that is what I meant as well, I am not removing the name prop right now and leaving it to use for the `Mz` and other operator classes which will be abstracted by `Measurement`.

Nice idea about defining `Mz`, I have made new changes according to that :)

Mayank447:
Doctest appears to be failing. @mudit06mah could you make the same change in `qip-basics.rst`

mudit06mah:
@Mayank447 looks like I forgot to add the `qip-basics.rst` file in my commit, should be fine now :)

src/qutip_qip/circuit/circuit.py
```diff
@@ -259,10 +259,9 @@ def add_measurement(
 
         Parameters
         ----------
-        measurement: string
-            Measurement name. If name is an instance of `Measurement`,
-            parameters are unpacked and added.
-        targets: list
+        measurement : :class:`.Measurement`
+            `Measurement` object
```
Mayank447:
```suggestion
            `Measurement` subclasses
```

src/qutip_qip/circuit/circuit.py
```diff
@@ -277,25 +276,27 @@ def add_measurement(
...
-
-        meas = Measurement(name, targets=targets, classical_store=classical_store)
+        if isinstance(measurement, str):
+            warnings.warn(
+                "Passing a string name to add_measurement is deprecated. "
+                "Please pass an instantiated Measurement object instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            measurement = Measurement(name=measurement)
```
Mayank447:
```suggestion
        if isinstance(measurement, str):
            warnings.warn(
                "Passing a string name to add_measurement has been removed. "
                "Defaulting to measurement in the Z- basis.",
                DeprecationWarning,
                stacklevel=2,
            )
            measurement = Mz
```

Object instantiation won't work since `Measurement` class will be abstracted in next couple of PRs.

src/qutip_qip/circuit/circuit.py
```diff
@@ -277,25 +276,27 @@ def add_measurement(
...
 
         if type(targets) is int:
             targets = [targets]
 
         if type(classical_store) is int:
             classical_store = [classical_store]
 
+        if targets is None or classical_store is None:
+            raise ValueError("'targets' and 'classical_store' must not be None")
+
```
Mayank447, line 291:
```suggestion
        if not isinstance(targets, IntSequence):
            raise ValueError("'targets' must be a sequence of positive integer")
            
         if not isinstance(classical_store, IntSequence):
            raise ValueError("'classical_store' must be a sequence of positive integer")
```

Also need to check for each `targets`, `classical_store` is >=0 and less than `num_qubits` for the circuit. I wrote some helper functions for this, they were used in `add_gate`. You can reuse those functions.

src/qutip_qip/operations/measurement.py
```diff
@@ -1,55 +1,44 @@
...
-
 import numpy as np
-
+import warnings
 import qutip
 from qutip import basis
 from qutip.measurement import measurement_statistics
 from qutip_qip.operations import expand_operator
 
+__all__ = ["Measurement", "Mz"]
```
Mayank447:
```suggestion
__all__ = ["Mz"]
```

`Measurement` is already exported in `__init__.py`. Although not relevant to this PR, later `measurement.py` can be renamed to `measurementclass.py` and a directory `measurement` can be created which stores all the standard measurements like Pauli basis, POVMs

src/qutip_qip/operations/measurement.py
```diff
@@ -1,55 +1,44 @@
...
-        for ind_list in [self.targets]:
-            if isinstance(ind_list, Iterable):
-                all_integer = all(
-                    [isinstance(ind, numbers.Integral) for ind in ind_list]
-                )
-                if not all_integer:
-                    raise ValueError("Index of a qubit must be an integer")
-
-    def measurement_comp_basis(self, state):
+    def measurement_comp_basis(self, state, targets):
```
Mayank447: This should be made a `classmethod`

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -308,16 +312,28 @@ def _evolve_state_einsum(self, gate, targets_indices, state):
...
 
         Parameters
         ----------
         operation: :class:`.Measurement`
             Measurement gate in a circuit object.
+        targets : tuple of int
+            The indices of the qubits to be measured.
+        classical_store : tuple of int
+            The indices of the classical registers where the measurement
+            results will be stored.
```
Mayank447:
```suggestion
        qubits : tuple of int
            The indices of the qubits to be measured.
        cbits : tuple of int
            The indices of the classical registers where the measurement
            results will be stored.
```

Same change needs to be made in the function.

tests/test_circuit.py
```diff
@@ -530,15 +531,19 @@ def test_run_teleportation(self):
         teleportation = _teleportation_circuit()
 
         state = tensor(rand_ket(2), basis(2, 0), basis(2, 0))
-        initial_measurement = Measurement("start", targets=[0])
-        _, initial_probabilities = initial_measurement.measurement_comp_basis(state)
+        initial_measurement = Mz
+        _, initial_probabilities = initial_measurement.measurement_comp_basis(
```
Mayank447:
```suggestion
        _, initial_probabilities = Mz.measurement_comp_basis(
```

tests/test_circuit.py
```diff
@@ -530,15 +531,19 @@ def test_run_teleportation(self):
...
+        )
 
         teleportation_sim = CircuitSimulator(teleportation)
         teleportation_sim_results = teleportation_sim.run(state)
         state_final = teleportation_sim_results.get_final_states(0)
 
-        final_measurement = Measurement("start", targets=[2])
-        _, final_probabilities = final_measurement.measurement_comp_basis(state_final)
+        final_measurement = Mz
+        _, final_probabilities = final_measurement.measurement_comp_basis(
```
Mayank447:
```suggestion
        _, final_probabilities = Mz.measurement_comp_basis(
```

Same change needs to be made at few more places in this file as well.

Mayank447 (changes requested):
Good work!

src/qutip_qip/operations/__init__.py
```diff
@@ -57,6 +57,7 @@
     "get_controlled_gate",
     "AngleParametricGate",
     "Measurement",
+    "Mz",
```
Mayank447:
```suggestion
```

Why are you exporting `Mz` in both measurement and operation's `__init__.py`.

mudit06mah: Oh yeah, sorry I guess I got a bit confused since I was thinking on where to keep and where to remove it, I have removed it now from `operations.__init__`

src/qutip_qip/operations/measurement.py
```diff
@@ -1,55 +1,45 @@
...
-            if isinstance(ind_list, Iterable):
-                all_integer = all(
-                    [isinstance(ind, numbers.Integral) for ind in ind_list]
-                )
-                if not all_integer:
-                    raise ValueError("Index of a qubit must be an integer")
-
-    def measurement_comp_basis(self, state):
+    @classmethod
+    def measurement_comp_basis(cls, state, targets):
```
Mayank447:
```suggestion
    def measurement_comp_basis(cls, state, qubits):
```

Same change in function definition.

src/qutip_qip/circuit/circuit.py
```diff
@@ -249,7 +250,7 @@ def add_state(
 
     def add_measurement(
         self,
-        measurement: str | Measurement,
+        measurement: Measurement,
```
Mayank447:
```suggestion
        measurement: Type[Measurement],
```

`Measurement` -> instantiated object from Measurement class, `Type[Measurement]` -> a Measurement subclass

Mayank447 (approved):
LGTM, just two minor changes.

mudit06mah:
@Mayank447 I have made the necessary changes now, could you please take a look?

Mayank447:
LGTM

src/qutip_qip/operations/measurement.py
```diff
@@ -83,8 +75,7 @@ def measurement_comp_basis(self, state):
             raise ValueError("target is not valid")
 
         measurement_ops = [
-            expand_operator(op, dims=[2] * n, targets=self.targets)
-            for op in measurement_ops
+            expand_operator(op, dims=[2] * n, targets=qubits) for op in measurement_ops
         ]
```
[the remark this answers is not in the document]
Mayank447: Ignore this, expanding Measurement beyond one qubit is unrrelated to this PR.

doc/source/qip-basics.rst
```diff
@@ -307,22 +308,23 @@ QuTiP-QIP offers three distinct methods for visualizing quantum circuits. Below
 
   from qutip_qip.circuit import QubitCircuit
   from qutip_qip.operations.gates import H, CX, ISWAP
+  from qutip_qip.operations.measurement import Mz
```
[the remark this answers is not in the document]
Mayank447: I missed this earlier, check line 300. There might be more instances.

src/qutip_qip/circuit/circuit.py
```diff
@@ -249,7 +250,7 @@ def add_state(
 
     def add_measurement(
         self,
-        measurement: str | Measurement,
+        measurement: Type[Measurement],
         targets: int | IntSequence,
         classical_store: int,
```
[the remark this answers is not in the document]
Mayank447: Ignore this

src/qutip_qip/circuit/circuit.py
```diff
@@ -249,7 +250,7 @@ def add_state(
...
         index: None = None,
```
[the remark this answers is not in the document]
Mayank447: I agree with Copilot on this

```suggestion
        measurement: Type[Measurement],
        targets: int | IntSequence,
        classical_store: int | IntSequence,
        index: None = None,
```

src/qutip_qip/circuit/circuit.py
```diff
@@ -277,25 +277,29 @@ def add_measurement(
...
+        check_limit("targets", targets, 0, self.num_qubits - 1)
+        check_limit("classical_store", classical_store, 0, self.num_cbits - 1)
 
         self._instructions.append(
             MeasurementInstruction(
-                operation=meas,
+                operation=measurement,
                 qubits=tuple(targets),
                 cbits=tuple(classical_store),
             )
```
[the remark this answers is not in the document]
Mayank447: Ignore

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -308,16 +312,28 @@ def _evolve_state_einsum(self, gate, targets_indices, state):
...
+        state : qutip.Qobj
+            The current state of the quantum circuit (ket or density matrix).
+
+        Returns
+        -------
+        state : qutip.Qobj
+            The collapsed state after the measurement.
         """
-        states, probabilities = operation.measurement_comp_basis(self.state)
+        states, probabilities = operation.measurement_comp_basis(self.state, qubits)
```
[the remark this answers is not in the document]
mudit06mah: I have removed the state argument from the `_apply_measurement.py` function.

Mayank447 (approved):

mudit06mah:
@BoxiLi Could you take a look as well?

BoxiLi (approved):

BoxiLi:
LGTM. Sorry for being the blocker here. A bit constrained by time at the moment.

BoxiLi:
Please also check quickly if any qip tutorial notebooks are broken because of this in the qutip-tutorial repo (feel free to use AI to do this if you want). We should update the notebook if there are any error messages. The notebook should work for both the released qutip-qip 0.4.1 and for this branch (maybe with some deprecation warnings)

Mayank447:
There were only deprecations introduced into this PR no removals.

While reviewing I missed one placed where the previous syntax was used https://github.com/qutip/qutip-qip/pull/386#discussion_r3178702413 and the tests passed. So the tutorials should work just fine I guess.

BoxiLi:
One tutorial is failing because we now explicitly ask to provide `qubits` for `measurement_comp_basis` . This argument was not there in the old versions. Can we make it optional? E.g if qubits is not provided, measure all qubits? @mudit06mah @Mayank447 

Also maybe we should rename it as `targets` since it can be general qudits, not just bit.

```python
    def measurement_comp_basis(cls, state, qubits):
        """
        Measures a particular qubit (determined by the target)
        whose ket vector/ density matrix is specified in the
        computational basis and returns collapsed_states and probabilities
        (retains full dimension).

        Parameters
        ----------
        state : ket or oper
                state to be measured on specified by
                ket vector or density matrix
        qubits : list or tuple of int
                The indices of the qubits to be measured.

```

```
_ /home/runner/work/qutip-tutorials/qutip-tutorials/notebooks/quantum-circuits/teleportation.ipynb _
---------------------------------------------------------------------------
initial_measurement = Measurement("start", targets=[0])
initial_measurement.measurement_comp_basis(state)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[9], line 2
      1 initial_measurement = Measurement("start", targets=[0])
----> 2 initial_measurement.measurement_comp_basis(state)

TypeError: Measurement.measurement_comp_basis() missing 1 required positional argument: 'qubits'
```

Mayank447:
This is a little tricky I will need to think about it.

mudit06mah:
@BoxiLi can we just ignore it for now? this method will be removed soon!

BoxiLi:
Sure, it is not that urgent since it is the master version, not the release, but it does cause one of the tutorial tests to fail. Even if we remove it, we need to keep the backward compatibility for some time for deprecation.

PR #389, opened 2026-04-21
Drop Python 3.10

Mayank447:
**Description**
Drop Python 3.10 in conjunction with [qutip#2855](https://github.com/qutip/qutip/pull/2875) and Python 3.10 reaching its official end-of-life later this fall.

Python 3.11 should allow us to use exception groups for better handling of errors (which is currently quite lacking) and `Self`, Variadic Generics and `LiteralStrings` for type hinting (#262).

Other changes:
- Update dependency and tooling versions in `pyproject.toml`
- Move `pytest.ini` to `pytest.toml`
- Update `MANIFEST.in` (`README.md`, `LICENSE`, `pyproject.toml` are included by default in sdist no need to mention them but `VERSION` is not which was previously missing).

**Related issues or PRs**
None

**AI Usage Disclosure**
None


BoxiLi (approved):

Mayank447:
@BoxiLi This PR is complete, it is ready for a re-review.

.pytest.toml
```diff
@@ -0,0 +1,22 @@
+pythonpath = ["src"]
```
BoxiLi, line 1: You don't need a header for this?

Mayank447: This is needed to allow `pytest` to run without an editable install. Also makes `sys.path` discovery a little faster.

BoxiLi: Oh I meant something like a `[pytest]` at the top. I guess maybe not, because all the tests passed.

BoxiLi (approved):

.pytest.toml
Mayank447: That's needed in case this config was in `pyproject.toml`

PR #392, opened 2026-04-25
Add CI action to test hyperlinks

Mayank447:
**Description**
I was updating the documentation and it appears quite a lot of links were broken. This check can be automated, adding a CI action for the same.

**Related issues or PRs**
None

**AI Tools Usage Disclosure**
None

### DIFF
--- a/.github/workflows/spell-check.yaml
+++ b/.github/workflows/spell-check.yaml
@@ -23,3 +23,22 @@ jobs:
 
       - name: Spell Check the repo
         uses: crate-ci/typos@v1.45.1
+
+  hyperlink:
+    name: Check all hyperlinks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone the repo 
+        uses: actions/checkout@v6
+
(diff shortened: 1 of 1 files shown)


BoxiLi:
I remember we had the same problem with broken links in qutip tutorials. Maybe there is something one can learn from there. The broken links that fail in the test are not all broken. For example, the physical review papers are just behind Cloudflare walls.

Mayank447:
> I remember we had the same problem with broken links in qutip tutorials. Maybe there is something one can learn from there.

qutip-tutorial [link fix PR](https://github.com/qutip/qutip-tutorials/pull/108) uses a GitHub action as well (though it's only restricted to markdown in their case).

> The broken links that fail in the test are not all broken. For example, the physical review papers are just behind Cloudflare walls.

Similar to qutip-tutorial, we can ignore the 403 Error (resource exists but lack of permission). This should not falsely flag links like the Cloudflare links which return 403 for bots.

PR #395, opened 2026-04-30
Add lockfile pylock.toml

Mayank447:
**Description**
[PEP 751](https://peps.python.org/pep-0751/) was accepted a year back. It introduced a python wide lockfile format `pylock.toml` (never thought I would see this day personally). But jokes aside, we require a lockfile for several reasons:

- We don't put an upper cap on dependencies in `pyproject.toml` as best practice e.g. `scipy>=1.9`. In future, if someone tries to install a previous version of qutip-qip, scipy might have deprecated or removed a function we used in qutip-qip. But dependency resolver like pip, uv would install the most recent version of scipy and won't actually complaint since they resolve based on version requirement specified in `pyproject.toml` and as per `pyproject.toml` we allow it. But when the person runs the test those fail, so in the end for that person it becomes a manual exercise trying different version permutations of dependencies until he succeeds. Leaving the lockfile allow anyone in the future to install a working version of qutip-qip from the release branch with a single command `pip install -r pylock.toml`

- `requirements.txt` although used by some with similar intentions does not capture transitive dependencies (i.e. dependencies of our dependency) and are insecure (no checksum, index and can be tempered). Also all major package manager pip, uv, poetry, pdm etc.  now allow for `pylock.toml` as an import/export format, while some like pdm have replaced their own lockfile with `pylock.toml` and hatch, poetry in work of doing the same. So we should be using the more modern `pylock.toml` instead of `requirements.txt`.

- We don't change dependency versions in bug fix (micro) release only between major and minor (feature) releases. So we just need to run one more command `pip lock .` before creating a new release branch just to ensure the dependency versions in lockfile are synced. uv already automates this sync, pip in future plans to do the same so this step although a little manual right now won't be required in future. This will make release branches more useful for installing previous qutip-qip versions, in addition to being a code reference checkpoint it is currently.

For reference, this is a really well written [blog](https://reinforcedknowledge.com/python-project-management-and-packaging-pep-751-update-and-some-of-the-remaining-issues-of-packaging/) on `pylock.toml` which explains lockfile, `pylock.toml` standards, its benefit and some limitations in detail.

**Related issues or PRs**
None

**AI Tools Usage Disclosure**
None


hodgestar:
As a well-behaved library, we support a wide range of the packages we depend on and their transitive dependencies. It is not our place to dictate the precise versions of dependencies used by our users or to ensure the overall security of their installations.

For example, we don't tell the Debian QuTiP package maintainer which versions of numpy and scipy to depend on. Those are decided by what is packaged with Debian.

Likewise we don't tell our users which versions of BLAS or MKL to use.

While I sympathize with the desire to have a "blessed working secure set of packages to install", I'm not sure that lock files provide much help when viewed from the perspective of the maintainers of a library like QuTiP. If a user installed qutip-qip and qutip-ctrl, which lockfile should they use? Or what if a user wants to simultaneously use another library like pandas? Should they use qutip's lockfile? Or pandas?

Lock files *are* useful for deployment, but we don't control deployment for our users.

For us lock files can be useful for describing *a possible set* of packages to install, but not *the set* to install.

The blog post you linked to implicitly acknowledges the above via the examples it uses, but I thought it worthwhile to make the situation for library maintainers like you and me explicit.

To be clear, I'm not against us having a lock file in our repositories, it's just important that their role and use is clearly understood.

Mayank447:
> To be clear, I'm not against us having a lock file in our repositories, it's just important that their role and use is clearly understood.

You are absolutely right about the last point. I forgot mentioning about it in the PR description - lockfiles are meant for reproducible source installs, they are not to be included in the wheels, sdist.

The supported version dependencies are still specified in `pyproject.toml`, `pip install qutip-qip` or `pip install .` (for source install) will use `pyproject.toml` to resolve dependency and install them. The lockfile install is only triggered in case of `pip install -r pylock.toml`only in case of source install from github, akin to `-r` flag previously used by the `requirements.txt` file. Lockfiles should be thought of as version 2 of `requirements.txt`.

 The use case scenario I had in mind were easier installs for new contributors, reproducible build for previous versions when downloaded from release branch (newer package version or master branch probably wouldn't need lockfile install). Its similar to how people leave `uv.lock` or `requirements.txt` in the repo.

hodgestar:
@BoxiLi I'd like to leave reviewing and deciding to you / the wider admin team. I just wanted to write a perspective here.

PR #400, opened 2026-05-21
fix: added threshold and top_m filtering for histogram generation

### LINKED ISSUE #388: Plot non-zero probability states in histogram
Currently in probability histogram for Circuit Simulation we are plotting all the states even the ones which have zero probability. This causes the no. of elements on x-axis to scale with O($2^n$), which is hard to interpret for $n \geq 5$. We should restrict plotting only non-zero probability states.

Additionally, there can be two options (as arguments):

1) Only plot top $m$ states, where m is an integer $\leq 2^n$, with default being $m=2^n$

2) Only plot states above a certain probability threshold with default being 0.

Partially based out of recommendation by @Copilot in https://github.com/qutip/qutip-qip/pull/352#discussion_r2975798077_

anshumansahu-stack:
**Description**
This PR adds a threshold filtering feature to the histogram generation logic in circuit\simulator\result.py . The user needs to pass a floating point argument called threshold in the plot_histogram() function, the default value of which is 0.0. The plot_dict automatically filters all the measurements strictly above the threshold and renders the values on the histogram.

Since the values having zero measurement are automatically filtered by the default value of threshold, no stray values are seen with empty measurements in the histogram.

Here is the before and after state of the histogram of a circuit having just 5 qubits and an X gate applied to qubit[0].

before
after

This PR also adds a top_m argument to the plot_histogram() function, the default value of which is none. Passing the top_m=m as an argument filters only the greatest m measurements, which in turn are plotted on the histogram.

I did not get an appropriate name for the above argument, any suggestions for the same are welcome.

If any of the above renders the histogram to be empty, a warning is also issued regarding the same and the blank histogram is plotted.

Another testcase for rendering the histogram of a 3-qubit circuit having hadamard gates on qubits [0] and [1] are as follows:
The screenshots attached are respectively for the following cases:
--> Older version(unchanged)
--> This version with no arguments
--> this version having threshold=0.3
--> warning message of the above with the same threshold when no histograms are rendered
--> histogram with top_m=2.

hadamard_unchanged
hadamard_zero_arguments
empty_histogram_threshold_0 3
warning_on_threshold_0 3
top_2_measurements

**Related issues or PRs**
fixes #388 

**AI Tools Usage Disclosure**
Google AI Studio: for referencing python documentation and syntax.

### DIFF
--- a/src/qutip_qip/circuit/simulator/result.py
+++ b/src/qutip_qip/circuit/simulator/result.py
@@ -150,10 +150,7 @@ class CircuitResult:
         for cbits, prob in zip(self.cbits, self.probabilities):
             if prob > threshold:
                 binary = "".join(str(b) for b in cbits)
-                if binary not in plot_dict:
-                    plot_dict[binary] = prob
-                else:
-                    plot_dict[binary] += prob
+                plot_dict[binary] = prob
 
         plot_dict = dict(
             sorted(plot_dict.items(), key=lambda item: item[1], reverse=True)
--- a/src/qutip_qip/circuit/simulator/result.py
+++ b/src/qutip_qip/circuit/simulator/result.py
@@ -103,7 +103,7 @@ class CircuitResult:
         fig: "matplotlib.figure.Figure | None" = None,
         ax: "matplotlib.axes.Axes | None" = None,
         color: str = "#1f77b4",
-        top_m: int = None,
+        top_m: int | None = None,
         threshold: float = 0.0,
     ) -> "tuple[matplotlib.figure.Figure, matplotlib.axes.Axes]":
         """
--- a/src/qutip_qip/circuit/simulator/result.py
+++ b/src/qutip_qip/circuit/simulator/result.py
@@ -124,7 +124,7 @@ class CircuitResult:
             The number of top m highest probability measurements to be plotted in the histogram. If not specified, defaults to all the measurements above the mentioned tolerance.
 
         threshold : float, optional
-            the percentage above which the measurements are desired to be seen in the histogram. excludive, default being 0.
+            The probability e.g. 1e-7 above which the measurements are plotted in the histogram. If not specified, defaults to 1e-10.
 
         Returns
         -------
--- a/src/qutip_qip/circuit/simulator/result.py
+++ b/src/qutip_qip/circuit/simulator/result.py
@@ -104,7 +104,7 @@ class CircuitResult:
         ax: "matplotlib.axes.Axes | None" = None,
         color: str = "#1f77b4",
         top_m: int | None = None,
-        threshold: float = 0.0,
+        threshold: float = 1e-10,
     ) -> "tuple[matplotlib.figure.Figure, matplotlib.axes.Axes]":
         """
         Plot a histogram of the measurement outcomes and their probabilities.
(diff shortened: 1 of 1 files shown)


src/qutip_qip/circuit/simulator/result.py
```diff
@@ -102,6 +103,8 @@ def plot_histogram(
         fig: "matplotlib.figure.Figure | None" = None,
         ax: "matplotlib.axes.Axes | None" = None,
         color: str = "#1f77b4",
+        top_m: int = None,
```
Mayank447:
```suggestion
        top_m: int | None = None,
```

src/qutip_qip/circuit/simulator/result.py
```diff
@@ -117,6 +120,12 @@ def plot_histogram(
         color : str, optional
             Bar color for the histogram. Default is '#1f77b4'.
 
+        top_m : int, optional
+            the number of top greatest measurements to be plotted in the histogram. In case ignored, all the measurements having measurements more than the mentioned tolerance will be plotted.
```
Mayank447:
```suggestion
            The number of top m highest probability measurements to be plotted in the histogram. If not specified, defaults to all the measurements above the mentioned tolerance.
```

src/qutip_qip/circuit/simulator/result.py
```diff
@@ -117,6 +120,12 @@ def plot_histogram(
...
+
+        threshold : float, optional
+            the percentage above which the measurements are desired to be seen in the histogram. excludive, default being 0.
```
Mayank447:
```suggestion
            The probability e.g. 1e-7 above which the measurements are plotted in the histogram. If not specified, defaults to 1e-10.
```

src/qutip_qip/circuit/simulator/result.py
```diff
@@ -102,6 +103,8 @@ def plot_histogram(
         fig: "matplotlib.figure.Figure | None" = None,
         ax: "matplotlib.axes.Axes | None" = None,
         color: str = "#1f77b4",
+        top_m: int = None,
+        threshold: float = 0.0,
```
Mayank447:
```suggestion
        threshold: float = 1e-10,
```

There are always floating point error, need to account for that.

src/qutip_qip/circuit/simulator/result.py
```diff
@@ -135,11 +144,29 @@ def plot_histogram(
...
 
         for cbits, prob in zip(self.cbits, self.probabilities):
-            binary = "".join(str(b) for b in cbits)
-            plot_dict[binary] += prob
+            if prob > threshold:
+                binary = "".join(str(b) for b in cbits)
+                if binary not in plot_dict:
+                    plot_dict[binary] = prob
+                else:
+                    plot_dict[binary] += prob
```
Mayank447: In which case will the else condition be triggered?

Mayank447 (changes requested):
LGTM, few minor comments.

src/qutip_qip/circuit/simulator/result.py
anshumansahu-stack: This was to avoid keyerrors in the blank dictionary, the else condition just had the old logic. ill update the code accordingly.

Mayank447 (approved):
@BoxiLi When merging do a squash merge

BoxiLi (approved):

BoxiLi:
Thank you @anshumansahu-stack and @Mayank447

PR #402, opened 2026-05-30
Measurement: Add measurement subclasses

mudit06mah:
**Description**
This PR adds measurement subclasses
Changes:
- Adds Measurement subclasses (Mx, My, Mz)
- Adds `get_measurement_ops()`
- Deprecates `measurement_comp_basis`
- Modifies `_apply_measurement` to call `measurement_statistics()`
- Adds helper to test files and changes test cases   

**Related issues or PRs**
Part of #331 

**AI Tools Usage Disclosure**
- Gemini 3.1 Pro: Implementation discussion
- Gemini 3.5 Flash: Docstring and changing tests

### DIFF
--- a/doc/source/qip-basics.rst
+++ b/doc/source/qip-basics.rst
@@ -53,7 +53,7 @@ A circuit with the various gates and registers available is demonstrated below:
   :options: +NORMALIZE_WHITESPACE
 
     [GateInstruction(operation=Gate(SWAP, num_qubits=2), qubits=(0, 1), cbits=(), style=None, cbits_ctrl_value=None),
-      MeasurementInstruction(operation= Measurement(M), qubits=(1,), cbits=(0,), style={}),
+      MeasurementInstruction(operation=<class 'qutip_qip.operations.measurement.Mz'>, qubits=(1,), cbits=(0,), style={}),
       GateInstruction(operation=Gate(CX, num_qubits=2), qubits=(0, 1), cbits=(), style=None, cbits_ctrl_value=None),
       GateInstruction(operation=Gate(X, num_qubits=1), qubits=(0,), cbits=(0,), style=None, cbits_ctrl_value=1),
       GateInstruction(operation=Gate(SWAP, num_qubits=2), qubits=(0, 1), cbits=(), style=None, cbits_ctrl_value=None)]
--- a/doc/source/qip-basics.rst
+++ b/doc/source/qip-basics.rst
@@ -53,7 +53,7 @@ A circuit with the various gates and registers available is demonstrated below:
   :options: +NORMALIZE_WHITESPACE
 
     [GateInstruction(operation=Gate(SWAP, num_qubits=2), qubits=(0, 1), cbits=(), style=None, cbits_ctrl_value=None),
-      MeasurementInstruction(operation=<class 'qutip_qip.operations.measurement.Mz'>, qubits=(1,), cbits=(0,), style={}),
+      MeasurementInstruction(operation=Measurement(Mz), qubits=(1,), cbits=(0,), style={}),
       GateInstruction(operation=Gate(CX, num_qubits=2), qubits=(0, 1), cbits=(), style=None, cbits_ctrl_value=None),
       GateInstruction(operation=Gate(X, num_qubits=1), qubits=(0,), cbits=(0,), style=None, cbits_ctrl_value=1),
       GateInstruction(operation=Gate(SWAP, num_qubits=2), qubits=(0, 1), cbits=(), style=None, cbits_ctrl_value=None)]
--- a/src/qutip_qip/circuit/instruction.py
+++ b/src/qutip_qip/circuit/instruction.py
@@ -19,7 +19,7 @@ def _validate_non_negative_int_tuple(T: any, txt: str = ""):
 
 @dataclass(frozen=True, slots=True)
 class CircuitInstruction(ABC):
-    operation: Gate | Type[Gate] | Measurement | Type[Measurement]
+    operation: Gate | Type[Gate] | Type[Measurement]
     qubits: tuple[int, ...] = tuple()
     cbits: tuple[int, ...] = tuple()
     style: dict = field(default_factory=dict)
--- a/src/qutip_qip/circuit/instruction.py
+++ b/src/qutip_qip/circuit/instruction.py
@@ -19,7 +19,7 @@ def _validate_non_negative_int_tuple(T: any, txt: str = ""):
 
 @dataclass(frozen=True, slots=True)
 class CircuitInstruction(ABC):
-    operation: Gate | Type[Gate] | Measurement
+    operation: Gate | Type[Gate] | Measurement | Type[Measurement]
     qubits: tuple[int, ...] = tuple()
     cbits: tuple[int, ...] = tuple()
     style: dict = field(default_factory=dict)
@@ -168,7 +168,13 @@ class MeasurementInstruction(CircuitInstruction):
 
     def __post_init__(self) -> None:
         super(MeasurementInstruction, self).__post_init__()
-        if not isinstance(self.operation, Measurement):
+        if not (
+            isinstance(self.operation, Measurement)
+            or (
+                isinstance(self.operation, type)
+                and issubclass(self.operation, Measurement)
+            )
+        ):
             raise TypeError(f"Operation must be a measurement, got {self.operation}")
 
         if len(self.qubits) != len(self.cbits):
--- a/src/qutip_qip/circuit/instruction.py
+++ b/src/qutip_qip/circuit/instruction.py
@@ -3,6 +3,7 @@ from typing import Type
 from dataclasses import dataclass, field
 import warnings
 from qutip_qip.operations import Gate, Measurement
+from qutip_qip.operations.measurement import Mz
 
 
 def _validate_non_negative_int_tuple(T: any, txt: str = ""):
@@ -185,6 +186,10 @@ class MeasurementInstruction(CircuitInstruction):
         return True
 
     def to_qasm(self, qasm_out) -> None:
+        if self.operation is not Mz:
+            raise NotImplementedError(
+                f"Exporting non-Z basis measurement ({self.operation}) to QASM is not supported."
+            )
         for qubit, cbit in zip(self.qubits, self.cbits):
             qasm_out.output(f"measure q[{qubit}] -> c[{cbit}];")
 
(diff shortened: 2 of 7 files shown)


src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -312,14 +314,19 @@ def _evolve_state_einsum(self, gate, targets_indices, state):
         )
         return state
 
-    def _apply_measurement(self, operation, qubits, cbits):
+    def _apply_measurement(
+        self,
+        operation: Measurement | Type[Measurement],
```
Mayank447: A general comment for the changes, we should restrict the measurement to a subclass only. In case of gates, there were parametrized gates for which we had to instantiate the object. But that's not a limitation for measurement.

```suggestion
        operation: Type[Measurement],
```

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -331,15 +338,25 @@ def _apply_measurement(self, operation, qubits, cbits):
         state : qutip.Qobj
             The collapsed state after the measurement.
         """
-        states, probabilities = operation.measurement_comp_basis(self.state, qubits)
+        current_state = self.state
+        n = int(np.log2(current_state.shape[0]))
```
Mayank447: Might be easier, declaring a class variable (internal) for `num_qubits`

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -331,15 +338,25 @@ def _apply_measurement(self, operation, qubits, cbits):
...
+        if isinstance(operation, type) and issubclass(operation, Measurement):
```
Mayank447, line 343: subclass already covers type.

src/qutip_qip/circuit/instruction.py
```diff
@@ -19,7 +19,7 @@ def _validate_non_negative_int_tuple(T: any, txt: str = ""):
 
 @dataclass(frozen=True, slots=True)
 class CircuitInstruction(ABC):
-    operation: Gate | Type[Gate] | Measurement
+    operation: Gate | Type[Gate] | Measurement | Type[Measurement]
```
Mayank447:
```suggestion
    operation: Gate | Type[Gate] | Type[Measurement]
```

src/qutip_qip/operations/measurement.py
```diff
@@ -1,46 +1,45 @@
...
-            raise AttributeError("argument index is no longer supported")
-
-        if targets is not None or classical_store is not None:
-            warnings.warn(
-                "'targets' and 'classical_store' arguments in Measurement are deprecated. "
-                "Please pass these directly to QubitCircuit.add_measurement() method.",
-                DeprecationWarning,
-                stacklevel=2,
-            )
+    def get_measurement_ops(self) -> list[Qobj]:
```
Mayank447, line 40: This should be an abstract method and `Measurement` an abstract class.

src/qutip_qip/operations/measurement.py
```diff
@@ -95,4 +100,31 @@ def __repr__(self):
...
+        op0 = basis(2, 0) * basis(2, 0).dag()
+        op1 = basis(2, 1) * basis(2, 1).dag()
+        return [op0, op1]
+
+
+class Mx(Measurement):
+    name = "Mx"
+    num_qubits = 1
+
+    def get_measurement_ops(self) -> list[Qobj]:
```
Mayank447, line 125: minus sign missing wrt. |-><-|

src/qutip_qip/operations/measurement.py
```diff
@@ -95,4 +100,31 @@ def __repr__(self):
...
+        plus = (basis(2, 0) + basis(2, 1)).unit()
+        minus = (basis(2, 0) - basis(2, 1)).unit()
+        return [plus * plus.dag(), minus * minus.dag()]
+
+
+class My(Measurement):
+    name = "My"
+    num_qubits = 1
+
+    def get_measurement_ops(self) -> list[Qobj]:
```
Mayank447, line 135: Same as above

src/qutip_qip/operations/measurement.py
```diff
@@ -95,4 +100,31 @@ def __repr__(self):
         return str(self)
 
 
-Mz = Measurement()
+class Mz(Measurement):
+    name = "Mz"
+    num_qubits = 1
+
+    def get_measurement_ops(self) -> list[Qobj]:
```
Mayank447, line 115: There is a minus sign missing wrt. |1><1|

Mayank447 (changes requested):

src/qutip_qip/operations/measurement.py
mudit06mah: I didn't make `Measurement` an abstract class yet for backward compatibility, what do you think we should do?

Mayank447: I don't think anyone instantiated measurement class (mostly used `add_meadurement` method instead) since `Measurement` class didn't provide much functionality earlier and pulse level simulation for measurement is not present. So we can break the backward compatibility and make it an abstract class

src/qutip_qip/operations/measurement.py
mudit06mah: @Mayank447 could you elaborate on this?

Mayank447: When you measure in the x-basis, the state $|+> = \frac{|0>+|1>}{2}$ gives eigenvalue 1, similarly $|-> = \frac{|0>-|1>}{2}$ gives -1.

mudit06mah: Oh I see what you mean, actually since we pass a list of Qobjs to `measurement_statistics` they're passed on to `measurement_statistics_povm` which expect POVMs, so there is no need for a minus sign here

doc/source/qip-basics.rst
```diff
@@ -53,7 +53,7 @@ A circuit with the various gates and registers available is demonstrated below:
   :options: +NORMALIZE_WHITESPACE
 
     [GateInstruction(operation=Gate(SWAP, num_qubits=2), qubits=(0, 1), cbits=(), style=None, cbits_ctrl_value=None),
-      MeasurementInstruction(operation= Measurement(M), qubits=(1,), cbits=(0,), style={}),
+      MeasurementInstruction(operation=<class 'qutip_qip.operations.measurement.Mz'>, qubits=(1,), cbits=(0,), style={}),
```
BoxiLi: Maybe we should define a string representation to make it shorter and more illustrative?

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -331,15 +338,25 @@ def _apply_measurement(self, operation, qubits, cbits):
...
         """
-        states, probabilities = operation.measurement_comp_basis(self.state, qubits)
+        current_state = self.state
+        n = self.qc.num_qubits
+        if issubclass(operation, Measurement):
+            operation = operation()
+        raw_ops = operation.get_measurement_ops()
+        measurement_ops = [
+            expand_operator(oper=op, dims=[2] * n, targets=qubits) for op in raw_ops
+        ]
```
BoxiLi, line 348: Just a random idea, not something to fix in this PR:

Just like we use einsum to save memory in gate simulation, we can probably also perform the measurement without explicitly constructing the full tensor operator.

src/qutip_qip/operations/measurement.py
```diff
@@ -64,7 +62,13 @@ def measurement_comp_basis(cls, state, qubits):
                         the probability of measuring a state in a the state
                         specified by the index.
         """
-
+        warnings.warn(
+            "'measurement_comp_basis' has been deprecated and will be removed "
+            "in future versions. Please use 'get_measurement_ops()' combined "
+            "with simulator logic, or the upcoming 'apply()' method.",
```
BoxiLi: It is a bit weird to refer to a "future" method

src/qutip_qip/operations/measurement.py
```diff
@@ -1,46 +1,44 @@
...
-        Create a measurement with specified parameters.
-        """
-        if name is not None:
+    def __init__(self, *args, **kwargs) -> None:
+        if type(self) is Measurement:
             warnings.warn(
-                "'name' argument in Measurement has been deprecated.",
+                "Direct instantiation of Measurement() is deprecated and will "
+                "be removed in future versions. Please use a specific subclass "
+                "like 'Mz()' instead.",
```
BoxiLi: Should one initialize `Mz()` or just `Mz`

src/qutip_qip/operations/measurement.py
BoxiLi: This apply function sounds very useful because I see a lot `get_measurement_ops`->`expand_operator`->`measurement_statistics` in the tests.

It will fix a more common interface and hide details that can potentially be further improved in the future.

BoxiLi (commented):
Could you reply to the above comments and click resolve if you think it is resolved? Otherwise it is bit hard to follow what is addressed.

We might also have to edit the qasm support, at least reject Mx and My for now.

src/qutip_qip/operations/measurement.py
mudit06mah: Changed this to `Mz`, good catch!

src/qutip_qip/operations/measurement.py
```diff
@@ -88,11 +101,42 @@ def measurement_comp_basis(cls, state, qubits):
...
-            return f" Measurement({self.name})"
+            return f"Measurement({self.name})"
         return "Measurement"
 
-    def __repr__(self):
+    def __repr__(self=None):
+        if self is None:
+            return "Measurement"
+        if isinstance(self, type):
+            return f"Measurement({getattr(self, 'name', self.__name__)})"
```
Mayank447:
```suggestion
    def __repr__(self):
        return f"Measurement({getattr(self, 'name', self.__name__)})"
```

In which case would `self` be None?

Mayank447 (commented):
Just one minor catch

src/qutip_qip/operations/measurement.py
mudit06mah: I added this because of a warning the doc test was throwing:
`TypeError: Measurement.__repr__() missing 1 required positional argument: 'self'`

But looking into it again it turned out to be because the method `_MeasurementMetaClass.__str__` had `return cls.__repr__()` which was invoking `__repr()__` (self was not being passed).
I changed it to `return repr(cls)` which now works as intended.

BoxiLi:
I think the only thing remains now is to reject Mx and My in qasm.

BoxiLi (approved):

PR #405, opened 2026-06-06
 feature: Added Oracle Gate implementation

anshumansahu-stack:
**Description**
This Pull Request adds a new "OracleGate" Feature to the qutip-qip Library. This gate takes a function logic_func as an input, and applies the transformation:

|x>|y> --> |x>|y ⊕ logic_func(x)>

hence, turning an arbitrary python function into a unitary Oracle Matrix.

The input register |x> is always preserved in the method |x,y⊕f(x)>. The |x> acts as a memory that makes the entire matrix unitary regardless of whether the function is reversible. Even if f(x) is a many-to-one(irreversible) function, the matrix just remains a permutation of the basis states(discussion #323). The XOR is bitwise for the multi-target case.

I have implemented the OracleGate as a function rather than a separate class, the reason being the num_qubits parameter needed strictly to be an int for this gate. Since the size of the matrix is variable, It failed the class-level validation and had to be made as a function that returns the Gate class type upon execution.

This feature also utilises the inspect module to take the python function and extract its source code as a string. Since standard python functions are hard to save or print, by storing the source code in _source, the gate is serialisable.

This PR also includes a separate test_oracle.py file in tests that covers its identity, CNOT, multi-qubit tests, Unitarity test, a separate named function test, serialisation and a Grover's Oracle test.

Any suggestions on the changing of naming conventions and other edits are welcome.

**AI Tools Usage Disclosure**
Google AI Studio and Claude: For referring Python documentation and generating Testcases.

### DIFF
--- a/src/qutip_qip/operations/__init__.py
+++ b/src/qutip_qip/operations/__init__.py
@@ -13,6 +13,7 @@ from .gateclass import Gate, get_unitary_gate
 from .parametric import ParametricGate, AngleParametricGate
 from .controlled import ControlledGate, get_controlled_gate
 from .measurement import Measurement
+from .gates import get_oracle_gate
 from .old_gates import (
     rx,
     ry,
@@ -55,6 +56,7 @@ __all__ = [
     "ControlledGate",
     "get_unitary_gate",
     "get_controlled_gate",
+    "get_oracle_gate",
     "AngleParametricGate",
     "Measurement",
     "rx",
--- a/src/qutip_qip/operations/gates/__init__.py
+++ b/src/qutip_qip/operations/gates/__init__.py
@@ -53,6 +53,7 @@ from .other_gates import (
     GLOBALPHASE,
     TOFFOLI,
     FREDKIN,
+    OracleGate,
 )
 
 GATE_CLASS_MAP = {
@@ -105,6 +106,7 @@ GATE_CLASS_MAP = {
     "CPHASE": CPHASE,
     "RZX": RZX,
     "CQASMU": CQASMU,
+    "ORACLE": OracleGate,
 }
 
 CONTROLLED_GATE_MAP = {
--- a/src/qutip_qip/operations/gates/__init__.py
+++ b/src/qutip_qip/operations/gates/__init__.py
@@ -53,7 +53,7 @@ from .other_gates import (
     GLOBALPHASE,
     TOFFOLI,
     FREDKIN,
-    OracleGate,
+    get_oracle_gate,
 )
 
 GATE_CLASS_MAP = {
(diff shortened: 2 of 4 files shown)


anshumansahu-stack:
Below is the reference link of the above feature:
https://learn.microsoft.com/en-us/azure/quantum/concepts-oracles

Mayank447:
Implementing oracles will have to wait until #362 is complete as they are not fundamental gates and ideally should be implemented as an Op.

BoxiLi:
@Mayank447  Should we block this completely? Would it be very hard to promote it later to Ops if needed?

But nevertheless, @anshumansahu-stack `OracleGate` is not a gate class by itself but rather a factory that generates gates. You can refer to `get_controlled_gate` for implementation conventions. Name should be something like `get_oracle_gate` (or later `get_oracle_op` )

anshumansahu-stack:
Hey, @BoxiLi, I have renamed the class names as per your suggestion to get_oracle_gate.

I have begun studying Discussion #362 and am understanding the way you want the library to have a layered approach using SubRoutines.Once i completely go through all the suggested features, I am planning to discuss with you and @Mayank447 and make a PR on the same there. My further Cosine-Sine-Decomposition technique for the Pulse Logic as we had discussed may also be affected by this, so i was interested in contributing to this issue first.

After that, I can upgrade this PR to match the latest version after that one so that these features can be integrated flawlessly. If you guys have any work in progress on the same, Ill be happy to continue with that as well. Till then, id suggest either to merge this now and refactor later, or keep this on hold till the Subroutines are constructed.

Mayank447:
> @Mayank447 Should we block this completely? Would it be very hard to promote it later to Ops if needed?

I would recommend completely blocking this. Couple of reasons:

1. Essentially what @anshumansahu-stack has implemented is the Boolean Oracle (|x>|y> -> |x>|y + f(x)>). I have only seen this Oracle been used in Deutsch Josza, nowhere else. So the function should be renamed to `get_boolean_oracle_gate`. Second, there is no check on the logic function being a Boolean function in the current implementation, I could very well pass `np.exp` and it would happy accept it. This check is exponential in no. of qubits `n` (brute forcing all the bitstrings of length n) and I am not sure whether there is a more general efficient method.

2. If we want to implement a Random Oracle, it needs to be a subclass of Ops, given the number of qubits to Oracle are variable and it's parameterized in the input function. Also preferably the [framework](https://arxiv.org/pdf/2112.06700) to define a random oracle needs to be discussed since it's non-trivial (Quantum Random Oracle Model encompasses all of theoretical Quantum Encryption Schemes).

anshumansahu-stack:
Thank you so much for the review. 
I'd probably seek guidance on the first prototype of the implementation of the layered structure that I am currently working on; for that, I'll ask you in the discussion itself.
About the oracleGate, I'll keep this on hold and consider @Mayank447's suggestions as of now.

PR #406, opened 2026-06-08
Refactor einsum execution path

mudit06mah:
**Description**
This PR refactors the `_evolve_state_einsum` function to use `qutip.einsum()` and handle other data types.

Changes:
- Adds `_generate_einsum_eq()` to generate string einsum equation.
- Modifies `_evolve_state_einsum()` to handle 3 different dtypes: Dense, jax and CuState.
- Adds test cases for the 3 dtypes
- Modifies `pyproject.toml` to add jax and cuquantum dependencies.

**Related issues or PRs**
Part of GSoC 2026

**AI Tools Usage Disclosure**
Gemini 3.1 Pro: Implementation discussion and debugging.

### DIFF
--- a/doc/source/qip-simulator.rst
+++ b/doc/source/qip-simulator.rst
@@ -348,3 +348,82 @@ One can save it in a ``.qasm`` file and import it using the following code:
     from qutip_qip.qasm import read_qasm
 
     qc = read_qasm("source/w-state.qasm")
+
+
+GPU-accelerated and Tensor-based Simulation
+===========================================
+
+What if you want to simulate a huge circuit with 20+ qubits? Traditional matrix
+multiplication can become a memory-guzzling bottleneck because expanding the
+gates to the full Hilbert space creates massive matrices.
+
+To save the day, the :class:`.CircuitSimulator` uses tensor contraction
+(Einstein summation, or ``einsum``) by default for both state vector and
+density matrix simulations. Instead of expanding gates, it contracts the gate
+directly with the target qubits.
+
+Even better, we can pair this with GPU backends to turbocharge the simulation!
+Note that because of GPU data transfer and compilation overheads, you will
+typically see a speedup for large circuits (specifically 20+ qubits) compared
+to standard CPU simulation.
+
+Using JAX (CPU/GPU)
+-------------------
+
+With `qutip-jax <https://github.com/qutip/qutip-jax>`_, we can run simulations
+using JAX's JIT compiler. Simply convert your state to a JAX array:
+
+.. code-block:: python
+
+    import qutip_jax
+    from qutip import tensor, basis
+    from qutip_qip.circuit import QubitCircuit, CircuitSimulator
+    import qutip_qip.operations.gates as gates
+
+    # Create a 20-qubit circuit
+    qc = QubitCircuit(20)
+    qc.add_gate(gates.H, targets=0)
+    for q in range(19):
+        qc.add_gate(gates.CX, controls=q, targets=q+1)
+
+    # Convert initial state to JAX
+    state = tensor([basis(2, 0)] * 20).to("jax")
+
+    sim = CircuitSimulator(qc, mode="state_vector_simulator")
+
+    # First run compiles the circuit (takes a moment)
+    _ = sim.run(state)
+
+    # Second run is lightning fast on your GPU!
+    result = sim.run(state)
+
+Using cuQuantum (GPU)
+---------------------
+
+For NVIDIA GPU enthusiasts, `qutip-cuquantum <https://github.com/qutip/qutip-cuquantum>`_
+links directly with NVIDIA's cuQuantum SDK for state-vector simulations.
+
+.. code-block:: python
+
+    import qutip_cuquantum
+    from cuquantum.densitymat import WorkStream
+    from qutip import tensor, basis
+    from qutip_qip.circuit import QubitCircuit, CircuitSimulator
+    import qutip_qip.operations.gates as gates
+
+    # Set up the cuQuantum default context
+    qutip_cuquantum.set_as_default(WorkStream())
+
+    try:
+        qc = QubitCircuit(20)
+        qc.add_gate(gates.H, targets=0)
+        for q in range(19):
+            qc.add_gate(gates.CX, controls=q, targets=q+1)
+
+        # Convert to CuState for GPU simulation
+        state = tensor([basis(2, 0)] * 20).to("CuState")
+
+        sim = CircuitSimulator(qc, mode="state_vector_simulator")
+        result = sim.run(state)
+    finally:
+        qutip_cuquantum.set_as_default(reverse=True)
--- a/doc/source/qip-simulator.rst
+++ b/doc/source/qip-simulator.rst
@@ -30,7 +30,7 @@ It corresponds to the following circuit:
 .. plot::
     :include-source: False
     :context: reset
-    
+
 
     from qutip_qip.circuit import QubitCircuit
     from qutip_qip.operations.gates import X, CX, CH, QASMU, TOFFOLI
@@ -47,7 +47,7 @@ It corresponds to the following circuit:
     qc.draw()
 
 
-    
+
 
 We will add the measurement gates later. This circuit prepares the W-state
 :math:`\newcommand{\ket}[1]{\left|{#1}\right\rangle} (\ket{001} + \ket{010} + \ket{100})/\sqrt{3}`.
@@ -283,7 +283,7 @@ followed just by measurement on the first qubit:
 .. testoutput::
     :options: +NORMALIZE_WHITESPACE
 
-    Quantum object: dims=[[2, 2, 2], [2, 2, 2]], shape=(8, 8), type='oper', dtype=Dense, isherm=True
+    Quantum object: dims=[[2, 2, 2], [2, 2, 2]], shape=(8, 8), type='oper', dtype=CSR, isherm=True
     Qobj data =
     [[0.      0.      0.      0.      0.      0.      0.      0.     ]
      [0.      0.33333 0.      0.      0.      0.      0.      0.     ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ license = "BSD-3-Clause"
 license-files = ["LICENSE"]
 requires-python = ">=3.11"
 # When Python version is updated same change needs to be made in classifier tags in pyproject.toml,
-# build.yaml and in contributing to code, Installation guide. 
+# build.yaml and in contributing to code, Installation guide.
 dependencies = [
     "numpy>=1.24",
     "scipy>=1.9.2",
@@ -70,6 +70,8 @@ Donate = "https://github.com/sponsors/qutip"
 graphics = ["matplotlib>=3.6"]
 control = ["qutip-qtrl>=0.1.5"]
 qiskit = ["qiskit>=2.0"]
+jax = ["qutip-jax", "jax", "jaxlib"]
+cuquantum = ["qutip-cuquantum", "cuquantum", "cupy"]
 full = [
     "qutip-qip[graphics,qiskit,control]",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,8 +70,7 @@ Donate = "https://github.com/sponsors/qutip"
 graphics = ["matplotlib>=3.6"]
 control = ["qutip-qtrl>=0.1.5"]
 qiskit = ["qiskit>=2.0"]
-jax = ["qutip-jax", "jax", "jaxlib"]
-cuquantum = ["qutip-cuquantum", "cuquantum", "cupy"]
+cuquantum = ["qutip-cuquantum>=0.1.0", "cuquantum>=23.3.0", "cupy>=12.2.0"]
 full = [
     "qutip-qip[graphics,qiskit,control]",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,7 +70,7 @@ Donate = "https://github.com/sponsors/qutip"
 graphics = ["matplotlib>=3.6"]
 control = ["qutip-qtrl>=0.1.5"]
 qiskit = ["qiskit>=2.0"]
-cuquantum = ["qutip-cuquantum>=0.1.0", "cuquantum>=23.3.0", "cupy>=12.2.0"]
+cuquantum = ["qutip-cuquantum>=0.1.0"]
 full = [
     "qutip-qip[graphics,qiskit,control]",
 ]
--- a/src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
+++ b/src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
@@ -309,15 +309,14 @@ class CircuitSimulator:
     def _evolve_state_einsum(self, gate, targets_indices, state):
         gate_qobj = gate.get_qobj()
 
-        orignal_dims = state.dims
-        orignal_shape = state.shape
+        original_dims = state.dims
+        original_shape = state.shape
 
         # Generate equation
         num_dims = len(self._tensor_dims)
         eq = self._generate_einsum_eq(targets_indices, num_dims)
 
         data_type = type(state.data).__name__
-        print(data_type)
 
         if data_type == "JaxArray":
             import jax.numpy as jnp
@@ -333,8 +332,8 @@ class CircuitSimulator:
 
             new_state_tensor = jnp.einsum(eq, gate_tensor, state_tensor)
 
-            reshaped_data = JaxArray(new_state_tensor.reshape(orignal_shape))
-            state = Qobj(reshaped_data, dims=orignal_dims)
+            reshaped_data = JaxArray(new_state_tensor.reshape(original_shape))
+            state = Qobj(reshaped_data, dims=original_dims)
 
         elif data_type == "CuState":
             import cupy as cp
@@ -351,17 +350,17 @@ class CircuitSimulator:
 
             new_state_tensor = contract(eq, gate_tensor, state_tensor)
 
-            reshaped_data = CuState(new_state_tensor.reshape(orignal_shape))
-            state = Qobj(reshaped_data, dims=orignal_dims)
+            reshaped_data = CuState(new_state_tensor.reshape(original_shape))
+            state = Qobj(reshaped_data, dims=original_dims)
 
         else:
             from qutip.core.dimensions import einsum
 
             malformed_state = einsum(eq, gate_qobj, state)
             reshaped_data = _data.reshape(
-                malformed_state.data, orignal_shape[0], orignal_shape[1]
+                malformed_state.data, original_shape[0], original_shape[1]
             )
-            state = Qobj(reshaped_data, dims=orignal_dims)
+            state = Qobj(reshaped_data, dims=original_dims)
 
         return state
 
--- a/src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
+++ b/src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
@@ -262,6 +262,9 @@ class CircuitSimulator:
         self._op_index += 1
 
     def _generate_einsum_eq(self, targets, num_qubits):
+        """
+        Generates the einsum string for tensor contraction.
+        """
         l_chars = string.ascii_lowercase
         u_chars = string.ascii_uppercase
 
--- a/src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
+++ b/src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
@@ -305,7 +305,10 @@ class CircuitSimulator:
         U: Qobj
             unitary to be applied.
         """
-        U = operation.get_qobj()
+        data_type = type(state.data).__name__
+        gate_dtype = "CuOperator" if data_type == "CuState" else "Dense"
+
+        U = operation.get_qobj().to(gate_dtype)
         U = expand_operator(
             U,
             dims=self.dims,
--- a/src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
+++ b/src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
@@ -6,6 +6,7 @@ import string
 
 from qutip import ket2dm, Qobj
 from qutip.core import data as _data
+from qutip.core.dimensions import einsum
 from qutip_qip.circuit.simulator import CircuitResult
 from qutip_qip.operations import expand_operator
 
@@ -263,13 +264,25 @@ class CircuitSimulator:
 
     def _generate_einsum_eq(self, targets, num_qubits):
         """
-        Generates the einsum string for tensor contraction.
-        """
-        l_chars = string.ascii_lowercase
-        u_chars = string.ascii_uppercase
+        Generates the einsum string for tensor contraction supporting up to 52 qubits.
+        Uses standard ASCII letters (a-z, A-Z) to map input and output indices.
 
-        state_in = list(l_chars[:num_qubits])
-        gate_out = list(u_chars[: len(targets)])
+        Parameters
+        ----------
+        targets : list of int
+            The target qubits the gate acts on.
+        num_qubits : int
+            The total number of qubits (tensor dimensions) in the state.
+
+        Returns
+        -------
+        eq : str
+            The einsum equation string (e.g., "ab,cde->cde").
+        """
+        chars = string.ascii_letters
+
+        state_in = list(chars[:num_qubits])
+        gate_out = list(chars[num_qubits : num_qubits + len(targets)])
 
         gate_in = [state_in[t] for t in targets]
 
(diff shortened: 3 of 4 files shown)


src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -283,33 +307,61 @@ def _evolve_state(self, operation, targets_indices, state):
         return state
 
     def _evolve_state_einsum(self, gate, targets_indices, state):
```
veronikakurth: Would be nice to add type hints

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -283,33 +307,61 @@ def _evolve_state(self, operation, targets_indices, state):
...
+
+        elif data_type == "CuState":
+            import cupy as cp
+            from cuquantum.tensornet import contract
+            from qutip_cuquantum import CuState
+
+            gate_qobj = gate_qobj.to("CuOperator")
+
+            raw_state = state.data.to_cupy()
+            raw_gate = cp.array(gate_qobj.full())
```
veronikakurth: Is it possible to avoid converting to the dense representation?

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -283,33 +307,61 @@ def _evolve_state(self, operation, targets_indices, state):
...
+
+        data_type = type(state.data).__name__
+
+        if data_type == "JaxArray":
+            import jax.numpy as jnp
+            from qutip_jax.jaxarray import JaxArray
+
+            gate_qobj = gate_qobj.to("jax")
+
+            raw_state = state.data._jxa
```
veronikakurth: Is there a way to access the same data via a public property instead of `_jxa`?

veronikakurth (changes requested):

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -2,8 +2,10 @@
 from operator import mul
 from functools import reduce
 import numpy as np
+import string
 
 from qutip import ket2dm, Qobj
+from qutip.core import data as _data
```
Mayank447, line 8: Any particular reason for importing as a private variable `_data`. Maybe

```suggestion
import qutip.core as qt
```

then reference it as `qt.data`?

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -259,6 +261,28 @@ def step(self):
         self._state = state
         self._op_index += 1
 
+    def _generate_einsum_eq(self, targets, num_qubits):
```
Mayank447: Add type hints.

Note: This scheme only works upto 26 qubits (but that should be sufficient for most classical simulation).

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -283,33 +307,61 @@ def _evolve_state(self, operation, targets_indices, state):
...
+
+        # Generate equation
+        num_dims = len(self._tensor_dims)
+        eq = self._generate_einsum_eq(targets_indices, num_dims)
+
+        data_type = type(state.data).__name__
+
+        if data_type == "JaxArray":
+            import jax.numpy as jnp
+            from qutip_jax.jaxarray import JaxArray
```
Mayank447: `gate.get_qobj()` allows for an argument for dtype. So it will be better to check state data type first and then access `gate_qobj`

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -283,33 +307,61 @@ def _evolve_state(self, operation, targets_indices, state):
...
+        original_dims = state.dims
+        original_shape = state.shape
+
+        # Generate equation
+        num_dims = len(self._tensor_dims)
+        eq = self._generate_einsum_eq(targets_indices, num_dims)
+
+        data_type = type(state.data).__name__
+
+        if data_type == "JaxArray":
```
Mayank447: String comparison for datatype checks are fragile (not recommended).

```suggestion
       type(state.data) is JaxArray:
```

Same change for `CuState`

pyproject.toml
```diff
@@ -70,6 +70,8 @@ Donate = "https://github.com/sponsors/qutip"
 graphics = ["matplotlib>=3.6"]
 control = ["qutip-qtrl>=0.1.5"]
 qiskit = ["qiskit>=2.0"]
+jax = ["qutip-jax", "jax", "jaxlib"]
+cuquantum = ["qutip-cuquantum", "cuquantum", "cupy"]
```
Mayank447: Add minimum version

tests/test_circuit.py
```diff
@@ -813,6 +827,50 @@ def test_circuit_chain_structure(self):
...
+            qc.add_gate(gates.X, targets=0)
+
+            state = basis(2, 0).to(dtype)
+
+            sim = CircuitSimulator(qc, mode="state_vector_simulator")
+            result = sim.run(state)
+
+            final_state = result.get_final_states()[0]
+
+            assert type(final_state.data) == type(state.data)
```
Mayank447: This only tests the dispatcher's data type. Not the final state i.e. |1>

tests/test_circuit.py
```diff
@@ -813,6 +827,50 @@ def test_circuit_chain_structure(self):
         assert qc2.input_states == [None] * 3
 
 
+class TestEinsumBackend:
+    """
+    Test suite for the einsum execution path and backend data types.
+    """
+
+    AVAILABLE_DTYPES = ["Dense"]
```
Mayank447, line 840:
```suggestion
    AVAILABLE_DTYPES = ["Dense", "CSR"]
```

tests/test_circuit.py
```diff
@@ -813,6 +827,50 @@ def test_circuit_chain_structure(self):
...
+    def test_dtype_preservation(self, dtype):
+        """
+        Validate that the dtype of the input state is preserved.
+        """
+        if dtype == "CuState":
+            import qutip_cuquantum
+            from cuquantum.densitymat import WorkStream
+
+            # Set the WorkStream as the default context for the backend
+            qutip_cuquantum.set_as_default(WorkStream())
```
Mayank447, line 858: This should ideally be made a fixture, since it affects the global state.

tests/test_circuit.py
```diff
@@ -813,6 +827,50 @@ def test_circuit_chain_structure(self):
...
+
+        try:
+            qc = QubitCircuit(1)
+            qc.add_gate(gates.X, targets=0)
```
Mayank447: Add test for 2, 3 qubit gates ideally within the same circuit.

Mayank447 (changes requested):

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
Mayank447: I don't think `.to` method in qutip supports cupy.

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
veronikakurth: This comes from the convention in `qutip` to access the methods of the data layer: [example](https://github.com/qutip/qutip/blob/69331146cc527c7683f8b30d328a193fcec0c907/qutip/solver/spectrum.py#L8).
In my understanding, this is to indicate that it is not a public API of `qutip` and should be used only internally. 

Hence, ideally, `qutip-qip` should not access the methods of `qutip`'s data layer directly and rather rely on public methods of `qutip` when it comes to data manipulation. 

However, since we decided on a workaround for now due to limitations of `qutip.einsum` on not handling arbitrary backends, we are somewhat forced to access methods of the `qutip`'s data layer, and especially in this case, we should reflect in the aliasing that it's not a public API.

Hence, I'd suggest leaving it the way it is. :)

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -259,6 +281,73 @@ def step(self):
...
+                if data_type == "JaxArray":
+                    import jax.numpy as jnp
+
+                    self._gate_cache[cache_key] = jnp.array(qobj.full())
+                elif data_type == "CuState":
+                    import cupy as cp
+
+                    self._gate_cache[cache_key] = cp.array(qobj.full())
+                else:
+                    self._gate_cache[cache_key] = qobj
```
Mayank447: This function is not needed. `gate.get_qobj(datatype)` already does lru caching.

Just as a sidenote if you want to cache anything, functools provides cache function which you can use as a decorator.

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -129,11 +140,20 @@ def state(self):
...
+        if self._raw_tensor is not None:
+            data_type = type(self._state.data).__name__
+            original_shape = self._state.shape
+            if data_type == "JaxArray":
+                from qutip_jax import JaxArray
+
+                reshaped = JaxArray(self._raw_tensor.reshape(original_shape))
+                return Qobj(reshaped, dims=self._state_dims)
+            elif data_type == "CuState":
+                from qutip_cuquantum import CuState
```
Mayank447: Change the datatype check to either `type` (preferred) or `instance` as previously mentioned.

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -158,6 +178,8 @@ def run(self, state, cbits=None, measure_results=None):
             output state and probability.
         """
         self.initialize(state, cbits, measure_results)
+        data_type = type(state.data).__name__
```
Mayank447: Same as above

Mayank447 (changes requested):
I will do a detailed review later. Couple of things I noticed at a glance

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -305,7 +305,10 @@ def _evolve_state(self, operation, targets_indices, state):
         U: Qobj
             unitary to be applied.
         """
-        U = operation.get_qobj()
```
BoxiLi: Redundant

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -259,6 +262,40 @@ def step(self):
...
+        for i, t in enumerate(targets):
+            state_out[t] = gate_out[i]
+
+        gate_str = "".join(gate_out + gate_in)
+        state_in_str = "".join(state_in)
+        state_out_str = "".join(state_out)
+
+        return f"{gate_str}, {state_in_str}... -> {state_out_str}..."
+
     def _evolve_state(self, operation, targets_indices, state):
```
veronikakurth: Please add type hints. Thanks!

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -268,7 +305,10 @@ def _evolve_state(self, operation, targets_indices, state):
         U: Qobj
             unitary to be applied.
         """
-        U = operation.get_qobj()
+        data_type = type(state.data).__name__
+        gate_dtype = "CuOperator" if data_type == "CuState" else "Dense"
```
BoxiLi: If it is not Dense, we should just keep the default type and not converting the dense. It does no big harm, but just a bit unnecessary.

pyproject.toml
Mayank447: I saw you earlier added the minimum version in a commit. Why did you remove it again?

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
mudit06mah: Good catch!

BoxiLi:
You can also update this PR with the compatible version for the qutip PR branch. So that we also have the code. Test failing here is fine. We can then try offline.

tests/test_circuit.py
```diff
@@ -813,6 +827,50 @@ def test_circuit_chain_structure(self):
...
+    if HAS_JAX:
+        AVAILABLE_DTYPES.append("jax")
+    if HAS_CUQANTUM:
+        AVAILABLE_DTYPES.append("CuState")
+
+    @pytest.mark.filterwarnings(
+        "ignore:ExternalStream is deprecated:DeprecationWarning"
+    )
+    @pytest.mark.parametrize("dtype", AVAILABLE_DTYPES)
+    def test_dtype_preservation(self, dtype):
```
veronikakurth: Suggestion: split tests into separate tests for every data type - may be easier to monitor which ones are skipped.

BoxiLi:
Merge it against the `einsum` branch

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -282,34 +393,79 @@ def _evolve_state(self, operation, targets_indices, state):
...
+
+        malformed_state = einsum(eq, gate_qobj, state)
+        reshaped_data = _data.reshape(
+            malformed_state.data, original_shape[0], original_shape[1]
         )
+        state = Qobj(reshaped_data, dims=original_dims).to(state_dtype)
+
+        return state
+
+    def _evolve_state_einsum_dm(self, operation, targets_indices, state):
```
veronikakurth: It would not be a reason not to merge the changes, but maybe there is a way to unify `_evolve_state_einsum` for state vectors and density matrices in order to maintain less duplicate code in the future - if not now, maybe as a separate PR. But only if time permits.

doc/source/qip-simulator.rst
```diff
@@ -347,4 +347,83 @@ One can save it in a ``.qasm`` file and import it using the following code:
...
+    try:
+        qc = QubitCircuit(20)
+        qc.add_gate(gates.H, targets=0)
+        for q in range(19):
+            qc.add_gate(gates.CX, controls=q, targets=q+1)
+
+        # Convert to CuState for GPU simulation
+        state = tensor([basis(2, 0)] * 20).to("CuState")
+
+        sim = CircuitSimulator(qc, mode="state_vector_simulator")
```
veronikakurth, line 426: (for future improvements by maintainers; just my opinion that doesn't have to be addressed before the final submission) 
From the user's perspective, I guess it would be more convenient not to have to specify the `mode` for `CircuitSimulator` explicitly when working with density matrices and to infer the mode from the passed state, but I guess it might be a bit error-prone to guess what the user wants, so we may want the user to be explicit. Probably it's the complexity we have to live with for now. 

Alternatively, to avoid using strings for specifying the mode, I think there could exist an alternative constructor that is a class method of `CircuitSimulator` and returns a circuit simulator with a selected mode. This doesn't remove the requirement for the user to be explicit, but may be a bit easier to code. 

Something like:

`CircuitSimulator.state_simulator(...)`
`CircuitSimulator.dm_simulator(...)`

This is just off the top of my head, so feel free to disagree. 
Just wanted to share an alternative idea for the user interface.

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -259,16 +265,121 @@ def step(self):
         self._state = state
         self._op_index += 1
 
-    def _evolve_state(self, operation, targets_indices, state):
+    def _generate_einsum_eq(self, targets, num_qubits, is_oper=False):
         """
-        Applies unitary to state.
+        Generates the einsum string for tensor contraction supporting up to 52 qubits.
```
BoxiLi, line 283: Is this number of 52 correct? Do we need also letters of the gate subsystem index?

pyproject.toml
```diff
@@ -70,6 +70,7 @@ Donate = "https://github.com/sponsors/qutip"
 graphics = ["matplotlib>=3.6"]
 control = ["qutip-qtrl>=0.1.5"]
 qiskit = ["qiskit>=2.0"]
+cuquantum = ["qutip-cuquantum>=0.1.0", "cuquantum>=23.3.0", "cupy>=12.2.0"]
```
BoxiLi: Is it sufficient to require only qutip-cuquantum, and let it automatically download the right cuquantum and cupy?

Also, we need a qutip-jax optional dependency

BoxiLi (commented):
We need some more tests. E.g. for circuit run with numpy/jax/cuquantum(skipped) for statevector and dm.

pyproject.toml
BoxiLi:
```
cuquantum = ["qutip-cuquantum>=0.1.0"]
```
Drop the JAX optional dependency for now.

tests/test_circuit.py
```diff
@@ -813,6 +832,171 @@ def test_circuit_chain_structure(self):
...
+        qc.add_gate(gates.CX, controls=0, targets=1)
+
+        state = ket2dm(tensor(basis(2, 0), basis(2, 0))).to(dtype)
+
+        sim = CircuitSimulator(qc, mode="density_matrix_simulator")
+        result = sim.run(state)
+
+        final_state = result.get_final_states()[0]
+
+        assert type(final_state.data) == type(state.data)
```
BoxiLi: Compare also if they are the same array.

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
Mayank447: The number 52 is correct for ket. For operator it is 25 qubits when applying a two qubit gate.

tests/test_circuit.py
```diff
@@ -27,12 +27,31 @@
...
+    HAS_JAX = True
+except ImportError:
+    HAS_JAX = False
+
+try:
+    import qutip_cuquantum
+
+    HAS_CUQANTUM = True
+except ImportError:
+    HAS_CUQANTUM = False
```
Mayank447:
```suggestion
    HAS_CUQUANTUM = True
except ImportError:
    HAS_CUQUANTUM = False
```

Same change at few other places

tests/test_circuit.py
```diff
@@ -813,6 +832,174 @@ def test_circuit_chain_structure(self):
...
+
+            init_state_cpu = tensor(basis(2, 0), basis(2, 0), basis(2, 1))
+            state = init_state_cpu.to(dtype)
+
+            sim = CircuitSimulator(qc, mode="state_vector_simulator")
+            result = sim.run(state)
+
+            final_state = result.get_final_states()[0]
+
+            assert type(final_state.data) == type(state.data)
```
Mayank447:
```suggestion
            assert type(final_state.data) is type(state.data)
```

Minor nitpick: We don't have a linter currently (in process of adding ruff), but any linter would flag this as `==` can be overwritten

tests/test_circuit.py
```diff
@@ -813,6 +832,174 @@ def test_circuit_chain_structure(self):
...
+            qc.add_gate(gates.CX, controls=2, targets=0)
+            qc.add_gate(gates.TOFFOLI, controls=[2, 0], targets=1)
+
+            dm_init_cpu = ket2dm(tensor(basis(2, 0), basis(2, 0), basis(2, 1)))
+            dm_init = dm_init_cpu.to(dtype)
+
+            sim_dm = CircuitSimulator(qc, mode="density_matrix_simulator")
+            res_einsum = sim_dm.run(dm_init).get_final_states(0)
+
+            assert type(res_einsum.data) == type(dm_init.data)
```
Mayank447:
```suggestion
            assert type(res_einsum.data) is type(dm_init.data)
```

tests/test_circuit.py
```diff
@@ -813,6 +832,174 @@ def test_circuit_chain_structure(self):
...
+            qc.add_gate(gates.TOFFOLI, controls=[2, 0], targets=1)
+
+            init_oper_cpu = identity([2, 2, 2])
+            oper = init_oper_cpu.to(dtype)
+
+            sim = CircuitSimulator(qc, mode="state_vector_simulator")
+            res_einsum = sim.run(oper).get_final_states(0)
+
+            assert res_einsum.isoper
+            assert type(res_einsum.data) == type(oper.data)
```
Mayank447:
```suggestion
            assert type(res_einsum.data) is type(oper.data)
```

doc/source/qip-simulator.rst
```diff
@@ -283,7 +283,7 @@ followed just by measurement on the first qubit:
 .. testoutput::
     :options: +NORMALIZE_WHITESPACE
 
-    Quantum object: dims=[[2, 2, 2], [2, 2, 2]], shape=(8, 8), type='oper', dtype=Dense, isherm=True
+    Quantum object: dims=[[2, 2, 2], [2, 2, 2]], shape=(8, 8), type='oper', dtype=CSR, isherm=True
```
Mayank447, line 286: Just a curious question what lead to this `dtype` change? Using `CSR` is definitely better for this Qobj.

I just wanted to understand where exactly dtype changes now cause we haven't modified `dtype` for gates or `zero_state`.

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -99,9 +103,19 @@ def initialize(self, state=None, cbits=None, measure_results=None):
...
                 state = np.exp(1j * self.qc.global_phase) * state
                 self._state = state
+
+            if type(self._state.data).__name__ == "CuState":
+                from qutip_cuquantum.state import CuState as CuStateClass
+
+                hilbert_dims = tuple(state.dims[0])
+                self._state = Qobj(
+                    CuStateClass(self._state.data.to_cupy(), hilbert_dims=hilbert_dims),
+                    dims=self._state.dims,
```
Mayank447, line 114: I didn't get what is this block doing? Is it because after multiplying the state by the global phase, scalar multiplication on a `CuState` returns something that doesn't have tensor-structure metadata?

Also in case of `density_matrix` or `global_phase == 0`, we can skip rebuilding a fresh `CuState` here.

tests/test_circuit.py
```diff
@@ -813,6 +832,174 @@ def test_circuit_chain_structure(self):
...
+
+            expected_oper = U_T_exp * U_CX_exp * U_H_exp * init_oper_cpu
+
+            np.testing.assert_allclose(
+                res_einsum.full(), expected_oper.full(), atol=1e-12
+            )
+        finally:
+            if dtype == "CuState":
+                qutip_cuquantum.set_as_default(reverse=True)
+
```
Mayank447, line 996: If we are skipping this test for `CuState`, there is no need for`if dtype == "CuState"` in rest of the function code.

tests/test_circuit.py
```diff
@@ -813,6 +832,174 @@ def test_circuit_chain_structure(self):
...
+            sim = CircuitSimulator(qc, mode="state_vector_simulator")
+            result = sim.run(state)
+
+            final_state = result.get_final_states()[0]
+
+            assert type(final_state.data) == type(state.data)
+
+            if dtype == "CuState":
+                qutip_cuquantum.set_as_default(reverse=True)
```
Mayank447, line 882: There is no need for this in try, `finally` block already does this. Same change needs to be made at few other places

Mayank447 (changes requested):
Sorry for the delay in reviewing, LGTM.

Couple of additional points:
- Is there a reason the target merge branch is not `master`?
- I see the typehints have been added in the docstrings, @mudit06mah could those be added in the function declaration as well.
- Could you add few comments in code for where we do some workaround especially for `CuQuantum`, it will be easier for people in future as code is read more times than its written.

doc/source/qip-simulator.rst
mudit06mah: @BoxiLi  Pointed out that this is because of the 3 qubit Toffoli gate which was instantiated as a `Dense` matrix and since the circuit was also 3 qubits, it skipped the expand operator and was not being converted to `CSR`.

src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
mudit06mah: Yes, When a state is constructed via `qutip.tensor(...)` or scaled by `exp(1j * global_phase)`, the `CuState` loses its `hilbert_dims` metadata (since there's no `_data.kron` specialization for `(CuState, CuState)`).

Because `qutip.tensor` drops this metadata even before phase multiplication happens, we cannot skip rebuilding when `global_phase == 0` or for density matrices. Hence this block reconstructs the `CuState` using the `hilbert_dims = tuple(state.dims[0])` to avoid tensor shape mismatch.

tests/test_circuit.py
mudit06mah: The `qutip_cuquantum.set_as_default(reverse=True)` call inside the try block is actually necessary before computing the CPU reference matrices (expected_state / expected_dm).
If the `cuQuantum` default context remains active, `expand_operator(...)` creates `CuOperator` instances. For non-contiguous/inverted targets like `CX(targets=[2, 0])` and `TOFFOLI(targets=[2, 0, 1])`, `CuOperator.permute()` raises an `IndexError`. We reverse the context there specifically so the baseline matrix multiplications evaluate on standard CPU Dense backend.

I will add a comment to clarify this in the code.

mudit06mah:
@Mayank447  Thank you for your review!

The target branch was changed from master to einsum because earlier we thought the tests will fail as the update `qutip.einsum()` which uses the dispatcher was not yet released, but no test actually fail because the tests for JAX and CuState don't run on github workflows so it would've worked out if it was in master anyway ig.

Regarding the typehints and comments, I have added both :)

BoxiLi (approved):

BoxiLi:
Yeah this a tricky question, I think we will merge this to master only after qutip and qutip-jax is released and we can formally tests them. Cuquantum is a bit hard because we have no GPU runner. So we have to rely on local tests.

Mayank447:
> Yeah this a tricky question, I think we will merge this to master only after qutip and qutip-jax is released and we can formally tests them. Cuquantum is a bit hard because we have no GPU runner. So we have to rely on local tests.

@BoxiLi I think we can add a CI workflow temporarily that tests using master branch of qutip, qutip-jax. I don't think qutip 5.4 is getting released anytime soon. This will avoid branches diverging and the merge conflicts.

tests/test_circuit.py
```diff
@@ -813,6 +832,176 @@ def test_circuit_chain_structure(self):
...
+                gates.TOFFOLI.get_qobj(), dims=[2, 2, 2], targets=[2, 0, 1]
+            )
+
+            expected_oper = U_T_exp * U_CX_exp * U_H_exp * init_oper_cpu
+
+            np.testing.assert_allclose(
+                res_einsum.full(), expected_oper.full(), atol=1e-12
+            )
+        except Exception:
+            raise
```
Mayank447: Raise an error here, otherwise you can actually remove this except block.

Mayank447 (approved):

BoxiLi:
> @BoxiLi I think we can add a CI workflow temporarily that tests using master branch of qutip, qutip-jax. I don't think qutip 5.4 is getting released anytime soon. This will avoid branches diverging and the merge conflicts.

But still even in this case the master tests will only partially pass?

Mayank447:
> But still even in this case the master tests will only partially pass?

Since the changes are already in qutip and qutip-jax's master branch wouldn't everything pass? Only cuquantum tests won'tt be tested right

BoxiLi:
Oh I see we can create *one* test that test against master qutip and qutip-jax. I agree. Let's finish this PR and merge to qutip-einsum first, then I will do that action udpate separately and bring this back to main if possible. So Mudit can continue working on the next PR without interuption

Mayank447:
Sounds good

tests/test_circuit.py
mudit06mah: Oh sorry that was actually some leftover code that I forgot to clean up!

PR #407, opened 2026-06-17
Add QuTiP 5-compatible solver options handling

BoxiLi:
qutip dropped the old interface `qutip.Options`. qutip-qip also dropped the support of old qutipv4, but it is not yet released. This patch is to fix qutip-v0.4

Ignore some deprecation warnings caused by matplotlib due to Python 3.9 env.

The failing VQA tests are fixed in #280, and will be included in the release. (Test will pass, verified here https://github.com/qutip/qutip-qip/actions/runs/27660840477)

GPT-5.4 is used to fix the pytest and matplotlib warning issue.

### DIFF
--- a/src/qutip_qip/device/cavityqed.py
+++ b/src/qutip_qip/device/cavityqed.py
@@ -13,7 +13,6 @@ from qutip import (
     Qobj,
     QobjEvo,
 )
-from ..compat import solver_options
 from ..circuit import QubitCircuit
 from ..operations import Gate
 from .processor import Processor, Model
--- a/src/qutip_qip/device/cavityqed.py
+++ b/src/qutip_qip/device/cavityqed.py
@@ -13,6 +13,7 @@ from qutip import (
     Qobj,
     QobjEvo,
 )
+from ..compat import solver_options
 from ..circuit import QubitCircuit
 from ..operations import Gate
 from .processor import Processor, Model
@@ -70,7 +71,7 @@ class DispersiveCavityQED(ModelProcessor):
         processor.load_circuit(qc)
         result = processor.run_state(
             qutip.basis([10, 2, 2], [0, 0, 0]),
-            options=qutip.Options(nsteps=5000))
+            options=solver_options(nsteps=5000))
         final_qubit_state = result.states[-1].ptrace([1, 2])
         print(round(qutip.fidelity(
             final_qubit_state,
--- a/src/qutip_qip/device/cavityqed.py
+++ b/src/qutip_qip/device/cavityqed.py
@@ -71,7 +71,8 @@ class DispersiveCavityQED(ModelProcessor):
         processor.load_circuit(qc)
         result = processor.run_state(
             qutip.basis([10, 2, 2], [0, 0, 0]),
-            options=solver_options(nsteps=5000))
+            options={"nsteps": 5000}
+        )
         final_qubit_state = result.states[-1].ptrace([1, 2])
         print(round(qutip.fidelity(
             final_qubit_state,
--- a/src/qutip_qip/device/processor.py
+++ b/src/qutip_qip/device/processor.py
@@ -1005,7 +1005,7 @@ class Processor(object):
             final_qu.args.update(args)
 
         # bring all c_ops to the same tlist, won't need it in QuTiP 5
-        if not parse_version(qutip.__version__) >= parse_version("5.dev"):
+        if not is_qutip5:
             temp = []
             for c_op in c_ops:
                 temp.append(_merge_qobjevo([c_op], final_qu.tlist))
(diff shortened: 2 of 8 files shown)


tests/test_gates.py
```diff
@@ -157,7 +157,7 @@ def test_all_elements_different(self):
                 fid = qutip.average_gate_fidelity(gate, other)
                 assert not np.allclose(fid, 1., atol=1e-3)
 
-    @pytest.mark.parametrize("gate", gates.qubit_clifford_group())
+    @pytest.mark.parametrize("gate", tuple(gates.qubit_clifford_group()))
```
veronikakurth, line 160: Is this casting to a tuple also related to keeping things compatible with qutip-v4?

src/qutip_qip/device/cavityqed.py
```diff
@@ -70,7 +71,8 @@ class DispersiveCavityQED(ModelProcessor):
         processor.load_circuit(qc)
         result = processor.run_state(
             qutip.basis([10, 2, 2], [0, 0, 0]),
-            options=qutip.Options(nsteps=5000))
+            options={"nsteps": 5000}
```
veronikakurth, line 73: We import `solver_options` but don't use it in this file. To temporarily maintain v4-compatibility, we should use them in L74, shouldn't we? 
I'm also a bit confused that the tests for v4 pass when using a dictionary here. :)

tests/test_device.py
```diff
@@ -13,7 +13,12 @@
                                 CircularSpinChain, SCQubits)
 
 from packaging.version import parse as parse_version
-from qutip import Options
+
+if parse_version(qutip.__version__) < parse_version("5.dev"):
```
veronikakurth: It looks to me like a duplication of logic from `compat.py` - should we just import `solver_options` from there to this test file and use it here?

src/qutip_qip/device/processor.py
```diff
@@ -1202,7 +1202,7 @@ def run_state(
         # is however, much harder to implement at this stage, see also
         # https://github.com/qutip/qutip-qip/issues/184.
         if is_qutip5:
-            options = kwargs.get("options", qutip.Options())
+            options = kwargs.get("options", {})
```
veronikakurth: If `None` is passed as a value of `kwargs["options"]`, the `options` variable here will not get an empty dictionary (or any other value) as a default value. (`dict.get` checks for the existence of the key)

A safer approach would be:

```python
options = kwargs.get("options")
if options is None:
    options = {}
```
(Unfortunately, resulting in more lines of code.)

P.S. Here is how to check it:
```python
>>> kwargs = {"options": None}
>>> options = kwargs.get("options", {})
>>> options
```

veronikakurth (changes requested):
As long as the changes are temporary, I agree with them. 🙃 
I guess the plan is to drop support of qutip-v4 in the next major release, i.e., `0.5.0`?

src/qutip_qip/device/cavityqed.py
veronikakurth: Or I guess this docstring test just doesn't get executed...

BoxiLi:
>I guess the plan is to drop support of qutip-v4 in the next major release, i.e., 0.5.0?

Yes, this has already been dropped at master, but I don't think there can be a release in one or two months. So this is a temporary fix for v0.4. I don't think it is good to drop support in a micro release.

src/qutip_qip/device/cavityqed.py
BoxiLi: Yes, we only build docstrings with v5

tests/test_device.py
BoxiLi: My bad, I'm hesitating between putting it in one file or splitting it into multiple files. In the end, I tried to just define it in each file to make it just more explicit, since it is just a temporary fix in this released branch. But forgot to delete it. What do you think?

src/qutip_qip/device/processor.py
BoxiLi: Good catch, I didn't think about someone really giving None to this...

tests/test_gates.py
BoxiLi: No this is because pytest has deprecated using non-collection iterables in parameterization

tests/test_device.py
veronikakurth: I see! Since both approaches imply touching that many files, I think I'd be in favor of having one definition imported everywhere where patching is needed. This way, we can be sure it's consistent. 
But to the point of being more explicit: I think in this case we could play a bit with import aliasing or function naming. 

E.g., we could do

```from compat import back_compatible_solver_options``` (very long, I know)

or 
```
import compat as version_adapter

version_adapter.solver_options()
```

tests/test_gates.py
veronikakurth: Ah I see! Probably, they had good reasons to do so :thinking:

veronikakurth (approved):
LGTM! Thanks Boxi. :)

PR #408, opened 2026-06-19
Add add_op method

Mayank447:
**Description**
This PR adds `add_op` method to `QuantumCircuit`. This is essentially a generalization over `add_gate`, `add_measurement` since `Op` (#362 ) is a generalization over gates, measurement. The plan is to deprecate `add_gate`, `add_measurement` once Ops are fully implemented.

Also since `targets`, `controls` are specific to gate types. I have instead used `qubits` as the arguments. The user passes control qubits followed by target qubits in the same list (same convention as OpenQASM). So `qc.add_gate(CX, targets=1, controls=0)` becomes `qc.add_op(CX, qubits=(0, 1))`.

**Related issues or PRs**
None

**AI Tools Usage Disclosure**
None


Mayank447:
@BoxiLi Have a look at the interface. If its fine to you, I will start by replacing `add_gate`, `add_measurement` in the codebase with `add_op`.

BoxiLi:
Thanks for making this! So this is supposed to be step zero, illustrating the user interface?

I would rather argue for keeping `add_gate` and `add_measurement`, instead of deprecating them. 
- The first reason is that people are very used to those names and I don't see strong drawbacks on keep those as user interface. We don't have to be minimalistic here.
- And second, we still have many backward compatibility handling there. I am a bit concerned about overburdening the new method `add_op`.

But internally, I believe we can direct everything to `add_op` if that gives us a more unified data handling (Or maybe we can even simplify that to `add`?). Maybe in the future we can add `extend_circuit(circuit)` etc which eventually all use this interface.

About `qubits`, that sounds reasonable. A slightly different proposal is something like `qureg`, so quantum register. Cause I was thinking that we can generalize the circuit to qudit, not just bit. But I am also fine if we find `qubits` more straightforward for 99% of users.

What about classical control flow?

Mayank447:
> Thanks for making this! So this is supposed to be step zero, illustrating the user interface?

Yes

> I would rather argue for keeping `add_gate` and `add_measurement`, instead of deprecating them.

So the reason I wanted to deprecate them was once Ops are implemented. Most users would be doing somthing like`qc.add_op(QFT, qubits=range(0, 8))`. So instead of them juggling between `add_gate`, `add_measurement`, `add_op`, I wanted to unify the interface.

> But internally, I believe we can direct everything to `add_op` if that gives us a more unified data handling (Or maybe we can even simplify that to `add`?).

I was actually planning of doing this. 

> About `qubits`, that sounds reasonable. A slightly different proposal is something like `qureg`, so quantum register. Cause I was thinking that we can generalize the circuit to qudit, not just bit. But I am also fine if we find `qubits` more straightforward for 99% of users.

Nice suggestion, how about the terms `qreg`, `creg`?

> What about classical control flow?

This is a little complicated, I plan to do it in the next PR. So essentially I did't want `classical_control` as an explicit argument for `add_op` since for most workflows an Op should encompass the classical control logic. Consider the 3 qubit repetition code, we need 2 ancillas for syndrome measurement and 2 cbits for storing measurement result. Let's say we have an Op called `RepetitionCodeOp` of 5 qubit (3 data + 2 ancilla), 2 cbits (syndrome) which implements this logic. As a user I would simply want to add this to my circuit by `qc.add_op(RepetitionCodeOp, qubits=(d1,d2,d3,a1,a2), cbits=(0, 1))`. The `Op` (RepetitionCodeOp) should internally handle the classical control flow.

For the explicit case in which user wants to add a classical control flow, my current plan is to introduce an Op `ClassicallyControlled`, which takes an Op, classical control state (bit string of 0,1s). e.g. `qc.add_op(ClassicallyControlled(QFT, classical_ctrl_value=(1, 0, 1)), qubits=range(0, 8), cbits=(0, 1, 2))` this performs a QFT on qubits 0 to 7 only if the first three cbits state is 101. The way I want to internally implement the `ClassicallyControlled` Op is via labels and conditional branching (jumping to a label), this is similar to how conditionals are classically implemented in assembly. Let's me know your opinion on this plan. To think of it, in future we can maybe similarly add an Op for looping as well?

BoxiLi:
> Nice suggestion, how about the terms qreg, creg?

Great choice.

> So the reason I wanted to deprecate them was once Ops are implemented. Most users would be doing somthing likeqc.add_op(QFT, qubits=range(0, 8)). So instead of them juggling between add_gate, add_measurement, add_op, I wanted to unify the interface.

I definitely agree that `add_op` as a unified is a nice choice. I was just a bit worried about forcing this change to the user side immediately and deprecating the others right now. What about letting them coexist for a while? Internally, we lean towards using `add_op`, but still keep add_gate and add_measurement as syntax sugar. We can update our user guide to teach the use of `add_op`. And only deprecate the other options after a while when we have tested `add_op` enough and are sure about its syntax, and sure that this will bring further benefits? It is kind of like you can use `qiskit.cnot` and `qiskit.append` to add a CNOT gate.

> So essentially I did't want classical_control as an explicit argument for add_op since for most workflows an Op should encompass the classical control logic. 

This is an interesting question. After we kick the quantum target bits' information out of the gate class. We will be trying to keep classical control info inside `op`? This also rings a bell, if add_gate should take classical control info and add_op should not, how can we replace the former one with the later one? 

This is kind of the difference between `to_gate` and `to_instruction` in qiskit.

I need to think about it a bit more. Essentially, `ClassicallyControlled` is asking the user to wrap this in a class instance explicitly. Maybe instead, we should provide them syntax sugars like `add_circuit` and handle this for them internally? So we don't expose all the design details, and they don't need to learn all the code design behind this? 

> . The way I want to internally implement the ClassicallyControlled Op is via labels and conditional branching (jumping to a label), this is similar to how conditionals are classically implemented in assembly. Let's me know your opinion on this plan. To think of it, in future we can maybe similarly add an Op for looping as well?

Sounds like a very ambitious plan, even qisikit `to_instruction` does not accpet control flow at the moment.

Mayank447:
> I was just a bit worried about forcing this change to the user side immediately and deprecating the others right now. What about letting them coexist for a while? Internally, we lean towards using `add_op`, but still keep add_gate and add_measurement as syntax sugar. We can update our user guide to teach the use of `add_op`. And only deprecate the other options after a while when we have tested `add_op` enough and are sure about its syntax, and sure that this will bring further benefits?

Sounds good

> This is an interesting question. After we kick the quantum target bits' information out of the gate class. We will be trying to keep classical control info inside `op`? This also rings a bell, if add_gate should take classical control info and add_op should not, how can we replace the former one with the later one? This is kind of the difference between `to_gate` and `to_instruction` in qiskit.
>
> I need to think about it a bit more. Essentially, `ClassicallyControlled` is asking the user to wrap this in a class instance explicitly. Maybe instead, we should provide them syntax sugars like `add_circuit` and handle this for them internally? So we don't expose all the design details, and they don't need to learn all the code design behind this?

My only point of confusion is `creg` (used internally by an Op) and `classically_controlled_creg` (whether to apply the Op). Also having an overlap between them is allowed. I will also need to think about this, I do feel a much simpler and cleaner API is possible.

> Sounds like a very ambitious plan, even qisikit `to_instruction` does not accpet control flow at the moment.

Yeah, that's why a separate PR for it.

Mayank447:
This PR is complete. The failing test is due to https://github.com/qutip/qutip-qtrl/issues/39

@BoxiLi Is there a plan of doing a micro (patch) release for `qutip-qtrl` to fix the bug due to Scipy 1.18

BoxiLi:
@veronikakurth What is our policy now on qutip-qtrl? `disp` is now removed in scipy, which means that the package is unfortunately broken for `scipy>1.18` :( We were suppressing the warning there, I think.

veronikakurth:
> @veronikakurth What is our policy now on qutip-qtrl? `disp` is now removed in scipy, which means that the package is unfortunately broken for `scipy>1.18` :( We were suppressing the warning there, I think.

This is not decided officially yet; de facto, `qutip-qtrl` is in maintenance-only mode and will start approaching end-of-life at some point. Once we have a new stable version for `qutip-qoc`, we will decide on the exact and gradual timeline for it, just like other mature libraries do. By the time we can officially archive and deprecate it, I believe we have to maintain it for the existing users: fix what we can fix and resolve such compatibility issues as you mentioned. 

As of a fix for `qutip-qtrl` with scipy 1.18, I can take care of it.

Thank you!

veronikakurth:
The release of `qutip-qtrl` (v.0.2.0) with the fix for Scipy 1.18 is there.

Mayank447:
Thank you @veronikakurth !

@BoxiLi Could you re-run the tests for this PR

src/qutip_qip/circuit/circuit.py
```diff
@@ -248,6 +252,44 @@ def add_state(
...
+        if type(qreg) is int:
+            qreg = [qreg]
+
+        if type(creg) is int:
+            creg = [creg]
+
+        # TODO validate the inputs
+        self._ops.append(OpInstruction(op=op, qreg=tuple(qreg), creg=tuple(creg)))
+
+    def build(self: Self):
```
BoxiLi, line 270: This `build` needs to be called manually?

I have seen this in some other simulators, but that was like a `flush` command, which brings the circuit gates to C-level code. Wouldn't it be easier if we convert at the point of `add_op`?

BoxiLi (commented):
The part of the changes in result.py seems independent and can be merged directly to master?

src/qutip_qip/circuit/circuit.py
Mayank447: Yes, build needs to be called manually. Later, I will also add an argument to it to pass the basis gate set along with a default gate set.

I didn't want to convert at the point of `add_op` as we can do a lot of optimization during the build like  for repeating Op reusing the compiled ops to the given basis sets.

For backward compatibility, `qc.run()`, `qc.run_statistics()` etc. will automatically call `build()`.

Mayank447:
The changes in `result.py` are already there in the master. This branch was one behind the master

src/qutip_qip/circuit/circuit.py
BoxiLi: > repeating Op reusing the compiled ops to the given basis sets.

Elaborate a bit more here?

Is this something like circuit optimization or just internal micro performance boosting? How should I understand a `compiled ops`?

Mayank447: Circuit optimization (basically locally for an Op in terms of gate count instead of the global circuit). Also since `build` process won't be instanteous, I want to avoid at every `add_op`.

Compiled should be thought of as that Op consisting of gates in the basis set

BoxiLi: Maybe then we could have a separate pass like `circuit.optimize` dedicated to this?

Mayank447: So essentially there will be 3 things that happen in build

1. Replacing the Op with its bloq implementation. This is run time configurable like I can replace QFT Op with an Approximate QFT Bloq (lower gate count) instead of standard QFT Bloq. Build will have an argument for BloqRegistory (maps each Op to a Bloq) with a default BloqRegistory ser. I didn't add this yet, cause Op, Bloq classes haven't been implemented.

2. For repeating Ops using the already compiled Ops.

3. Converting all the gates in the circuit to the basis gates.

I essentially have build so that we convert Ops to only gate and measurement instructions, so that what is passed to the simulator remains the same. Also we can have an argument for optimization level similar to how Qiskit does, but we can only work on that once Ops are fully implemented.

BoxiLi:
On this topic, I was thinking again whether we should really merge measurement into gate ops. 

The difference is that measurement includes classical bits where the result is stored, while general gates don't have any information. And if measurement is a bit independent Op. The add_op needs to take classical bits as input. But it doesn't for other gates. Thoughts on this?

Mayank447:
I see where the confusion is. Op are allowed to have classical bits (creg). I just didn't want controlled cbits to entangle (not in physics sense) with cbits used in an Op like in mid circuit measurement etc.

There is no gate op or measurement op (each Op is composed of gate, measurement at the fundamental level, only relevant during the build process). Each Op will have the number of qubits, cbits it takes and verify what is passed on as the first argument in `add_op`. For gates, number of cbits will obviously be zero while for measurement it will have both a qubit and cbit. It might sound it a little complex right now, but the design and interfaces will be become simple and unify beautifully by the end.

Feel free to let me know if there are any questions.

PR #412, opened 2026-07-05
Classical control flow for Ops

Mayank447:
**Description**
This PR adds Classical control statements based of conditional branching and labels as done in assembly. Instead of adding a `ClassicallyControlledOp` or an argument to the `add_op` method, with context manager is used and is allowed to be nested. Example:

```
with qc.test_if((bit_1, bit_3), 0b11):
    qc.add_op(H, qreg=1)
```

This should also allow one to have multiple add_op statements within the same classical control condition without duplication that would have occurred in case of argument or ClassicallyControlledOp

Label, Conditional Branching are added to the fundamental `CircuitInstruction` and will in future serve as the fundamental instruction for any Quantum Program (basically QuantumAssembly Instruction) and hence won't be exposed to the user, only the general `OpInstruction` will be exposed to user.

@BoxiLi In the next PR I was planning on moving circuit draw from Gates, Measurements to Ops. This should free up the `style` attribute we current have in our `CircuitInstruction` class.

**Related issues or PRs**
None

**AI Tools Usage Disclosure**
None

### DIFF
--- a/src/qutip_qip/circuit/circuit.py
+++ b/src/qutip_qip/circuit/circuit.py
@@ -310,7 +310,15 @@ class QubitCircuit:
         self._instructions = []
 
         for op_instruction in self._ops:
-            if isinstance(op_instruction.op, Gate) or issubclass(
+            if isinstance(op_instruction.op, Label):
+                self._instructions.append()
+
+            elif isinstance(op_instruction.op, Cbz) or isinstance(
+                op_instruction.op, Cbnz
+            ):
+                self._instructions.append()
+
+            elif isinstance(op_instruction.op, Gate) or issubclass(
                 op_instruction.op, Gate
             ):
                 self._instructions.append(
--- a/src/qutip_qip/circuit/circuit.py
+++ b/src/qutip_qip/circuit/circuit.py
@@ -10,18 +10,19 @@ from typing import Iterable, Type, Self
 from qutip import qeye, Qobj, basis, tensor
 import numpy as np
 
-from qutip_qip.circuit import (
-    CircuitSimulator,
-    CircuitInstruction,
-    GateInstruction,
-    MeasurementInstruction,
-    OpInstruction,
-)
+from qutip_qip.circuit import CircuitSimulator, OpInstruction
 from qutip_qip.circuit._decompose import (
     _resolve_to_universal,
     _resolve_2q_basis,
 )
-from qutip_qip.circuit.conditional import Cbnz, Cbz, Label
+from qutip_qip.circuit.instruction import (
+    CircuitInstruction,
+    ConditionalBranchInstruction,
+    GateInstruction,
+    LabelInstruction,
+    MeasurementInstruction,
+)
+from qutip_qip.circuit.conditional import Cbnz, Cbz, Conditional, Label
 from qutip_qip.operations import (
     Gate,
     Measurement,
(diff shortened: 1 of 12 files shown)


src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -209,32 +217,39 @@ def run_statistics(self, state, cbits=None):
...
+            op = current_instruction.operation
+            label_name = op.label.name
+
+            if (
+                isinstance(op, Cbz) and self.cbits[current_instruction.cbits[0]] == 0
+            ) or (
+                isinstance(op, Cbnz) and self.cbits[current_instruction.cbits[0]] == 1
+            ):
+                self._op_index = self._label_map[label_name]
+
```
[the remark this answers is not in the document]
Mayank447: Collision, label missing will never happen given how build increments the label counter

src/qutip_qip/circuit/circuit.py
```diff
@@ -45,6 +48,13 @@ def DisplaySVG(data, *args, **kwargs):
         return data
 
 
+class ClassicalControlCheck(StrEnum):
+    EQ = "EQ"
+    NEQ = "NEQ"
+    GT = "GT"
+    LT = "LT"
```
BoxiLi, line 55: What do they mean? Maybe some doc strings?

I know equal not equal etc, but GT LT was not that familar to me.

src/qutip_qip/circuit/circuit.py
```diff
@@ -252,6 +265,34 @@ def add_state(
             for i in targets:
                 self.output_states[i] = state
 
+    @contextmanager
+    def test_if(
```
BoxiLi: Is `test_if` deliberately distinguished from `qiskit.if_test` ?

src/qutip_qip/circuit/circuit.py
Mayank447: I just picked the latex notation for these checks. Btw these haven't been implemented yet except EQ.

src/qutip_qip/circuit/circuit.py
```diff
@@ -252,6 +265,34 @@ def add_state(
...
+        self: Self,
+        cbits: int | IntSequence,
+        value: int,
```
BoxiLi, line 274: Here we need to be careful to use consistent representation: "01" is int 1 or 2, just to double check.

BoxiLi (commented):
There are a few places I don't fully understand yet, such as how we are going to handle the ConditionalBranchOperation. But I am ok to merge it to the developement branch

src/qutip_qip/circuit/circuit.py
Mayank447: I just found `test_if` to sound a little better

src/qutip_qip/circuit/circuit.py
Mayank447: So the way to enter binary numbers in python is using the prefix 0b. For example

10 is ten
0b10 is two

Both have the datatype int.

Mayank447:
> There are a few places I don't fully understand yet, such as how we are going to handle the ConditionalBranchOperation.

So essentially conditional branch currently skips to a label (just a position holder in the set of instructions, it doesn't effect the circuit). It's easier to visualise conditional branching if the circuit is represented as a DAG. I was also thinking of later replacing the linear storage of operations with a DAG representation, it conforms naturally to the Op representation. Thoughts?

> But I am ok to merge it to the developement branch

I will complete the conditional branching for neq etc. by tonight. After that we can merge this PR

src/qutip_qip/circuit/circuit.py
BoxiLi: Oh, I meant more on the quantum side, which one represents qubit 0, and wether that is the largest or smallest index.

Mayank447: So cbits is an array. Example: (cbit1, cbit2, cbit3). I then go and convert `value` (say 4) into its binary representation (100) and compare index wise.

Note: Passing (cbit2, cbit1, cbit3) is not same as passing (cbit1, cbit2, cbit3). The order matters

Mayank447:
@BoxiLi This PR is complete

src/qutip_qip/circuit/circuit.py
Mayank447: Btw.. I renamed it to `if_test`. Less confusion on the user end

BoxiLi (approved):

BoxiLi:
> It's easier to visualise conditional branching if the circuit is represented as a DAG. I was also thinking of later replacing the linear storage of operations with a DAG representation, it conforms naturally to the Op representation. Thoughts?

THe plan is some internal IR structure or a replacement of the current instruction list. I see some packages use this, the advantage is the representation of classical control?

Mayank447:
> THe plan is some internal IR structure or a replacement of the current instruction list. I see some packages use this, the advantage is the representation of classical control?

Internal IR structure. The final circuit after `qc.build()` will remain the same as current instruction list for pulse, gate simulation. Advantage of using DAG is mainly for writing transpiler passes and in our case for substituting Ops with their actual gate level implementation (Bloq), this is a little harder to do in a linear list.

PR #415, opened 2026-07-19
Fix plot_pulses colors on matplotlib 3.11

BoxiLi:
matplotlib 3.11 changed the default `axes.prop_cycle` to yield RGB tuples instead of hex strings. `plot_pulses` passed the cycle color as the third positional argument of `ax.fill/ax.plot`. That slot is only treated as a format string when it is a string, so the tuple was instead consumed as the x data of a further line: the pulse was drawn with no color at all (falling back to the first entry of the cycle, blue) and the requested color was rendered as a spurious three-point line of its own R, G and B values. On all-zero pulse rows, where `plot_pulses` skips `set_ylim`, that stray line autoscaled the axis and showed up as a visible spike.

Pass the color as a keyword argument, which accepts both spellings and so works on all supported matplotlib versions.

Before:
image
After
image

AI usage:
Claude Code Opus 4.8 was used to analyze and fix the bug.

### DIFF
--- a/src/qutip_qip/device/processor.py
+++ b/src/qutip_qip/device/processor.py
@@ -844,8 +844,13 @@ class Processor:
                 grid = grids[pulse_ind]
                 ax = plt.subplot(grid)
                 axis.append(ax)
-                ax.fill(tlist, coeff, color_list[i], alpha=0.7)
-                ax.plot(tlist, coeff, color_list[i])
+                # The color must be passed as a keyword argument. The third
+                # positional argument is only treated as a format string if it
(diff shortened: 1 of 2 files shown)


BoxiLi:
This is also seen in the latest rendered doc
https://qutip-qip.readthedocs.io/en/latest/qip-processor.html#spinchain

And the doc in the test
https://qutip-qip--415.org.readthedocs.build/en/415/qip-processor.html#spinchain

Mayank447 (approved):
LGTM

Issue #382, opened 2026-04-09
Make Device, Compiler classes abstract


Mayank447:
Make `Model`, `Processor`, `ModelProcessor` and `GateCompiler` classes an abstract class. Already in the codebase, all custom ModelProcessors, compiler etc. inherit from these classes. Which methods to be made an `@abstractmethod`, common class attributes shared with the base class etc. needs to be discussed beforehand.

syan-h:
hey can i take this issue if no one is doing it??

Mayank447:
Sure you're free to work on this, please read [CONTRIBUTING.md](https://github.com/qutip/qutip-qip/blob/master/CONTRIBUTING.md) to get started.

Also since it is a core issue, please discuss your base class design before opening a PR.

syan-h:
i found these methods that are candidates for abstratmethods 
1) Processor.load_circuit()
2) ModelProcessor.set_up_params()
3) ModelProcessor.topology_map()
and i couldn't find for the Model and GateCompiler cause they have default implementations . so are these okay and i should start working or maybe some other methods should be abstract that i didn't include

Mayank447:
I am not sure about any of them. @BoxiLi can better help with this

1. Load circuit essentially transpilers and then initializes the hamitonian model.

2. I don't remember `set_up_params` being initialized or used by any processor.

3. Topology map might work as an abstract parameter (Linear spin chain is just an abstract spin processor with topology set to a linear lattice). But currently we don't have a transpiler for arbitrary graph topology.

BoxiLi:
After a closer look, among the four of the classes, I think only `Model` can be made into an ABC or Protocol. The other three have real initializations and can in priciple work by themselves.

But the problem here is that if we make it abstract, we have to create a default backing model for Processor anyway... `Model` itself it not a pure template, but rather a complete workflow if `_control`, `_drift` or `_noise` are given. It includes the basic workflow. And the the paper use example one create MyModel(Model) and only define a get_control function. I don't see that we save a lot by making it abstract either.

`set_up_params` is somehow a dead function. I think it was a leftover of a clean up. This indeed should be remove

`topology_map` cannot be forced for every subclasses, because a processor can work without it in priciple.

> we don't have a transpiler for arbitrary graph topology.

This is going to be a lot of work. And I am not sure if we will go in that direction because transpiler is a huge research field. Here, we focus better on Hamiltonian modelling capability, closer integration with core qutip.

syan-h:
[...]

PR #416, opened 2026-07-19
Remove unused set_up_params method

syan-h:
This PR removes the unused set_up_params method from ModelProcessor.

In the discussion in #382, it was found that making the classes abstract is not the right  and set_up_params was found as dead code left over from an earlier cleanup. I verified that the method is not used elsewhere in the codebase, and removed it and confirmed that the test suite still passes.

Testing
Ran the full test suite with pytest.
Tests passed (589 passed, 13 skipped).

AI Tools Usage Disclosure

AI was used to understand the issue and some parts of codebase to get a better idea of what is it doing

### DIFF
--- a/src/qutip_qip/device/modelprocessor.py
+++ b/src/qutip_qip/device/modelprocessor.py
@@ -54,7 +54,6 @@ class ModelProcessor(Processor):
         self.transpile_functions = []
         self._default_compiler = None
 
-
     def run_state(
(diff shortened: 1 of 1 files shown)


Mayank447 (approved):

BoxiLi:
@syan-h thank you! Looks good, just fix the linting requirement. Please ignore the other failing test; that is just a network issue.

BoxiLi (approved):

PR #418, opened 2026-08-16
Op Bloq design

Mayank447:
**Description**
Add Op, Bloq Design as per #362 and integrates `BloqBuilder` within `QuantumCircuit` internally.

@BoxiLi In next PR I will implement the Op resolving part and will also clean up the `qc.resolve_gates` and decomposition patching we had using the former, so keeping the work until now as a separate PR.

Also I have deprecated, removed (with warning) a few methods in `QubitCircuit` which are either trivial like `reverse_circuit` or will no longer be possible e.g. `remove_gates_and_measurements`.

Note for myself: qasm generation for `if_test` needs to be looked into after the next PR.

**Related issues or PRs**
None

**AI Tools Usage Disclosure**
None

### DIFF
--- a/src/qutip_qip/circuit/circuit.py
+++ b/src/qutip_qip/circuit/circuit.py
@@ -584,29 +584,22 @@ class QubitCircuit:
         start : int
             The qubit on which the first gate is applied.
         """
+        warnings.warn(
+            "QubitCircuit.add_circuit has been deprecated, instead use Ops",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
         if self.num_qubits - start < qc.num_qubits:
             raise NotImplementedError("Targets exceed number of qubits.")
 
-        for circuit_op in qc.instructions:
-            if circuit_op.is_gate_instruction():
-                self.add_gate(
-                    circuit_op.operation,
-                    targets=[start + t for t in circuit_op.targets],
-                    controls=[start + c for c in circuit_op.controls],
-                    classical_controls=circuit_op.cbits,
-                    classical_control_value=circuit_op.cbits_ctrl_value,
-                )
-
-            elif circuit_op.is_measurement_instruction():
-                self.add_measurement(
-                    circuit_op.operation,
-                    targets=[target + start for target in circuit_op.qubits],
-                    classical_store=list(circuit_op.cbits),
-                )
-
-            else:
-                raise TypeError(f"The circuit to be added contains unknown \
-                    operator {circuit_op[0]}")
+        for circuit_op in qc.ops:
+            self.add_op(
+                op=circuit_op.op,
+                qreg=[start + t for t in circuit_op.qreg],
+                creg=circuit_op.creg,
+                style=circuit_op.style,
+            )
 
     def adjacent_gates(*args, **kwargs):
         raise AttributeError(
--- a/src/qutip_qip/circuit/circuit.py
+++ b/src/qutip_qip/circuit/circuit.py
@@ -92,6 +92,7 @@ class QubitCircuit:
 
         self.reverse_states = reverse_states
         self._instructions: list[CircuitInstruction] = []
+        self._built_count = -1
 
         if input_states:
             self.input_states = input_states
@@ -149,7 +150,7 @@ class QubitCircuit:
             DeprecationWarning,
             stacklevel=2,
         )
-        return self._instructions
+        return self.instructions
 
     @gates.setter
     def gates(self, value: any) -> None:
@@ -227,6 +228,8 @@ class QubitCircuit:
 
     @property
     def instructions(self) -> list[CircuitInstruction]:
+        if self._built_count != len(self._builder._op_instructions):
+            self.build()
         return self._instructions
 
     def add_state(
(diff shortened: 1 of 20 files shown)


src/qutip_qip/circuit/circuit.py
```diff
@@ -824,22 +637,13 @@ def reverse_circuit(self):
...
-                    targets=circ_instruction.qubits,
-                    classical_store=circ_instruction.cbits[0],
-                )
+        for circ_op in reversed(self.ops):
+            temp.add_op(
+                op=circ_op.op,
+                qreg=circ_op.qreg,
+                creg=circ_op.creg,
+                style=circ_op.style,
+            )
```
Mayank447, line 646: This reversing does not work for Classical Conditioning. Cbnz -> H -> CX is not the same as CX -> H -> Cbnz, it should be Cbnz -> CX -> H.

I will fix this in the next PR.

BoxiLi:
Ok, looks cool, might take some time for me to digest it. I may post some more questions piece by piece.

One fundamental change (not entirely new from here, but also in previous small PRs) I see is that classical conditions, instead of being carried by each instruction, are now separated out with independent ConditionalBranchInstruction. What is the advantage of this? To make the hanlding of classical workflow more nature?

Mayank447:
> I may post some more questions piece by piece.

Sure.

> One fundamental change (not entirely new from here, but also in previous small PRs) I see is that classical conditions, instead of being carried by each instruction, are now separated out with independent ConditionalBranchInstruction. What is the advantage of this? To make the hanlding of classical workflow more nature?

So I wanted to decouple classical instruction and gate/measurement we apply based on the classical condition. Example if cbit0=1 apply a sequence of these gates, we earlier had to add that classical controlled conditions for each each gate in the sequence. This made making things modular harder in case of an Op.

Example - 3QubitCode has classically controlled gates. Now when that 3QubitCode Op is used within ShorCode where it is classically controlled, how to map the classically controlled Op which also has a classically controlled gates (do take an union of cbits?).

The ConditionalBranch flow we added makes the decoupling easier (also bonus if we ever want to support MLIR as an IR for export that is now easy as well as it basically a directly translation). There is one catch that inverting a circuit with Condition branching and jump statements is hard as explained above for reversing a circuit. So I want to introduce a `ConditionalBlock` so that inverting circuits becomes, also separating the transpilation from within and outside the ConditionalBlock. I am still working the `ConditionalBlock` design.

src/qutip_qip/circuit/circuit.py
```diff
@@ -723,29 +571,22 @@ def add_circuit(self, qc, start=0):  # TODO Instead of start have a qubit mappin
         start : int
             The qubit on which the first gate is applied.
         """
+        warnings.warn(
+            "QubitCircuit.add_circuit has been deprecated, instead use Ops",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
```
BoxiLi, line 579: I think we can still keep this syntax sugar? It is just conceptually more straightforward, internanly the implementation is already quite straightforward. That is the gain we can see. This can still be convenient when we e.g. provide qubit mappting

src/qutip_qip/circuit/circuit.py
```diff
@@ -814,8 +621,14 @@ def reverse_circuit(self):
...
             qubit circuit in the reverse order.
-
         """
+
+        warnings.warn(
+            "QubitCircuit.reverse_circuit has been deprecated and will be future in future versions",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
```
BoxiLi, line 631: Same argument, unless we have to make a lot of detour to implement this (like for add_gate), we can keep the method

src/qutip_qip/circuit/circuit.py
```diff
@@ -769,41 +610,7 @@ def remove_gate_or_measurement(
...
-                    break
-
-        elif name is not None and remove == "all":
-            for i in reversed(range(len(self.instructions))):
-                if name == self.instructions[i].operation.name:
-                    self._instructions.pop(i)
-
-        else:
-            self._instructions.pop()
+        raise AttributeError("remove_gate_or_measurement method has been removed. ")
```
BoxiLi, line 614: This one is really hard to keep functioning?

BoxiLi (approved):
I left a few minor comments. And I am happy to merge. This is build machanism seems make circuit composition much eaiser.

src/qutip_qip/circuit/circuit.py
Mayank447: Yeah, since the gate/measurement binding would be now lazy.

src/qutip_qip/circuit/circuit.py
Mayank447: So basically the way I thought of writing Quantum Programs as an analogy to classical programs was the main function (Quantum Circuit) and functions Op - Bloq pairs) which we will call in our main function. Concatenating two main loops felt a little weird to me and I am still uncertain how qubit mapping will work in case of ancillas.

So to avoid logic repetition (DRY) I was in favour of deprecating `add_circuit`. But this is open for discussion.

src/qutip_qip/circuit/circuit.py
Mayank447: I wanted to replace this method with `inverse_circuit`.

src/qutip_qip/circuit/circuit.py
BoxiLi: Ok, let's keep them now and revisit them when things are more complete

PR #419, opened 2026-08-22
Measurement: Einsum Optimisation

mudit06mah:
**Description**
This PR adds an optimized `_apply_measurement()` method called `_apply_measurement_einsum()`, it replaces the former in the `step()` method for the `"state_vector_simulator" and "density_matrix_simulator"` (which are technically all of the modes).
This PR also adds a test for the same (currently in `test_circuit.py`)
This PR does not yet add any other documentation (other than a docstring)

Since this is intended for a new qutip release, some workflow tests fail as the old `qutip.einsum()` did not return the desired output(it split the tensor in half while in the new `qutip.einsum` the `_infer_out_dims()` method can use the equation string to determine the desired output dims), all tests pass locally:
image

**Related issues or PRs**
This PR is related to the GSoC project for qutip-qip, Related PRs are #402 and #406

**AI Tools Usage Disclosure**
Gemini 3.7 flash: Docstring, helping with rebasing and understanding the physics.

### DIFF
--- a/src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
+++ b/src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
@@ -1,15 +1,18 @@
 from itertools import product
 from operator import mul
 from functools import reduce
+from typing import Type
 import numpy as np
 import string
 
 from qutip import ket2dm, Qobj
 from qutip.core import data as _data
 from qutip.core.dimensions import einsum
+from qutip.measurement import measurement_statistics
 from qutip_qip.circuit.simulator import CircuitResult
-from qutip_qip.operations import expand_operator, Gate
+from qutip_qip.operations import expand_operator, Gate, Measurement
 from qutip_qip.typing import IntSequence
+from qutip_qip.circuit.simulator import CircuitResult
 
 
 def _decimal_to_binary(decimal, length):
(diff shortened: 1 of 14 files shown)


src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -1,14 +1,15 @@
...
 from functools import reduce
-import numpy as np
+from typing import Type
 import string
+import numpy as np
 
-from qutip import ket2dm, Qobj
-from qutip.core import data as _data
-from qutip.core.dimensions import einsum
+from qutip import ket2dm, Qobj, einsum
```
veronikakurth, line 8: Let's have qutip from its master as a dependency, to have the tests passed 🤞

BoxiLi:
It seems that it is not possible to pass all the tests unless we make them all against qutip master. 

How should we handle this branch then @Mayank447. So that it will not cause problems for your op implementation? A proper release of qutip will only be out in September

Mayank447:
Merging this branch to einsum should be fine. We can merge einsum to master once qutip 5.4 is out.

After this branch is merged to einsum, I will merge einsum to Op branch to prevent conflicts. It will take me make around 3-4 weeks to complete ops completely so the new qutip release will probably be out before that.

@BoxiLi Since the tests fail for qutip <= 5.3.1 , we should raise the minimum qutip version to 5.4 in `pyproject.toml`?

BoxiLi:
> Since the tests fail for qutip <= 5.3.1 , we should raise the minimum qutip version to 5.4 in pyproject.toml?

Yes, you are right, @mudit06mah >=5.4

PR #420, opened 2026-09-17
Issue #401 fixed

### LINKED ISSUE #401: Add a dedicated method to visualise states after measurement.
### Bug Description

While working on my PR#400, i noticed that the run() function while measuring the qubits gave only one qubit as the output in plot, but its probability plot showed its probability as something less than 1. 

Thats because the run() function in the qutip_qip\simulator\matrix_mul_simulator.py was calling the CircuitResult object via imports from results.py in the same directory, where the probabilities of the measured qubits are equated to the probability of the same qubit's measurement, which is conceptually wrong on the grounds of measurement of a state. The current behaviour is correct for the run_statistics() method, but is faulty for the run() method.

The histogram after the measurement should show the observed frequencies. If a particular state is observed, Its probability must be 1.0, after which the state collapses into the measured state.

Therefore, a dedicated mechanism for the histogram plotting for a state after measurement is required. Id suggest making a dedicated frequency_histogram for the same.

### Code to Reproduce the Bug

```shell
import qutip
from qutip_qip.circuit import QubitCircuit, CircuitSimulator
from qutip_qip.operations.gates import H
from qutip_qip.operations.measurement import Mz
import matplotlib.pyplot as plt


qc = QubitCircuit(1, num_cbits=1)
qc.add_gate(H, targets=[0])
qc.add_measurement(Mz, targets=[0], classical_store=0)

sim = CircuitSimulator(qc)

result = sim.run(state=qutip.basis(2, 0)) 

print(f"Cbits: {result.cbits}")
print(f"Probabilities: {result.probabilities}")

result.plot_histogram()
plt.show()
```

### Code Output

```shell
Cbits: [[np.int64(1)]]
Probabilities: [np.float64(0.5)]
```

### Expected Behaviour

Cbits: [[np.int64(1)]]
Probabilities: [np.float64(1.0)]

Image

### Your Environment

```shell
QuTiP Version:      5.2.3
Numpy Version:      2.4.4
Scipy Version:      1.17.1
Cython Version:     None
Matplotlib Version: 3.10.8
Python Version:     3.14.3
Number of CPUs:     12
BLAS Info:          scipy-openblas
INTEL MKL Ext:      None
Platform Info:      Windows (AMD64)
```

### Additional Context

The current code, when run, shows the following plot. 

Image

harsh-patches:
**Description**
Previously the run() function used to give the probability of the measured instance. This change will make it always return probabilty = 1 unless measure_results parameter != None.

**Related issues or PRs**
fixed #401 

**AI Tools Usage Disclosure**
Gemini 3.8 flash was used to understand the source code.


src/qutip_qip/circuit/simulator/matrix_mul_simulator.py
```diff
@@ -357,7 +357,10 @@ def _apply_measurement(
                 probabilities = [p / sum(probabilities) for p in probabilities]
                 outcome_indices = np.arange(len(probabilities))
                 i = np.random.choice(outcome_indices, p=probabilities)
-            self._probability *= probabilities[i]
+
+            if self._measure_results != None:
```
Mayank447, line 361: Thanks for this PR, first step to fixing the core issue would be to find why `measure_results`is becoming None.

Also the example in the issue needs to be added as a test case for verification

